PR #53, opened 2014-11-08
Access control

pidydx:
Merged in the AccessControl work I did refactoring and adding logging.  Originally these were in CoreAccessControl and AdminUIAccessControl classes, but it was suggested I should merge it all in rather than maintaining separate classes.  This also rearranges the inheritance a little so BaseAccessControl has the effect of NullAccessControl and all classes add to and override functions from Base to extend functionality.


destijl:
I'm going to merge this in to test internally.  Don't delete your branch in case I encounter problems and we need to export again before its committed.

Issue #87, opened 2015-01-06
wmi_parser errors


pidydx:
We get a lot of errors associated with the wmi_parser.  The clients are all 3.0.0.3 and the server is pretty recent.

ERROR:2015-01-02 00:56:59,498 flow_runner:873] Error in flow aff4:/hunts/W:F5AF45EF/C.0e550087c9a8a545/W:42170A93/W:635150B6 (aff4:/C.0e550087c9a8a545). Trace: Traceback (most recent call last):
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_2-py2.7.egg/grr/lib/flow_runner.py", line 528, in RunStateMethod
    responses=responses)
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_2-py2.7.egg/grr/lib/flow.py", line 310, in Decorated
    res = f(*args[:f.func_code.co_argcount])
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_2-py2.7.egg/grr/lib/flows/general/collectors.py", line 479, in ProcessCollected
    artifact_name, collector)
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_2-py2.7.egg/grr/lib/flows/general/collectors.py", line 650, in _ParseResponses
    for result in result_iterator:
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_2-py2.7.egg/grr/parsers/wmi_parser.py", line 103, in Parse
    size = int(result.get("Size"))
TypeError: int() argument must be a string or a number, not 'NoneType'

grrrrrrrrr:
This should now be fixed.

PR #88, opened 2015-01-07
Fixed missing TypeError exception

pidydx:
Workers/Enrollers can get hung up if this is not properly handled.  I'm not sure if this was just a missing TypeError or if there is some other behavior that is causing the WMI disk stuff to have TypeError instead of ValueError.  result.get("Size") will return None and then int() fails with TypeError as described here: https://github.com/google/grr/issues/87

This fix appears to top the WMI disk artifacts from causing worker/enroller hangups.

### DIFF
--- a/parsers/wmi_parser.py
+++ b/parsers/wmi_parser.py
@@ -101,12 +101,12 @@ class WMILogicalDisksParser(parsers.WMIQueryParser):
 
     try:
       size = int(result.get("Size"))
-    except ValueError:
+    except (ValueError, TypeError):
       size = None
 
     try:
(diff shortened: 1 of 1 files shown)


grrrrrrrrr:
Hey, I actually did the exact same bugfix at the same time so this PR doesn't change anything anymore. Good to see we agree though! :) Thanks!!

pidydx:
Hah perfect!  I was just staring at it thinking "This can't possibly be this simple.  This is killing workers and enrollers...  Something else must be wrong..."  I feel better now because either neither of us is wrong or both of us are. :)

PR #89, opened 2015-01-07
Fixes dmidecode artifact and corrects the whitelists for dmidecode and s...

pidydx:
...shd


grrrrrrrrr:
Argh, I actually pointed that bug out in a review (I missed the whitelist issue though), somehow it still made it in. Thanks for the fix!!

PR #96, opened 2015-01-12
MySQLAdvanced Datastore & Updated Benchmark Tests

pidydx:


destijl:
@grrrrrrrrr for review

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+
+  def __enter__(self):
+    return self
+
+  def __exit__(self, unused_type, unused_value, unused_traceback):
+    if self.queue:
+      self.queue.put(self)
+
+  def Execute(self, *args):
+    retries = 50
```
grrrrrrrrr: I suggest using fewer retries but sleeping a bit longer in between so the DB server doesn't get hammered.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+from grr.lib import utils
+from grr.lib.data_stores import common
+
+# pylint: disable=nonstandard-exception
+class Error(data_store.Error):
+  """Base class for all exceptions in this module."""
+# pylint: enable=nonstandard-exception
+
+
+class MySQLConnection(object):
```
grrrrrrrrr: The connection and the pool should be reused between the two MySQL based data stores - the changes here are good, we should take advantage of them.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+
+    query, args = self._BuildQuery("DELETE", subject)
+    self._ExecuteQuery(query, args)
+
+  def ResolveMulti(self, subject, predicates, token=None, timestamp=None):
+    """Resolves multiple predicates at once for one subject."""
+    self.security_manager.CheckDataStoreAccess(
+        token, [subject], self.GetRequiredResolveAccess(predicates))
+
+    for predicate in predicates:
```
grrrrrrrrr: This seems terribly inefficient to me. Doing lookups in bulk should give us *much* better performance. I ran the microbenchmarks on this PR yesterday and the results are pretty bad for this data store - there is a slowdown of factor 5 to sometimes factor 100 compared to the standard MySQL DS. I think the main reason is that we use so many unnecessary round trips here. Can we maybe try to eliminate them?

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+
+        if entry_timestamp is None:
+          entry_timestamp = timestamp
+
+        predicate = utils.SmartUnicode(attribute)
+        data = self._EncodeValue(value)
+
+        # Replacing means to delete all versions of the attribute first.
+        if replace or attribute in to_delete:
+          duplicates = self._CountDuplicateAttributes(subject, predicate)
```
grrrrrrrrr: This also should not be done one by one but in bulk.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+  def _CountDuplicateAttributes(self, subject, predicate):
+    query = "SELECT count(*) AS total FROM aff4 WHERE \
+subject_hash=unhex(md5(%s)) AND attribute_hash=unhex(md5(%s))"
+    args = [subject, predicate]
+    result = self._ExecuteQuery(query, args)
+    return int(result[0]["total"])
+
+  def _BuildReplaces(self, values):
+    transaction = []
+    value_sets = zip(values[0::4], values[1::4], values[2::4], values[3::4])
```
grrrrrrrrr: I wonder if it would be easier to just have this as a list of tuples?

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+    return [subjects, attributes, aff4]
+
+  def _ExecuteInsert(self, transaction):
+    """Get connection from pool and execute query"""
+    for query in transaction:
+      with self.pool.GetConnection() as cursor:
+        cursor.Execute(query["query"], query["args"])
+
+  def _ResolveRegex(self, subject, predicate_regex, token=None, timestamp=None, limit=None):
+    if limit and limit == 0:
```
grrrrrrrrr: This will never trigger.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+      return []
+
+    if isinstance(predicate_regex, str):
```
grrrrrrrrr: use basestring here instead of str

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+    """Build a mapping between column names and types."""
+    self.attribute_types = {}
+
+    for attribute in aff4.Attribute.PREDICATES.values():
+      self.attribute_types[attribute.predicate] = (
+          attribute.attribute_type.data_store_type)
+
+  def _EncodeValue(self, value):
+    """Encode the value for the attribute."""
+    if hasattr(value, "SerializeToString"):
```
grrrrrrrrr: I think I'd prefer catching attribute error instead of hasattr checking.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+      else:
+        #Always order results
+        sorting = "ORDER BY aff4.timestamp DESC"
+      #Add limit if set
+      if limit:
+        sorting += " LIMIT %s" % int(limit)
+    else:
+      fields = "aff4"
+
+    query = action + " " + fields + " " + tables + " " + criteria + " " + sorting
```
grrrrrrrrr: you could use " ".join([action, fields, tables, criteria, sorting]). Also, I feel like string interpolation is nicer than using + concatenation for most cases. Would you mind changing to that?

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -0,0 +1,636 @@
...
+    #Set fields, tables, and criteria and append args
+    if is_regex:
+      tables += " JOIN attributes ON aff4.attribute_hash=attributes.hash"
+      regex = re.match(r'(^[a-zA-Z0-9_]+:[a-zA-Z0-9_:]*)(.*)', predicate)
+      if regex:
+        #If predicate has prefix then break down for query optimizations
+        parts = predicate.split(":", 1)
+
+        #Add prefix usage for all regex queries
+        criteria += " aff4.subject_hash=unhex(md5(%s)) AND aff4.prefix=(%s)"
```
grrrrrrrrr: criteria starts with the same prefix on all branches, I suggest moving it before the if.

grrrrrrrrr:
Sorry that it took so long to look at this PR.

Overall this change looks really good, I'm a bit worried though about the many unnecessary round trips. I think we could make this data store a lot faster even by merging round trips together. Otherwise you might run into unexpected problems since GRR assumes that for example aff4.FACTORY.MultiOpen is pretty fast but with the current implementation it might take really long.

I know that this will increase the complexity by quite a bit but I'm sure it's worth it. If you don't have time to look at this, we can also merge it in as it is for now and we can try to optimize it later. What do you think?

lib/data_stores/mysql_advanced_data_store.py
pidydx: Yeah, this is really excessive, but it was there to deal with the constant deadlocking that the delete/insert operations were causing.  Deadlocks are basically down to 0 so this can be safely reduced.

lib/data_stores/mysql_advanced_data_store.py
pidydx: These are the ones I was using for benchmarking.  Is there a different set of tests I should also check?
MysqlAdvancedDataStoreCSVBenchmarks.manySubjectsFewAttrs --labels=benchmark
MysqlAdvancedDataStoreCSVBenchmarks.manySubjectsFewWithManyAttrs --labels=benchmark
MysqlAdvancedDataStoreCSVBenchmarks.fewSubjectsManyAttrs --labels=benchmark
MysqlAdvancedDataStoreCSVBenchmarks.manySubjectsManyAttrs --labels=benchmark

I suspect this also may be an artificiality of the benchmarks.  If I remember right the problem is that when you do the bulk lookup on the predicates you can't use the indexes efficiently and so in a live system there is a massive slowdown.  I will verify again with our DBA that this is why this got changed.

lib/data_stores/mysql_advanced_data_store.py
pidydx: I'm not sure what you mean.  This is the same basic behavior as the original except that it is doing some checks to determine if it should insert, delete/insert, or replace.  It still does the bulk inserts when called that way.  I tried to do bulk replaces in a similar fashion, but it caused some strange behavior that I couldn't figure out.

lib/data_stores/mysql_advanced_data_store.py
pidydx: I will look into this.  Looking at it now I'm not sure why I didn't just make a list of tuples in the first place.  I think at the time there was something preventing me from altering the MultiSet behavior for building those lists, but I'm not seeing it now.

lib/data_stores/mysql_advanced_data_store.py
pidydx: The line itself made sense at first glance, but now it occurs to me that limit being 0 is false.  I used the other datastores as models for a few things.  This came from the sqlite datastore.   

Should this just be changed to if limit == 0 in both datastores, or should the check just be removed?

lib/data_stores/mysql_advanced_data_store.py
pidydx: Also shamelessly lifted from the SQLite datastore to store all value blobs in a single column.  I can change both to whichever is the preferred method.

lib/data_stores/mysql_advanced_data_store.py
scudette: Limit can be None if it is not specified - If we think that limit =0 is the same as limit = None then we should also change the default value.

lib/data_stores/mysql_advanced_data_store.py
pidydx: Will do!

pidydx:
I am actually going to be pulling in the last few commits soon to start doing testing again (my test environment got taken over for flow testing for a few weeks).  I can get some of these things changed and make sure they play nice.  

Would it be possible to get a sanitized set of data that is more representative of what a running GRR instance looks like?  I did notice a lot of the benchmark tests were calling things in particularly inefficient ways and were using test data that is going to hide a lot of performance problems, particularly wrapped around using indexes and regexes.

lib/data_stores/mysql_advanced_data_store.py
pidydx: Is the "aff4.subject_hash=unhex(md5(%s))" piece what you are referring to?  That shouldn't be a problem to move.

lib/data_stores/mysql_advanced_data_store.py
pidydx: I think limit = 0 normally gets treated as literally 0 results allowed while limit = None is no limit.  This line just doesn't work for returning empty results for a limit of 0 because the logic in the if is bad.  However, the calling code should prevent this from even being called with limit = 0 in the first place.

lib/data_stores/mysql_advanced_data_store.py
grrrrrrrrr: I was running the microbenchmarks:

PYTHONPATH=. python grr/lib/data_stores/mysql_advanced_data_store_test.py --tests=MysqlAdvancedDataStoreBenchmarks

I agree that those benchmarks might not be accurately representing a real world instance, I know though that those round trips will cause slowdown in a big GRR deployment that we probably can avoid.

lib/data_stores/mysql_advanced_data_store.py
grrrrrrrrr: What this does is it sends one request to the mysql server for each predicate we want to set. It should be more efficient to send one query that counts them all at once...

lib/data_stores/mysql_advanced_data_store.py
grrrrrrrrr: Hm interesting, in the MySQL DS we do try/except. Can you maybe change both here and SQLite so we have some consistency?

lib/data_stores/mysql_advanced_data_store.py
pidydx: I'm pretty sure the round trips actually speed things up in a real GRR environment.  The problem is when you do the bulk query using WHERE IN or OR MySQL can't use the indexes efficiently so the bulk query takes much longer doing table scans than the individual optimized queries take in aggregate using indexes.  I am going to see if our DBA can come up with a better optimization though.

lib/data_stores/mysql_advanced_data_store.py
pidydx: Ah I see, for the pre check.  This is going to fall under the same MySQL index problem as the other bulk query.

lib/data_stores/mysql_advanced_data_store.py
scudette: It may be that round trips are insignificant in mysql compared to the time
it takes to process the data. In sqlite we kind of assume that round trips
are 0 because its really a local file anyway. Not sure how much we should
worry about round trips in mysql.

pidydx: Yup, just talked to DBA.  The specific problem is doing an ORDER BY after IN.  Multiple high speed queries vs really poorly performing GROUP BY + ORDER BY bulk query.

scudette: is it possible to do the order by client side then?

pidydx: The original code did, but I suspect that the MySQL engine does it considerably faster due to the indexes being more efficient than the python loops.  My guess would be that the overhead of using the connection pool to send multiple queries with small exact results is going to be lower than sending a single query with substantially larger results and then having to do the sorting and grouping in python.  Without the connection pool the round trip would probably be a larger factor.

The benchmark tests can't really test this properly due to the data and access patterns not quite lining up with an operational instance and trying to accurately test it live was painfully difficult.  I can say that during earlier testing the bulk queries caused by this were routinely taking much longer and showing up in the slow queries log and then also getting sorted in python, but that doesn't necessarily mean that the aggregate of individual faster queries is better or that other changes are responsible for the large scale performance improvement.

lib/data_stores/mysql_advanced_data_store.py
pidydx: The more I think about this the more I suspect I might actually be able to change this.  Since there is no ORDER BY involved in this so MySQL should be able to use the indexes to get the results back quickly.  I will see what I can do on this.

pidydx:
@grrrrrrrrr I think I got all of the changes in except for the optimization discussions.  I think the MultiSet checks could be optimized, but I think the ResolveMulti loop is already going to be substantially faster due to the index usage.   If the rest of it looks good now I think merge and do any additional optimizations later is the best plan.  Thanks!

grrrrrrrrr:
This looks really good now, can you please put that one function back and I'll merge it in. Sorry again that this takes forever, I totally blame the github code review process which is really tedious...

pidydx:
No worries.  There have been a lot of changes and cleanups along the way too.  For something like this I would definitely get it as close to right the first time as possible.  ResolveRow should be back in place now.

PR #99, opened 2015-01-20
Add -E to sudo easy_install selenium

tbartlett0:
This was required so that I could pass the http[s]_proxy environment vars to easy_install.


destijl:
Thanks!

destijl:
We're actually going to change this a bit and require the whole thing to run as root, this should still fix your problem.

PR #100, opened 2015-01-20
Add pip requirements

dionyziz:
This patch provides a requirements.txt file which lists the pip requirements. It was generated using `pip freeze` after installing the required packages manually. These packages are enough to run the test suite.

Using this, it should be possible to install grr on platforms without depending on aptitude for python packages.


destijl:
LGTM

PR #101, opened 2015-01-20
Move install script to GitHub

dionyziz:
The installation script used to be hosted on Google Drive. This has problems:

1. The URL is not authenticating. Even though it is HTTPS, anyone can host
   things on Google Drive. While users can rely on us providing a link via
   HTTPS-served documentation on a different domain, the URL of the script
   would be nice to be self-evidently authentiating.

2. The URL is not human-recognizable or memorable.

3. It's difficult to update the Google Drive file. For example, I don't know
   how to go about it currently. But everyone knows how to open a GitHub pull
   request.

4. The file is not versioned, with all that this entails. We can't blame or
   find when a feature was introduced.

This patch moves the installation file to GitHub.

### DIFF
--- a/install_script_ubuntu.sh
+++ b/install_script_ubuntu.sh
@@ -1,0 +1,213 @@
+#!/bin/bash
+#
+# Script to install GRR from scratch on an Ubuntu 12.04, 12.10 or 13.04 system.
+#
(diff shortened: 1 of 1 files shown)


sroberts:
:metal:

PR #102, opened 2015-01-23
Unfork

pidydx:
Forking baaad.  The multiprocessing stuff does not play nice with stats_store, datastore connection pools, and monitor ports.  Since GRR does not really support running in this configuration in a stable fashion it should probably just be removed.


grrrrrrrrr:
Awesome, thanks for this PR!

Removing the multiprocessing options has been on our list for a while now but we were reluctant to remove it without offering a suitable replacement. We should really do this asap though so if nobody complains, I'll just merge this in next week and we can think about alternatives later.

PR #109, opened 2015-02-02
Add vagrant linux build environment

destijl:
- Fix debian configs to work with new debhelper
- Make postinstall work on old ubuntu
- Add vagrant configs for building on lucid and trusty
- Add script to install all build dependencies from source

This is a clean commit history version of https://github.com/google/grr/pull/105 see there for comment history.


vagrant/install_from_src.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
grrrrrrrrr: I'd recommend a set +e here, there are so many things that could go wrong...

scripts/install_script_ubuntu.sh
```diff
@@ -42,6 +49,14 @@ then
   GRR_TESTING=1;
 fi
 
+if [ "$1" == "--build-deps" ] || [ "$2" == "--build-deps" ]; then
```
grrrrrrrrr: should we do this in a while / shift loop? install.sh --build-deps --localtest would already not work...

vagrant/install_from_src.sh
destijl: Done. Exposed a couple of non-obvious bugs too.  Openssl changed their symlinking/hashing scheme!

scripts/install_script_ubuntu.sh
destijl: Added proper option parsing.

destijl:
PTAL

grrrrrrrrr:
We need to also update the documentation that the install script parameters have changed. Otherwise, LGTM

destijl:
@grrrrrrrrr I'm going to merge this now, can address any comments on those last changes when I move it into the internal repo.

PR #110, opened 2015-02-05
Initial check in of cpp client comms program.

bgalehouse:
Goal of this project is for a small c++ program to communicate with
the grr server and to only start the python based client when necessary.

### DIFF
--- a/client/minicomm/subprocess_delegator.cc
+++ b/client/minicomm/subprocess_delegator.cc
@@ -12,7 +12,7 @@
 
 #include "base.h"
 #include "config.pb.h"
-#include "ops/security/grr/proto/jobs.pb.h"
+#include "grr/proto/jobs.pb.h"
 
 namespace grr {
 SubprocessDelegator::SubprocessDelegator(ClientConfig* config,
--- a/client/minicomm/subprocess_delegator.cc
+++ b/client/minicomm/subprocess_delegator.cc
@@ -38,7 +38,7 @@ SubprocessDelegator::~SubprocessDelegator() {
   child_spawned_.notify_all();
   // Writer thread might now be waiting to read from the inbox. Add an empty
   // message to unstick it.
-  inbox_->AddMessage(MessageQueue::Message());
+  inbox_->AddMessage(GrrMessage());
   writer_thread_.join();
   reader_thread_.join();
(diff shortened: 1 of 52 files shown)


bgalehouse:
PTAL @scudette @dionyziz @mbushkov @destijl

client/minicomm/subprocess_delegator.cc
```diff
@@ -0,0 +1,298 @@
...
+    return;
+  }
+
+  kill(child_pid_, SIGTERM);
+  // Give it time to end gracefully, then make sure it is really dead.
+  std::this_thread::sleep_for(std::chrono::seconds(5));
+  kill(child_pid_, SIGKILL);
+  int child_status;
+  // Don't let it become a zombie.
+  waitpid(child_pid_, &child_status, 0);
```
scudette: Maybe specify WNOHANG here and spin in a sleep loop to limit the length of time we wait?

bgalehouse: Good idea. When I wrote this, my thought was that we can count on SIGKILL to finish the job reasonably quickly. However, I've since discovered that if the child is waiting on a system call, we might have to wait for that to finish. If we do a WNOHANG/sleep loop here and it times out, it would be good to prevent the child from spending an unbounded time as a zombie. Any suggestions on how to do this? I see two options:

1) Keep a list of undead children and regularly try to reap them. Seems a bit ugly, opens the possibility of complaining if the list becomes large (probably indicates we are trying to do something buggy)

2) disable SIGCHLD. Downside is that we no longer can count on child_pid existing. Perhaps it could even be reused on a sufficiently crazy busy system.

scudette: I am not sure the best course of action. On the one hand waiting
indefinitely for the child to die would block the parent. But if we do
something else then it might spawn an infinite number of children which all
end up zombies.

The current code waits a reasonable time (2 sec) before giving up on
killing the child FWIW:
https://github.com/google/grr/blob/master/client/nanny/windows_nanny.cc#L165

bgalehouse: I've updated it to perform option 1), with a 5 zombie limit.

PR #113, opened 2015-02-09
Basic travis build script

dionyziz:
This pull request introduces a first version of the travis build. It makes the necessary changes required to get this running on travis systems.

I'm introducing two separate installation scripts that are responsible for different stages of setting up the grr environment in order to be able to run it on travis:

* `before_install.sh` installs all the necessary dependencies on the vanilla travis system. This includes ubuntu repository `apt-get install` commands and `apt-get install` commands from `plaso`,  and some manual installation of `.deb` files. These are explained in detail below.

* `install.sh` fires up the `pip` installation of packages and runs the required `make` command on the pure source in order to do the necessary building for artifacts and protobufs.

The several `apt-get install` commands from the official ubuntu repositories are mostly taken from the previous installation script that was available before this PR. However, additional packages had to be introduced, as the travis environment is a relatively clean ubuntu environment which is missing a lot of libraries. The travis environment also runs Ubuntu 12.04. This has some minor incompatibilities compared to Ubuntu 14.04 which is what the previous installation script was written for.

Some `apt-get install` commands from the `plaso` repository are required for packages that are not available in the regular ubuntu repositories (or do not include patches we need), such as m2crypto and pytsk.

The `pip` installation is based on the newly introduced `requirements.txt` file which has been separated out. These used to work with `apt-get`, but having a requirements.txt file is a better practice for python packages. The new requirements.txt has already been merged in my previous pull request #100 which was cherry-picked from this pull request, so only minor changes in that file are visible in this PR.

I'm also introducing the `run_tests.sh` which is used to run all the python unit tests. This is needed to include a black-list of excluded tests. The rationale behind excluded tests is to start with a minimal travis build and go from there. We can start adding more tests as we go. However, as travis will be used to evaluate external pull-requests, it's important that we use a black list instead of a white list.

The `run_tests.sh` file also filters out verbose error messages using `grep`. This is needed as travis runs out of both buffer sizes if we don't apply it. The two buffer sizes are the visible buffer (on the travis website) and the download buffer (for when downloading the log). Travis thinks a huge log indicates an infinite loop, so we have to keep our logging somewhat limited. While this should be fixed at the python logging module level, it's currently easier to achieve it with a grep, as there's some awkward things going on with config files in the test runner.

run_tests.py has been modified to include some bug fixes and work-arounds. While it is not perfect, it's a first step in getting this up. Additional bugfixes will be coming in next PRs.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -182,7 +182,7 @@ def main(argv=None):
     processes = {}
     print "Running tests with labels %s" % ",".join(flags.FLAGS.labels)
 
-    with utils.TempDirectory() as flags.FLAGS.temp_dir:
+    with utils.TempDirectory() as temp_dir:
       start = time.time()
       labels = set(flags.FLAGS.labels)
 
@@ -190,7 +190,7 @@ def main(argv=None):
         if labels and not DoesTestHaveLabels(cls, labels):
           continue
 
-        result_filename = os.path.join(flags.FLAGS.temp_dir, name)
+        result_filename = os.path.join(temp_dir, name)
 
         argv = [sys.executable] + sys.argv[:]
         if "--output" not in argv:
--- a/run_tests.py
+++ b/run_tests.py
@@ -175,6 +175,7 @@ def main(argv=None):
 
     suites = flags.FLAGS.tests or test_lib.GRRBaseTest.classes
     for test_suite in suites:
+      print "Running test %s" % test_suite
       RunTest(test_suite, stream=stream)
 
   else:
--- a/run_tests.py
+++ b/run_tests.py
@@ -36,6 +36,9 @@ from grr.worker import worker_test
 flags.DEFINE_string("output", None,
                     "The name of the file we write on (default stderr).")
 
+flags.DEFINE_string("exclude_tests", None,
+                    "A comma-separated list of tests to exclude form running.")
+
 flags.DEFINE_integer("processes", 0,
                      "Total number of simultaneous tests to run.")
 
@@ -157,6 +160,8 @@ def DoesTestHaveLabels(cls, labels):
 
 def main(argv=None):
   if flags.FLAGS.tests or flags.FLAGS.processes == 1:
+    print "Running test in single process mode..."
+
     stream = sys.stderr
 
     if flags.FLAGS.output:
@@ -174,13 +179,19 @@ def main(argv=None):
       sys.argv.append("--debug")
 
     suites = flags.FLAGS.tests or test_lib.GRRBaseTest.classes
+
     for test_suite in suites:
       print "Running test %s" % test_suite
+      sys.stdout.flush()
       RunTest(test_suite, stream=stream)
 
   else:
     processes = {}
     print "Running tests with labels %s" % ",".join(flags.FLAGS.labels)
+
+    exclude_tests = []
+    if flags.FLAGS.exclude_tests:
+      exclude_tests = flags.FLAGS.exclude_tests.split(",")
 
     with utils.TempDirectory() as temp_dir:
       start = time.time()
@@ -188,6 +199,10 @@ def main(argv=None):
 
       for name, cls in test_lib.GRRBaseTest.classes.items():
         if labels and not DoesTestHaveLabels(cls, labels):
+          continue
+
+        if name in exclude_tests:
+          print "Skipping test %s" % name
           continue
 
         result_filename = os.path.join(temp_dir, name)
--- a/run_tests.py
+++ b/run_tests.py
@@ -84,6 +84,20 @@ class GRREverythingTestLoader(test_lib.GRRTestLoader):
 
 
 def RunTest(test_suite, stream=None):
+  """Run an individual test.
+
+     Ignore the argument test_suite passed to this function, then
+     magically acquire an individual test name as specified by the --tests
+     flag, run it, and then exit the whole Python program completely.
+
+     Args:
+       test_suite (string): Ignored.
+       stream: The stream to print results to.
+
+      Returns:
+        This function does not return; it causes a program exit.
+     """
+
   out_fd = stream
   if stream:
     out_fd = StringIO.StringIO()
--- a/run_tests.py
+++ b/run_tests.py
@@ -198,10 +198,10 @@ def main(argv=None):
                               "processing mode, but %i were specified" %
                               len(suites))
 
-    for test_suite in suites:
-      print "Running test %s" % test_suite
-      sys.stdout.flush()
-      RunTest(test_suite, stream=stream)
+    test_suite = suites[0]
+    print "Running test %s" % test_suite
+    sys.stdout.flush()
+    RunTest(test_suite, stream=stream)
 
   else:
     processes = {}
--- a/run_tests.py
+++ b/run_tests.py
@@ -194,6 +194,10 @@ def main(argv=None):
 
     suites = flags.FLAGS.tests or test_lib.GRRBaseTest.classes
 
+    assert len(suites) == 1, ("Only a single test is supported in single "
+                              "processing mode, but %i were specified" %
+                              len(suites))
+
     for test_suite in suites:
       print "Running test %s" % test_suite
       sys.stdout.flush()
--- a/run_tests.py
+++ b/run_tests.py
@@ -103,12 +103,23 @@ def RunTest(test_suite, stream=None):
     out_fd = StringIO.StringIO()
 
   try:
+    # Here we use a GRREverythingTestLoader to load tests from.
+    # However, the fact that GRREverythingTestLoader loads all
+    # tests is irrelevant, because GrrTestProgram simply reads
+    # from the --tests flag passed to the program, so not all
+    # tests are ran. Only the test specified via --tests will
+    # be ran. Because --tests supports only one test at a time
+    # this will cause only an individual test to be ran.
+    # GrrTestProgram then terminates the execution of the whole
+    # python program using sys.exit() so this function does not
+    # return.
     test_lib.GrrTestProgram(argv=[sys.argv[0], test_suite],
                             testLoader=GRREverythingTestLoader(
                                 labels=flags.FLAGS.labels),
                             testRunner=unittest.TextTestRunner(
                                 stream=out_fd))
   finally:
+    # Clean up before the program exits.
     if stream:
       stream.write("Test name: %s\n" % test_suite)
       stream.write(out_fd.getvalue())
--- a/run_tests.py
+++ b/run_tests.py
@@ -173,7 +173,7 @@ def DoesTestHaveLabels(cls, labels):
 
 
 def main(argv=None):
-  if flags.FLAGS.tests or flags.FLAGS.processes == 1:
+  if flags.FLAGS.tests:
     print "Running test in single process mode..."
 
     stream = sys.stderr
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+cd ..
+echo 'Running tests...'
+PYTHONPATH=. python grr/run_tests.py --processes=1 2>&1|grep -v 'DEBUG:'|grep -v 'INFO:'
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -4,3 +4,5 @@ echo 'Running tests...'
 
 EXCLUDE_TESTS=HTTPDataStoreCSVBenchmarks,HTTPDataStoreBenchmarks,MicroBenchmarks,TDBDataStoreBenchmarks,FakeDataStoreBenchmarks,AverageMicroBenchmarks,SqliteDataStoreBenchmarks,DataStoreCSVBenchmarks,TDBDataStoreCSVBenchmarks,AFF4Benchmark
 PYTHONPATH=. python grr/run_tests.py --processes=1 --exclude_tests=$EXCLUDE_TESTS 2>&1|grep -v 'DEBUG:'|grep -v 'INFO:'
+TEST_STATUS=${PIPESTATUS[0]}
+exit $TEST_STATUS
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
 cd ..
 echo 'Running tests...'
-PYTHONPATH=. python grr/run_tests.py --processes=1 2>&1|grep -v 'DEBUG:'|grep -v 'INFO:'
+
+EXCLUDE_TESTS=HTTPDataStoreCSVBenchmarks,HTTPDataStoreBenchmarks,MicroBenchmarks,TDBDataStoreBenchmarks,FakeDataStoreBenchmarks,AverageMicroBenchmarks,SqliteDataStoreBenchmarks,DataStoreCSVBenchmarks,TDBDataStoreCSVBenchmarks,AFF4Benchmark
+PYTHONPATH=. python grr/run_tests.py --processes=1 --exclude_tests=$EXCLUDE_TESTS 2>&1|grep -v 'DEBUG:'|grep -v 'INFO:'
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.pyc
 proto/*_pb2.py
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+*.pyc
+proto/*_pb2.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ python:
   - "2.7"
 before_install: /bin/bash before_install.sh
 install: /bin/bash install.sh
-script: cd ..; PYTHONPATH=. python grr/run_tests.py --processes=1
+script: /bin/bash run_tests.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ virtualenv:
 language: python
 python:
   - "2.7"
+before_install: /bin/bash before_install.sh
 install: /bin/bash install.sh
 script: cd ..; PYTHONPATH=. python grr/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ python:
   - "2.7"
 before_install: /bin/bash before_install.sh
 install: /bin/bash install.sh
-script: cd ..; PYTHONPATH=. python grr/run_tests.py
+script: cd ..; PYTHONPATH=. python grr/run_tests.py --processes=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,5 @@ virtualenv:
 language: python
 python:
   - "2.7"
-install:
-  - "pip install -r requirements.txt"
-  - "cd proto; make"
-  - "cd artifacts; make"
+install: /bin/bash install.sh
 script: cd ..; PYTHONPATH=. python grr/run_tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+virtualenv:
+  system_site_packages: true
+language: python
+python:
+  - "2.7"
+install:
+  - "pip install -r requirements.txt"
+  - "cd proto; make"
+  - "cd artifacts; make"
+script: cd ..; PYTHONPATH=. python grr/run_tests.py
--- a/artifacts/Makefile
+++ b/artifacts/Makefile
@@ -8,7 +8,7 @@ all: sync
 
 sync:
 	git clone https://github.com/ForensicArtifacts/artifacts.git $(TEMPDIR)
-	-rm *.yaml
+	-rm -f *.yaml
 
 	cp $(TEMPDIR)/definitions/*.yaml .
 
@@ -17,4 +17,4 @@ sync:
 .PHONY: sync clean
 
 clean:
-	-rm *.yaml
+	-rm -f *.yaml
--- a/before_install.sh
+++ b/before_install.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+sudo apt-get install python-protobuf libprotoc-dev protobuf-compiler m2crypto python-support libdistorm64-1 libdistorm64-dev python-psutil pytsk3 ncurses-dev python-pip
+
+PLAT=amd64
+DEB_DEPENDENCIES_URL=https://googledrive.com/host/0B1wsLqFoT7i2aW5mWXNDX1NtTnc/ubuntu-12.04-${PLAT}-debs.tar.gz
+DEB_DEPENDENCIES_DIR=ubuntu-12.04-${PLAT}-debs
+SLEUTHKIT_DEB=sleuthkit-lib_3.2.3-1_${PLAT}.deb
+
+DEB_DEPENDENCIES_TARBALL=$(basename ${DEB_DEPENDENCIES_URL})
+wget --no-verbose ${DEB_DEPENDENCIES_URL} -O ${DEB_DEPENDENCIES_TARBALL}
+tar zxfv ${DEB_DEPENDENCIES_TARBALL}
+sudo dpkg -i ${DEB_DEPENDENCIES_DIR}/${SLEUTHKIT_DEB}
--- a/before_install.sh
+++ b/before_install.sh
@@ -1,5 +1,20 @@
 #!/bin/bash
-sudo apt-get install python-protobuf libprotoc-dev protobuf-compiler m2crypto python-support libdistorm64-1 libdistorm64-dev python-psutil pytsk3 ncurses-dev python-pip
+PROTO_PACKAGES=(libprotobuf9_2.6.1-1ppa1~precise_amd64.deb libprotobuf-lite9_2.6.1-1ppa1~precise_amd64.deb libprotobuf-dev_2.6.1-1ppa1~precise_amd64.deb libprotobuf-java_2.6.1-1ppa1~precise_all.deb libprotoc9_2.6.1-1ppa1~precise_amd64.deb libprotoc-dev_2.6.1-1ppa1~precise_amd64.deb protobuf-compiler_2.6.1-1ppa1~precise_amd64.deb python-protobuf_2.6.1-1ppa1~precise_amd64.deb)
+
+mkdir protobuf-debs
+cd protobuf-debs
+
+echo "Installing protobuf debs"
+for i in ${PROTO_PACKAGES[@]}; do
+  echo "Downloading custom package $i"
+  wget https://dionyziz.com/protobuf-debs/$i
+  echo "Installing $i"
+  sudo dpkg -i $i
+done
+cd ..
+echo "Protobuf debs installed"
+
+sudo apt-get install m2crypto python-support libdistorm64-1 libdistorm64-dev python-psutil pytsk3 ncurses-dev python-pip
 
 PLAT=amd64
 DEB_DEPENDENCIES_URL=https://googledrive.com/host/0B1wsLqFoT7i2aW5mWXNDX1NtTnc/ubuntu-12.04-${PLAT}-debs.tar.gz
--- a/before_install.sh
+++ b/before_install.sh
@@ -14,6 +14,9 @@ done
 cd ..
 echo "Protobuf debs installed"
 
+sudo apt-get install python-software-properties
+sudo add-apt-repository ppa:kristinn-l/plaso-dev -y
+sudo apt-get update -q
 sudo apt-get install m2crypto python-support libdistorm64-1 libdistorm64-dev python-psutil pytsk3 ncurses-dev python-pip
 
 PLAT=amd64
--- a/install.sh
+++ b/install.sh
@@ -1,16 +1,5 @@
 #!/bin/bash
 pip install -r requirements.txt
-sudo apt-get install python-protobuf libprotoc-dev protobuf-compiler m2crypto python-support libdistorm64-1 libdistorm64-dev python-psutil pytsk3 ncurses-dev python-pip
-
-PLAT=amd64
-DEB_DEPENDENCIES_URL=https://googledrive.com/host/0B1wsLqFoT7i2aW5mWXNDX1NtTnc/ubuntu-12.04-${PLAT}-debs.tar.gz
-DEB_DEPENDENCIES_DIR=ubuntu-12.04-${PLAT}-debs
-SLEUTHKIT_DEB=sleuthkit-lib_3.2.3-1_${PLAT}.deb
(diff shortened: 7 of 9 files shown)


run_tests.py
```diff
@@ -80,24 +83,47 @@ class GRREverythingTestLoader(test_lib.GRRTestLoader):
...
-                            testRunner=unittest.TextTestRunner(
-                                stream=out_fd))
+    program = test_lib.GrrTestProgram(
+                argv=[sys.argv[0], test_suite],
+                testLoader=GRROneTestLoader(
+                    test_class_name=test_suite),
+                testRunner=unittest.TextTestRunner(
+                    stream=out_fd))
+  except SystemExit as e:
+    # must include this section, as GrrTestProgram issues
```
scudette: that sounds like a bug? Which test is doing this? These run in their own process though anyway.

run_tests.sh
```diff
@@ -0,0 +1,7 @@
...
...t,UserTests,TestSearchFileContentWithFixture,TestCollections,ArtifactParserTests,DarwinPersistenceMechanismsParserTest,RDFValueCollectionApiObjectRendererTest,WebHistoryFlowTest,HTTPDataStoreTest,EmbeddedRDFValueTest,DataAgnosticExportConverterTest,GRRSeleniumTest,SqliteDataStoreBenchmarks,TestSystemRootSystemDriveFallbackFlow,OutputpluginsTest,ConfigFileTest,FlowTest,RegistryFlowTest,ExportTest,MatchMethodTests,ResultSetTest,FakeDataStoreTest,HintDefinitionTests,TestAdministrativeFlows,GRRTempFileTestFilename,FilestoreStatsCronFlowTest,ForemanTests,ReportsTest,GRRBaseTest,TestEndToEndTestFlow,DictTest,TestArtifactCollectorsRealArtifacts,KnowledgeBaseUserMergeTest,HandlerTests,NetstatTest,BasicContextTests,ArtifactHandlingTest,ObjcTest,TestWindowsMemory,WindowsActionTests,GrepTest,SystemCronFlowTest,UtilsTest,DataStoreCSVBenchmarks,CheckLoaderTests,TDBDataStoreCSVBenchmarks,ArtifactKBTest,MultiShardedQueueManagerTest,ConfigActionTest,SearchTest,RegistryVFSTests,OSXDriverTests,RekallTestSuite,GeneralFlowsTest,FlowStateTest,TestCryptoTypeInfos,AFF4Benchmark,GRRArtifactTest,GRRFuseDatastoreOnlyTest,TestOSXFileParsing,AFF4SymlinkTest,FlowTestsBaseclass,ConditionTest,RekallVADParserTest
```
scudette: why exclude all these? is it better to just include the ones you want instead?

run_tests.py
scudette: commit list is longer than cl - did not read. Please squash

dionyziz: This is a work-in-progress as indicated in the "[WIP]" title. I'm in the process of squashing my history to something clearer so that you can review without trouble. I'll indicate when this is ready for review by requesting it with a PTAL and a removal of the WIP from the title. Thanks! :)

run_tests.sh
dionyziz: I'm excluding these because they fail and I plan to fix them one-by-one. However, fixing them one-by-one takes a long time and this pull request is starting to diverge from master in a way that makes future conflict resolution painful.

We'd like to have a working travis build ready and merged soon, as it will help with the C++ build process, so this is a temporary solution for expediting this merge. That's why they're being excluded.

I see this as a positive first step towards getting a proper travis build going: It's better to have only a handful of tests running instead of nothing. So I think this PR is worth merging as it is. This strategy is followed by other GitHub projects gradually enabling travis as well.

As for including the ones we want instead of excluding, this is a situation of a blacklist VS whitelist. These tests are explicitly black listed as "potentially failing" and must be looked at. However, I don't want to use a white list, as from now on we should use travis as an indicator of a successful build. Therefore, if another pull request comes in and includes a new test, it should be included by default.

That way, we will be able to validate the build and verify the correctness of patches by open source contributors and start being more open and cooperating with external parties in the development of Grr.

run_tests.py
scudette: Hmm ok sounds good.

run_tests.sh
scudette: if you expect these tests to fail, then let them fail - travis will tell
you they are failing and you can fix it then. disabling failed tests seem
like cheating :-). I wonder why they are failing because they do pass in a
regular source checkout. It might be something simple due to the way you
are installing the travil environ?

dionyziz: We can't let them fail, because then the purpose of travis is defeated. Travis is being installed so that we can check new PRs automatically against a build we know is succeeding. If it fails always, there's no point in it and it doesn't serve any purpose.

If we exclude the failing tests, the build process is still useful. We can then use it to evaluate future PRs with slowly increasing confidence as we start enabling more tests. "Letting them fail" won't help in that direction.

Yes, you're right, it is cheating, and this is fair at this stage. We're gradually migrating to travis instead of in one big go, which I'm sure you will agree is a reasonable approach for any non-trivial migration in software. One big go is impossible, because the failing tests are too many and too complicated to fix. They are succeeding in the regular source checkout because the environment is different. Travis is a different environment and some things have to be taken into account. Travis is not Goobuntu. Travis is not Ubuntu 14.04, but Ubuntu 14.02. Travis doesn't have all the same libraries that we have installed. Travis has different limits in memory, number of concurrent processes, log size. All these things must be taken into account. Therefore, you are right that it's something "due to the way I am installing in the travis environment", but it's not simple, and it will take time to fix.

For these reasons, I am making this pull request in order to start the process and I'll follow up with more PRs from there. It's hard to make this PR wait, because we're working with @bgalehouse on the C++ build files which I'd also like to include in our travis build. The more this divergences, the harder it will become to resolve conflicts later.

run_tests.py
dionyziz: Regarding your comment about the GrrProgram bug: Yes, it is a bug.

In travis, they don't run in their own process, because I'm specifying `--processes=1` in the invocation, so it never runs in multiprocessing mode: It runs in single processing mode, and all the tests run sequentially in a single process. This is intentional, as otherwise we run out of memory in travis, so we must run processes sequentially. There's also additional limits in travis, but they're not all clear to me. For example, I suspect there's a limit in open files and a limit of running processes that may or may not be causing problems. [See this failing build as an example](https://s3.amazonaws.com/archive.travis-ci.org/jobs/45939849/log.txt)

The bug is not that the GrrTestProgram issues a `sys.exit` – this is normal, as it extends the unittest [TestProgram](https://pythonhosted.org/gchecky/unittest.TestProgram-class.html), which is supposed to be a stand-alone CLI program that runs tests. It's not meant to be called from within functions like we are doing here, so this code was and still is wrong. This is what the TestRunner is for. All tests are behaving like this. It's not a bug in a specific test. It's a bug in the fact that a unit test program is invoked within a function, a misuse of the Python unit testing framework.

In fact, automated test discovery and running frameworks such as [nose](https://nose.readthedocs.org/en/latest/) provide for better such mechanisms and we should not reinventing the wheel with buggy behavior or incorrect use of the provided framework. The only reason why nose is not easily pluggable to the current implementation is our use of Google flags, but I believe this can be worked around.

So, for now I recommend that we don't require the whole testing framework to be rewritten in this one pull request to be correct. This is a pull request that improves what we currently have and works around some of the bugs, and fixes others, while it doesn't introduce new ones. In future pull requests I'll try to tidy up the code more and to properly fix this wrong invocation. This should still be merged as it is.

dionyziz:
Rebased to clean history of 16 carefully chosen commits. PTAL @grrrrrrrrr.

@scudette If you find it tiresome to look at individual commits, please take a look at the PR-level diff.

run_tests.py
scudette: Maybe that's the problem - they need to run in their own processes even if there is only one process. This way the test suite can be aware if a test crashes for example (otherwise it brings the entire test suite down with it). I can see that it uses a different code path when processes == 1 - maybe we should change that.

The failure you are referring to shows that the test is killed probably due to OOM. We see this occasionally in Rekall too - it seems that travis does not have a lot of memory for tests and its easy to exhaust it. Maybe we should either rerun the test if its killed or consider it to pass.

The whole idea with the test suite is that it will run all the tests as stand alone programs - this allows us to run anything including c++ tests the same way. The test is not called from a function - it is called from a runner which is forked for it.

BTW I have no idea what you mean by Google flags? We use argparse which is a standard python module.

run_tests.sh
scudette: No. The purpose of travis is to let you know when tests are failing not to render a nice green button. If some tests pass and some fail then this is how it is - no point turning tests off just to get that nice green feeling :).

We dont have to require all new PRs to be travis clean if that is not the case - turning off tests doesnt give you any information anyway since it hides the fact they are failing anyway. 

The C++ client will likely require its own test program to run as part of the travis setup and so does not really depend on these tests passing anyway.

run_tests.py
dionyziz: > Maybe that's the problem - they need to run in their own processes even if there is only one process. This way the test suite can be aware if a test crashes for example (otherwise it brings the entire test suite down with it).

This problem does not exist currently. I can see your point for this, but most tests fail in expected (assertion failure) or unexpected (uncaught exception) ways which are handleable by the testing framework. It doesn't make sense to require this one PR to make everything perfect right now, so I will keep it the way it is. Even if we wanted to isolate every test, running them in parallel is not a requirement; they could still run serially. I don't want to do this refactoring right now; let's file an issue if you consider it necessary. Feel free to assign it to me and I'll be happy to work on it later.

> I can see that it uses a different code path when processes == 1 - maybe we should change that.

There are many wrong things with this code. I don't want to fix everything in this PR. This PR improves the code in meaningful ways, so I suggest we merge it as it is. The rest we can report as issues on GitHub and I'll be happy to address them individually.

> The failure you are referring to shows that the test is killed probably due to OOM. We see this occasionally in Rekall too - it seems that travis does not have a lot of memory for tests and its easy to exhaust it. Maybe we should either rerun the test if its killed or consider it to pass.

I don't consider this a good option, because OOM causes the whole build to fail, not an individual test. So just marking it green in this case doesn't help us evaluate whether the tests are passing or not. OOM errors should cause the test to be red. We should also look into optimizing GRR tests for memory usage.

Rerunning is not an option. Please take a look at the build history, and you will see that the OOM error is reproducible every time on travis when multiprocessing mode is used. Therefore, I am suggesting single process mode in this PR.

> The whole idea with the test suite is that it will run all the tests as stand alone programs - this allows us to run anything including c++ tests the same way. The test is not called from a function - it is called from a runner which is forked for it.

This is not what the `run_tests.py` program did. The `run_tests.py` program had two modes of invocation: Multiprocess and single process modes. While multiprocess mode invoked single process mode, single process mode was and is invocable on its own, as you can see from the `processes` flag which can be set to 1. So, it should work for multiple tests as well. You can see that `GrrTestProgram` was expected to work in the way I am describing without exiting, because it was called within a `for` loop in single processing mode. It seems that this was never tested with multiple tests, so it is wrong, but it was [indeed called from a function](https://github.com/google/grr/blob/master/run_tests.py#L178).

As for the point of the test suite, you don't always want to run all the tests as a stand alone program each. Most testing frameworks and test runners don't do that, including Python's and nose. The behavior of running things within one process is desired for various reasons, including maintainability, simplicity, and sometimes performance depending on the context. While process-level isolation may be good practice in certain circumstances, I don't see why that's "the whole idea" of testing.

> BTW I have no idea what you mean by Google flags? We use argparse which is a standard python module.

Individual tests are not invocable on their own. The called has to go through the `run_tests.py` program, which maintains the `flags` global variable. This behavior causes problems with the `nose` test discovery mechanism. Feel free to try running `nosetests` within the grr directory to see for yourself. While this problematic behavior doesn't allow us to get rid of `run_tests.py` easily, it should be possible to replace with some effort. This should go in a different PR also.

run_tests.sh
dionyziz: > No. The purpose of travis is to let you know when tests are failing not to render a nice green button. If some tests pass and some fail then this is how it is - no point turning tests off just to get that nice green feeling :).

The purpose of travis is to let you know if the requested pull is breaking the build or not, and if it's safe to merge. If it is always red, it doesn't serve any purposes, because we're going to be ignoring it. If we adopt a policy of merging only green builds, we'll have to freeze all development until the test suite becomes green as a whole. This will take time and is inconvenient, as other things are going on currently.

I'm not turning tests off to get the nice green feeling. I'm turning tests off so that the rest of the tests, that are enabled, constitute a meaningful build which every new PR can be checked against. Once that is green, we can adopt a policy of merging only green pull requests and requesting from contributors to fix their build if they submit something that is red.

This is exactly the reason why I'll be turning on tests gradually as I fix them. Keep in mind that the failing tests are not failing because the tests are wrong or the implementation is wrong, but because the environment is alien and they need to be migrated carefully. I'll be happy to make an issue for each one and try to fix it individually in subsequent pull requests.

> We dont have to require all new PRs to be travis clean if that is not the case - turning off tests doesnt give you any information anyway since it hides the fact they are failing anyway.

It gives you the information that the tests which are enabled are still succeeding after a pull request will be merged. This gives you confidence that everything that you knew was working previously is still working. 10% coverage is still better than no coverage, so this is meaningful and informative.

If we don't require PRs to be green, then travis doesn't do us any favour. We might as well run tests manually. The automation of the build and the information provided to external contributors that they've done something wrong is what's valuable. If it's always red, there's no point in it, and that's what you're suggesting here.

This technique [has been used successfully in projects such as OpenBazaar](https://github.com/OpenBazaar/OpenBazaar/pull/1106/files) with great results in gradually improved testability for both unit tests and coding style.

> The C++ client will likely require its own test program to run as part of the travis setup and so does not really depend on these tests passing anyway.

It depends on the .travis.yml file being merged.

run_tests.py
scudette: The existing test script can run in two modes - inline: That is really only designed to run one test at the time and its used for debugging mostly (i.e. so you can put break points etc). The single process mode is really designed to run interactively only.

The script is also able to run in multiprocessing mode (which we usually use) and thats used to run the entire suite.

It appears Travis cant run more than one test at the time due to resource issues, so we need to reduce the number of concurrent tests to one - currently running the tests inline is not really a supported mode of operation and the code specifically checks for processes == 1 and goes into that code path. You claim that this is not reliable since some tests are exiting so I am suggesting we stick to the multiprocess mode since it is the current way things work and will require minimal new code.

All the grr test programs are stand alone (i.e. they have their own main()) and are actually designed to run stand alone. This is the best practice for running tests because you ensure that you have test isoltion when you start a new process each time (you can not guarantee this if you dont because we have a lot of shared globals like config library etc). In Rekall we actually spawn the test binary itself but I think in GRR we fork (maybe we should spawn).

Not all tests should run as a separate program but test suites (which are collections of tests) should.

Individual tests are invocable on their own by simply running the test program. They can also be specified as an arg to the run_tests.py script but this is just a convenience. I cant comment on nose since I never used it and I think changing the test framework is probably outside the scope of this cl.

I just suggested to remove that processes == 1 condition from the top and just see if that improves things on travis by running the test program in a child process.

Note that tests can fail also due to a seg fault and catching the exception wont help here.

run_tests.sh
scudette: I can see your point but my point is that hiding a test is the same as making it red - it is failing so you disable it. Therefore incoming PRs may very well be breaking it but you dont know. Its the head in the sand approach. I think its fair to require a new PR to not break tests and when we get to a green state this will be easier to see but disabling tests is not really helping here.

What exactly is the goal of this CL? If the goal is to get travis build/testing working then it means all tests must pass. Since I know that the test suite passes when run on their own I was hoping this CL can get it to run on travis completely. If you are saying that not all tests are passing then it means this CL is not complete and we should wait until its finished. I dont see the point of submitting a half finished effort just to have that undone in the next CL. Lets get it finished before we use it.

Disabling tests feels like an attempt to get something to be submitted without it being finished. It feels sloppy and there is no reason for it other than shortness of time. If indeed you suggest we adopt a policy of not accepting external PRs which break the build (i.e. they introduce bugs that affect existing tests) that means that those PRs are not finished either. Do you suggest we accept PRs that say - look at this new feature. I know it breaks some tests but i will fix those in the next PR. Clearly not.

Submitting a half working travis test environment is no different. Not all the tests are run so we dont know if its going to work at all.
It is concerning that tests are being killed due to OOM - does this mean travis as a platform is too small for us to run on? If this is the case we dont necessarily want to invest more time in it. This is why we want to see that all tests pass before we commit to using travis at all.

run_tests.py
dionyziz: > The existing test script can run in two modes - inline: That is really only designed to run one test at the time and its used for debugging mostly (i.e. so you can put break points etc). The single process mode is really designed to run interactively only.

It is very unclear to me why the single-process mode is "designed to run one test at the time". There are no comments indicating this. There is no documentation indicating this. The code is written in a way that strongly indicates otherwise. As I said in my previous comment, and as you can readily see in the code, there is a **for loop** looping through multiple tests in the single-process mode path (also quoted below). Why is there a for loop looping through an array which is only supposed to have one element by design according to what you said? Why is it even an array then and not just a single value? This doesn't make any sense. This code is just badly designed and buggy and it's not aware of the fact that there is an exit condition within the GrrProgram class. It seems that simply single-process mode with multiple tests was never used, even though the program seems designed to handle this situation, but the implementation is failing. Please [take a look at the code](https://github.com/google/grr/blob/master/run_tests.py#L177) and inspect for yourself. This for loop goes through all the suites and it's in the single process path.

This PR works around this bug for now so that we can use single-process mode to run multiple tests, which is the intention indicated by the semantics of the previous code.

> It appears Travis cant run more than one test at the time due to resource issues, so we need to reduce the number of concurrent tests to one - currently running the tests inline is not really a supported mode of operation and the code specifically checks for processes == 1 and goes into that code path.

This PR works around the bugs of single-process mode and makes it possible to run multiple tests in single-process mode.

>  You claim that this is not reliable since some tests are exiting

I'm not claiming "some tests are exiting". I'm claiming *all* tests are exiting. This is the expected behavior for Python unittest TestProgram. [Here is the relevant python standard library code patch that defines the current intentional exit behavior for Python unittest TestProgram](https://bugs.python.org/file13790/unittest-no-exit.patch) which `GrrTestProgram` extends. The previous code in multiprocessing mode only ever worked because there was a peculiar set of bugs that was making the program behave correctly. Here they are:

* **The first bug**: In the previous code, the `GrrTestProgram` is expected to **not** exit upon completion. This is why it is called within a function which is called within a `for` loop which invokes it several times. Here is the relevant code snippet that illustrates we don't expect this to exit:

```
for test_suite in suites:
  RunTest(test_suite, stream=stream)                                                                                                                
```

This code snippet clearly doesn't assume `RunTest` to exit the python process. Otherwise it wouldn't have been a `for` loop.

* **The second bug**: In the previous code, `GrrTestProgram` ignores its test_suite parameter. However, the code is not aware of this. Here is the relevant code snippet that illustrates this:

```
def RunTest(test_suite, stream=None):                                                                                                                   
  out_fd = stream

  test_lib.GrrTestProgram(argv=[sys.argv[0], test_suite],
                          testLoader=GRREverythingTestLoader(
                              labels=flags.FLAGS.labels),
                          testRunner=unittest.TextTestRunner(
                              stream=out_fd))
```

By reading this function, you can see that `test_suite` is expected to be the name of the test that we are about to run. This expectation is also clear from the signature of the function `RunTest`:

```
RunTest(test_suite, stream=None)
```

By reading this function signature and with no docstrings provided, it is reasonable to assume that this function runs the test specified in the `test_suite` parameter. Instead, the behavior is to ignore the `test_suite` parameter.

* **The third bug**: In the previous code, `GrrTestProgram` reads directly from `flags`. This behavior is unexpected and magical. Flags are incorrectly used as global variables. For the same reason [global variables are bad](http://c2.com/cgi/wiki?GlobalVariablesAreBad), this use of flags is bad and makes things hard to debug. This should be treated like a bug and fixed. Either `GrrTestProgram` must be invoked at the top-level and left to handle `flags` itself, or it should be invoked as a runner within the `RunTest` function and pay attention to its parameters without magically reading from `flags`.

In the end, the use of a `TestProgram` is not appropriate at this place at all. I will investigate more into alternatives of refactoring this, either using `nose` or by refactoring the existing use of `GrrTestProgram`.

>  so I am suggesting we stick to the multiprocess mode since it is the current way things work and will require minimal new code.

No. Multiprocess mode doesn't work on travis. I once again point you to the [relevant tests I ran](https://s3.amazonaws.com/archive.travis-ci.org/jobs/45939849/log.txt) which clearly illustrate that multiprocess mode will not work in this case. It may work with some effort, but this effort may require refactoring and is certainly not "minimal" in code changes. Single process mode, the way I worked-around it in this PR, works. I don't think the code changes made in this PR are that humongous. We can refactor later to use multiprocessing mode if you insist.

> All the grr test programs are stand alone (i.e. they have their own main()) and are actually designed to run stand alone. This is the best practice for running tests because you ensure that you have test isoltion when you start a new process each time (you can not guarantee this if you dont because we have a lot of shared globals like config library etc). In Rekall we actually spawn the test binary itself but I think in GRR we fork (maybe we should spawn).

> Not all tests should run as a separate program but test suites (which are collections of tests) should.

OK, sounds good, but this is not the intention of this PR.

> I think changing the test framework is probably outside the scope of this cl.

Agreed.

> I just suggested to remove that processes == 1 condition from the top and just see if that improves things on travis by running the test program in a child process.

This sounds like an experiment worthy exploring. Let's do that in a different PR. Could you open an issue?

> Note that tests can fail also due to a seg fault and catching the exception wont help here.

I agree in principle, but this is out of the scope of this PR. This PR's purpose is to get our tests running on travis, one way or another. Optimizing for performance or extreme cases such as segfaults can be done later. In fact, [I have verified that running all the tests without exclusions does not cause any segfaults currently](https://s3.amazonaws.com/archive.travis-ci.org/jobs/46216964/log.txt), even though it causes expected and unexpected failures which are handleable, so, while this may not be the best course of action and I see your point in desiring isolation, it's something we can live with for a while and segfaults can be regarded as non-existing extremities at this stage.

scudette: The single test mode is designed for debugging. When you run the test suite and it tells you test XXYY is failing, then you just run the same with the --test XXYY and then it runs the test inline. You can then put break points which work because its running inline (if the subprocess is forked you cant interact with it since it has no terminal). I dont  think anyone really runs it on multiple test cases inline. It may be that it does not work when given multiple test cases due to the loop exiting on the first test - it is only ever run on one test at the time. If this is a bug lets fix it rather than working around a bug in a hacky way.

Multiprocess mode does work fine on Travis - we use it in Rekall without trouble. We use a similar test runner to the GRR one in Rekall (It might have been slightly updated, im not sure). Running each test suite in its own process is the right way to go rather than running all tests in the same process. Note that one of the nice things about running in multiprocess mode is that we can separate the output of the test into different files and then the runner can just present a green or red overview for a failed test. This output is much more useful then the output you show. Check this for example:
https://travis-ci.org/google/rekall/builds/49599483

In Rekall we also get OOM failures and we are thinking how to solve these. It happens rarely but when we have so many tests it means one random test might fail. We might re-run the test in the case of OOM at least a couple of times. This seems to be a limitation travis as a platform. 

I agree you may have discovered some bug, that's great. Lets fix them and move on.

If the purpose of the CL is to run the test suite on travis, I can not consider it complete until all the tests are passing on travis. I dont care too much if you run them in multiprocess or single process mode (this sounds to me like a limitation of the platform) and if you find bugs in the existing code that's great - lets fix those. The output you show indicates some tests are legitimately failing - we should fix them to ensure that they all pass.

run_tests.sh
dionyziz: > disabling tests is not really helping here.

It's helping because you can see that this limited set of tests that is enabled is passing or failing. And the new tests that are checked in are passing or failing. You may still need to run the rest of the test suite manually to check the rest of the tests, but this provides a basic level of automation, which is useful. In particular, we can mindlessly ask external contributors to "fix the build" when we see red, without having to go into any manual labor. I think there's value in this, especially as we start welcoming additional external contributors.

> What exactly is the goal of this CL? 

The goal of this pull request is as follows:

1. To investigate the travis environment and the requirements for using travis as a build system for continuous integration. To look into its resource limits, dependencies, to time its behavior and to understand how we need to adapt to be able to run our build on it.
2. To define a clear path of what course of action will be required to achieve our quarter-goal of getting Travis fully integrated.
3. To enable the travis build system gradually in order to be able to automate some of the review process for external contributors and internal contributors who are using the PR system to submit review requests.
4. To increase the confidence in our open source build for us and for third parties in a way that is easily accessible and verifiable with a boolean indicator.
5. To make our build process more transparent to the open source community by illustrating the way tests are supposed to run.
6. To highlight the shortcomings of our current build process and to acknowledge that we are not perfect and we're willing to improve our coverage.

Just because the open source tests work for you in your machine with specific resources, specific dependencies pre-installed, and a specific version of your operating system doesn't mean they can run for everyone. One of the points of open source is to make things available to others as well, and googlers just claiming the test suite works for them is not enough. The migration to travis clearly points out the environment and configuration needed to run the tests.

> If the goal is to get travis build/testing working then it means all tests must pass.

This is not a black and white, all-or-nothing issue. All non-trivial software migration happens gradually.

> If you are saying that not all tests are passing then it means this CL is not complete and we should wait until its finished.

[Like Netscape 6.0]?

> just to have that undone in the next CL

Why would anything be undone? The next CLs will slowly improve our coverage in the way planned, until we are able to run all the tests.

> It feels sloppy and there is no reason for it other than shortness of time.

There are several reasons as indicated above. The most striking one is review rejection automation.

>  If indeed you suggest we adopt a policy of not accepting external PRs which break the build (i.e. they introduce bugs that affect existing tests) that means that those PRs are not finished either. Do you suggest we accept PRs that say - look at this new feature. I know it breaks some tests but i will fix those in the next PR. Clearly not.

No. Here's what I suggest: A PR that breaks the build should never be merged and it should be dismissed with a simple "Please fix the build" comment without further ado. A PR that passes the build should be reviewed and the reviewer should inspect the readability and semantics of the code and run the tests manually.

> It is concerning that tests are being killed due to OOM - does this mean travis as a platform is too small for us to run on?

No. The tests [are able to run just fine in single-process mode without running out of memory](https://s3.amazonaws.com/archive.travis-ci.org/jobs/46216964/log.txt). This is an indicator that travis is suitable for the open source build. Regardless of that matter, I don't think using travis is being disputed as a decision.

scudette: These are worthy goals and I agree with them. It does not appear that this CL achieves those goals:

1. Do you know the resource limitation? we are turning off most of the tests - maybe we have just enough resources to run helloworld.py and no more? 

Note that I was not suggesting running more that one test concurrently (as it is done by the multiprocessing mode) but running them in their own process each. This should have the same memory requirements as running them in their own process with the advantage that you have a new global state each time (more manageable for tests).

2. What course of action will be required to get the tests to pass? Why not just do it now? It is still not clear by this cl because we dont really know why the tests fail since you turned them off.

3. We can illustrate the way tests are supposed to run by documenting it - on a normal machine with normal amount of ram the tests pass fine. It seems to me that you are arguing that we should change the entire test mechanism to fit a memory constraint setup which is specially only applicable to travis. This decision can not be taken lightly - we must consider what we are giving up (nice formatting of test output, multiprocessing support etc). 

If one has to run the tests manually anyway for a PR since most tests are switched off then automation is not an issue here. At this stage we dont have enough PRs coming in that we would benefit in automatically turning away a PR without review just because it breaks some tests.

Your link is showing that some tests are failing due to installation problems. Its also very hard to read because it includes debug output of all the tests - even passing ones. It is not useful at this stage since if a PR comes in that breaks something it would be very hard for someone to find exactly what was broken and why.

BTW there are other options than travis that we should explore - if travis is simply not suitable for running our tests we should not settle for it. Is it possible to get a larger memory allocation on  travis - that's also another option.

dionyziz:
@scudette I will see if some more tests can be enabled to satisfy you.

dionyziz:
For the record, based on today's face to face discussion with @scudette, some points were made clear and we decided on the following:

1. Multiprocessing mode is better because it provides better, cleaner output and test isolation. While testability and isolation should be ensured at the test-level, isolation using separate processes is beneficial. We should use it.
2. Existing unit tests constitute good tests for the environment change to travis; failing tests are indicators that the environment change is incomplete, and therefore the pull request is incomplete. Therefore, all of them should be fixed before it is merged.
3. Build times on travis are high and we should consider the commercial options if these allow us to use pre-created system images.
4. Small bugs such as early exiting in the test libraries should be fixed in this pull request if possible.

So, I'm going to follow up with some changes in this pull request based on the above points.

dionyziz:
All taken care of. PTAL @scudette @grrrrrrrrr.

run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+cd ..
+echo 'Running tests...'
+
+EXCLUDE_TESTS=HTTPDataStoreCSVBenchmarks,HTTPDataStoreBenchmarks,MicroBenchmarks,TDBDataStoreBenchmarks,FakeDataStoreBenchmarks,AverageMicroBenchmarks,SqliteDataStoreBenchmarks,DataStoreCSVBenchmarks,TDBDataStoreCSVBenchmarks,AFF4Benchmark
```
scudette: I assume that these are excluded since they are just benchmark tests? They would pass if included?

scudette: Dont the benchmark tests get excluded anyway when running tests with the small label?

scudette:
This looks good! LGTM. I also saw the travis page and it looks very useful. Does it print the backtrace inline for a failure?

run_tests.sh
dionyziz: Yes, I'm excluding them because they are just benchmark tests. They pass if included now. I don't know why, but they are not excluded anyway if the label specified is "small". I don't know why the label exclusion doesn't work for them. Would you like me to investigate further? Perhaps the label directive is ignored for some reason.

scudette: Yeah that might be a bug we need to check. You can do it in another PR -
lets just get this one in.

Thanks. This is awesome!

dionyziz:
Yes; [here's how a failure looks like](https://travis-ci.org/dionyziz/grr/builds/51985340).

mbushkov:
Just checking - by looking at the output it seems that Selenium and JS tests are not supported yet, right?

scudette:
No. selenium tests probably can not run on travis due to resource
constraints.

PR #121, opened 2015-02-23
Added reason to access logging

pidydx:


pidydx:
Wow, that was fast.  Thanks!

PR #130, opened 2015-03-03
Added parser for Launchd plists

pidydx:
Reworked Launchd plist parser for OSXLaunchDaemons and OSXLaunchAgents artifacts with unit tests and everything.


parsers/osx_file_parser.py
```diff
@@ -53,3 +54,115 @@ def Parse(self, cmd, args, stdout, stderr, return_val, time_taken,
...
+      plist = binplist.readPlist(file_object)
+    except binplist.FormatError:
+      plist["Label"] = "Could not parse plist."
+
+    # These are items that can be directly copied
+    for key in direct_copy_items:
+      kwargs[key] = plist.get(key)
+
+    # These are arrays of strings that need to be combined
+    for key in string_array_items:
```
grrrrrrrrr: The string array type should be a repeated string in the protobuf definition. Then you can just directly assign the list to the rdfvalue:

if isinstance(elements, list):
  kwargs[key] = elements
# Can it actually happen that this is not a list?
else:
  kwargs[key] = [elements]

pidydx: # Can it actually happen that this is not a list?
Let me tell you about the documentation for launchd plists... A whole bunch of the values get defined as this could be X, or Y, or maybe Z, good luck! So I don't think it will ever be something other than a list, but I totally don't trust it to not be.

The flag_only_items are largely because trying to write a protobuf that accounts for all the possible structures was going to get fairly ridiculous.

I can go ahead and change those things to repeated and move them up to direct copy.

pidydx:
Does this look better?

pidydx:
Strike that.  Don't merge yet.  That question you asked...yes...sometimes it is an array and sometimes it is a string so I need to properly check those.  Fix coming soon.

pidydx:
This handles things correctly now.

destijl:
@grrrrrrrrr to finish off the review here.

grrrrrrrrr:
Looks good, let's merge this.

PR #131, opened 2015-03-04
Removed unecessary forced regex

pidydx:
This forced modification of the regex causes a brutal performance hit (7s queries vs 0.3s queries in the tested example).  It would be better to send the regex as desired in any calls to this.  This appears to be what it is doing in filestore_stats.py, but I'm not clear on all of the other places where this is getting called.  This passes unit tests and I am going to begin testing it in a live test environment, but any heads up on other places where this change needs to be accounted for would be appreciated.

### DIFF
--- a/lib/aff4_objects/filestore.py
+++ b/lib/aff4_objects/filestore.py
@@ -200,7 +200,7 @@ class FileStoreImage(aff4.VFSBlobImage):
       index.
     """
     # Make the regular expression.
-    regex = ["index:target:.*%s.*" % target_regex.lower()]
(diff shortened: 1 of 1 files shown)


grrrrrrrrr:
Yeah I did some testing too and it seems like we pass a proper regex already / don't really use this function at all. I'll merge it in, we can always find a more performant solution later in case this actually leads to problems at some point.

pidydx:
Awesome!  The regex I use for trying to break up and avoid regexes is already kind of scary so the cleaner attribute regexs make it easier to convert rlike to like where possible.

PR #136, opened 2015-03-10
RHEL Repack fix

pidydx:
This line breaks rpm repacking on RHEL.  If this is necessary for proper behavior on other server platforms I will try to work out specifically why there is a difference in behavior.

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -786,7 +786,6 @@ class CentosClientDeployer(LinuxClientDeployer):
           rpmbuild_binary,
           "--define", "_topdir " + rpm_root_dir,
           "--target", client_arch,
-          "--buildroot", rpm_build_dir,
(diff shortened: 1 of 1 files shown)


destijl:
Blarg, it wouldn't work on centos without this.  You can test with the vagrant image to see what I mean.  With your previous spec file it was copying content into a subdirectory of itself.  Removing that change and specifying buildroot fixed it.  At least this way you can build on the centos VM we provide.  It would be good to make it work on both though if you can.

pidydx:
I will look into it again.  If I remember right I had to make the changes because rpmbuild blows away that directory expecting the build process to populate it with compiled stuff, but this process just copies python stuff into that dir before "build" happens so then rpmbuild empties it and fails.

PR #140, opened 2015-03-17
Added parser for crontab artifacts

pidydx:
This will also serve as the starting point for future parsers relating to Linux at jobs and Windows Scheduled Tasks.  This uses the artifacts created/modified in https://github.com/ForensicArtifacts/artifacts/pull/24

[Review on Reviewable](https://reviewable.io/reviews/google/grr/140)


destijl:
This adds a dependency on the crontab library, which needs to go into requirements.txt

pidydx:
Fixed.  I should have caught that too since I spent 20 minutes trying to figure out why it wasn't working in my test environment after I forgot to install that.

proto/sysinfo.proto
```diff
@@ -479,3 +479,23 @@ message HardwareInfo {
     }];
 
 }
+
+// CronTab Entry
+message CronTabEntry{
```
destijl: make the numeric ones ints?

test_data/crontab
```diff
@@ -0,0 +1 @@
+* * * * * /usr/bin/echo "test"
```
destijl: use a unique value for each piece here and verify things get set correctly in the test

parsers/cron_file_parser.py
```diff
@@ -0,0 +1,39 @@
...
+  def Parse(self, stat, file_object, knowledge_base):
+    """Parse the crontab file."""
+    _ = knowledge_base
+    entries = []
+
+    crondata = file_object.read(100000)
+    jobs = CronTab(tab=crondata)
+
+    for job in jobs:
+      minute = utils.SmartStr(job.minute)
```
destijl: you don't need all these temp variables

proto/sysinfo.proto
pidydx: I thought about that, but it is possible that they aren't always ints due to the allowed special chars

parsers/cron_file_parser.py
pidydx: Initially did that for clarity, but I don't think it adds much so I will remove.

pidydx:
Should be good now

proto/sysinfo.proto
destijl: ah, yep, of course.

parsers/cron_file_parser_test.py
```diff
@@ -0,0 +1,43 @@
...
+  """Test parsing of cron files."""
+
+  def testCronTabParser(self):
+    """Ensure we can extract jobs from a crontab file."""
+    parser = cron_file_parser.CronTabParser()
+    client = "C.1000000000000000"
+    results = []
+
+    path = os.path.join(self.base_path, "crontab")
+    plist_file = open(path)
```
destijl: crontab_file, I'm guessing this is is a copy paste error.

parsers/cron_file_parser_test.py
```diff
@@ -0,0 +1,43 @@
...
+    stat = rdfvalue.StatEntry(
+        aff4path=rdfvalue.ClientURN(client).Add("fs/os").Add(path),
+        pathspec=rdfvalue.PathSpec(path=path,
+                                   pathtype=rdfvalue.PathSpec.PathType.OS),
+        st_mode=16877)
+    results.extend(list(parser.Parse(stat, plist_file, None)))
+
```
destijl: check length of results, so that an empty list won't pass

pidydx: What is the best way to handle that? Just assert on a length of 1?

parsers/cron_file_parser_test.py
pidydx: the files are named cron_file_parser because ultimately windows sched tasks and *nix at jobs will live in here and they are "cron" things that aren't crontabs.

destijl: Wasn't clear: I'm referring to renaming plist_file to something else, since it isn't a plist.

parsers/cron_file_parser_test.py
destijl: yep

pidydx: Done

destijl:
This looks good now, I'm going to hold off on merging for a little bit until I can go through the process to add the extra dependency internally.

pidydx:
Updated to resolve merge conflicts with master.

PR #142, opened 2015-03-18
RPM Build fixes

pidydx:
This should fix the build issues.  The expectation is that %install empties BUILDROOT and then copies BUILD into BUILDROOT (typically using make install).  Because these were being used as the same dir RHEL blows everything away and then fails to pack.  This change should assemble the client in BUILD as normal and then declare BUILDROOT as the target and then the cp command replaces what make install would normally do.

### DIFF
--- a/config/centos/grr.spec.in
+++ b/config/centos/grr.spec.in
@@ -19,6 +19,7 @@ URL: https://github.com/google/grr
 # Do nothing here.
 
 %install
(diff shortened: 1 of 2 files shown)


destijl:
I tested this on centos and it works great.  Thanks for fixing.  I'll merge when I get a chance to do another export first.

pidydx:
Yay!  Merge all the things! _o/

PR #148, opened 2015-04-06
Add command file for batch mode

nawien-sharma:


tools/console.py
```diff
@@ -73,6 +73,9 @@
                     "the flag is run instead (comparable to the -c option of "
                     "IPython).")
 
+flags.DEFINE_string("command_file", None,
+                    "If present, no console is started but the code given in "
+                    "command file supplied as input ")
```
destijl: "but the code in command_file is executed."

tools/console.py
```diff
@@ -122,6 +125,9 @@ def main(unused_argv):
   if flags.FLAGS.code_to_execute:
     logging.info("Running code from flag: %s", flags.FLAGS.code_to_execute)
     exec(flags.FLAGS.code_to_execute)  # pylint: disable=exec-used
+  elif flags.FLAGS.command_file:
+    logging.info("Running code from flag: %s", flags.FLAGS.command_file)
```
destijl: s/flag/file/

nawien-sharma: correct should be file

PR #149, opened 2015-04-09
Fix bug with "never" policy in fuse_mount

darrenbilby:


darrenbilby:
travis failures were due to unrelated artifact dependency changes

destijl:
LGTM

PR #150, opened 2015-04-10
Restored RPM i386 repacking

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/150)


grrrrrrrrr:
Looks good, thanks!

dionyziz:
@grrrrrrrrr For the sake of clarity in history, it may be worth asking for a rebase instead of repeated merge commits, especially as we work towards a more open development process.

pidydx:
Previously I was told to just merge, but I can do this another way to keep things clean.  I can also just leave them as is if there are no conflicts if that is preferred.

PR #151, opened 2015-04-12
ACL Token expiry

nawien-sharma:
config_lib is not instantiated except for constants when LIFETIME is defined.  So, ACL.token_expiry cannot override LIFETIME and thus LIFETIME is not used in CheckAccess().

### DIFF
--- a/config/server.py
+++ b/config/server.py
@@ -101,6 +101,9 @@ config_lib.DEFINE_list("Cron.enabled_system_jobs", [],
 config_lib.DEFINE_integer("ACL.approvers_required", 2,
                           "The number of approvers required for access.")
 
+config_lib.DEFINE_integer("ACL.token_expiry", 480,
+                          "The duration in minutes of a valid approval token.")
+
 config_lib.DEFINE_string("Frontend.bind_address", "::",
                          "The ip address to bind.")
 
--- a/config/server.py
+++ b/config/server.py
@@ -101,7 +101,7 @@ config_lib.DEFINE_list("Cron.enabled_system_jobs", [],
 config_lib.DEFINE_integer("ACL.approvers_required", 2,
                           "The number of approvers required for access.")
 
-config_lib.DEFINE_integer("ACL.token_expiry", 480,
+config_lib.DEFINE_integer("ACL.token_expiry", 10080,
                           "The duration in minutes of a valid approval token.")
 
 config_lib.DEFINE_string("Frontend.bind_address", "::",
--- a/config/server.py
+++ b/config/server.py
@@ -101,8 +101,8 @@ config_lib.DEFINE_list("Cron.enabled_system_jobs", [],
 config_lib.DEFINE_integer("ACL.approvers_required", 2,
                           "The number of approvers required for access.")
 
-config_lib.DEFINE_integer("ACL.token_expiry", 10080,
-                          "The duration in minutes of a valid approval token.")
+config_lib.DEFINE_integer("ACL.token_expiry", 604800,
+                          "The duration in seconds of a valid approval token. Default of one week.")
 
 config_lib.DEFINE_string("Frontend.bind_address", "::",
                          "The ip address to bind.")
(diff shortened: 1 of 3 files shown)


config/server.py
```diff
@@ -101,6 +101,9 @@
 config_lib.DEFINE_integer("ACL.approvers_required", 2,
                           "The number of approvers required for access.")
 
+config_lib.DEFINE_integer("ACL.token_expiry", 10080,
```
pidydx: This should probably be defined in seconds and defaulted in a similar way to the statsstore TTL with 60 * 60 * 24 * 7

lib/aff4_objects/security.py
```diff
@@ -151,10 +151,11 @@ class ApprovalWithApproversAndReason(Approval):
 
   class SchemaCls(Approval.SchemaCls):
     """The Schema for the ClientAccessApproval class."""
+
     LIFETIME = aff4.Attribute(
         "aff4:approval/lifetime", rdfvalue.RDFInteger,
-        "The number of microseconds an approval is valid for.",
-        default=4 * 7 * 24 * 60 * 60 * 1000000)  # 4 weeks
+        "The number of seconds an approval is valid for.",
```
pidydx: This should remain microseconds because all of the timestamps and comparisons are in microseconds.  This is probably where it should be getting the value out of the config.

lib/aff4_objects/security.py
```diff
@@ -192,7 +193,9 @@ def CheckAccess(self, token):
       return True
 
     # Check that there are enough approvers.
-    lifetime = self.Get(self.Schema.LIFETIME)
+    lifetime = config_lib.CONFIG["ACL.token_expiry"]
```
pidydx: If the default is defined in seconds and the changes are handled in the Schema it should be unnecessary to override the use of the schema or convert to seconds here.

lib/aff4_objects/security.py
destijl: I think this should probably be an rdfvalue.Duration so you can specify "1w" and it will handle seconds/microseconds properly in comparison and arithmetic operations.  You'll have to test what happens to old approvals that are stored with the old format though.

destijl: Sean is right too, this is where it should get the config_lib default.

nawien-sharma: I agree that config_lib.CONFIG["ACL.token_expiry"] should override the default hard coding but I was seeing an exception thrown that config_lib.CONFIG["ACL.token_expiry"] was not loaded yet. A sequence/timing issue ? So, in CheckAccess(), I removed the reference to LIFETIME and used the config_lib variable instead since it has a default value defined in server.py. If we still want to use the LIFETIME attribute, then I need to resolve the original exception.

destijl: oh you're right, it needs to be defined at import time when the class is declared.  I think the best course of action would be to get rid of the LIFETIME attribute altogether and just use the config variable.  A quick grep of the code shows this attribute isn't used anywhere else.  I suspect the intent was to allow granting of customizable approval times, but we've never used it that way.  We do lose the possibility of future flexibility if all approvals are hard-coded to use the config variable though. @grrrrrrrrr what do you think?

grrrrrrrrr: We do use this lifetime (at least we try, I'm not sure it actually works) in CreateApprovalRaw, it would make sense to keep this functionality. Why don't we set the default here to 0 and below we do

lifetime = self.Get(self.Schema.LIFETIME) or config_lib.CONFIG["ACL.token_expiry"]

That way we can have the best of both worlds.

Also, I agree that this should be kept in microseconds everywhere, adding conversions somewhere in the middle of the code just before we use the value seems a bit error prone.

nawien-sharma:
I have incorporated the comments above to set default for LIFETIME to 0 and to config_lib.CONFIG["ACL.token_expiry"] to seconds.

grrrrrrrrr:
Sorry that this took a bit, this looks good now!

PR #152, opened 2015-04-13
Added config option to allow deletion of hunts with results from UI

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/152)


config/server.py
```diff
@@ -264,3 +264,6 @@
                        "processed recursively depth-first, later values "
                        "override earlier ones.")
 
+config_lib.DEFINE_bool("AdminUI.protect_hunt_results", default=True,
```
destijl: I think I'd call this "AdminUI.allow_hunt_results_delete", default=False

lib/hunts/standard.py
```diff
@@ -154,6 +154,7 @@ class DeleteHuntFlow(flow.GRRFlow):
 
   @flow.StateHandler()
   def Start(self):
+    protect_results = config_lib.CONFIG["AdminUI.protect_hunt_results"]
```
destijl: you don't need this temporary variable.

config/server.py
pidydx: I did this for the logic later of if X and Y for the delete.  I can change that to if not X and Y and make this change if you prefer.

lib/hunts/standard.py
pidydx: Similar to above I mostly did that to make the if X and Y line easier to read.  I can just use it directly if that is preferred

destijl:
one minor comment, this can go in with the next merge.

---





<sup>**[config/server.py, line 272 \[r2\]](https://reviewable.io:443/reviews/google/grr/152#-Jo0Lz9TYWh1y_SubrMK)** ([raw file](https://github.com/google/grr/blob/687e711e1a09e475cff3c9009d8527ee06f059bb/config/server.py#L272)):</sup>
remove "will"

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/152)

PR #165, opened 2015-04-23
Added unit test

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/165)

### DIFF
--- a/lib/aff4_test.py
+++ b/lib/aff4_test.py
@@ -435,8 +435,6 @@ class AFF4Tests(test_lib.AFF4ObjectTest):
     """Run basic tests on a subclass of AFF4ImageBase."""
     path = "/C.12345/aff4image"
 
-    config_lib.CONFIG.Set("Server.max_unbound_read_size", 1)
-
     fd = aff4.FACTORY.Create(path, classname, token=self.token)
     fd.SetChunksize(10)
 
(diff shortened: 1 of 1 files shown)


lib/aff4_test.py
```diff
@@ -435,6 +435,8 @@ def ExerciseAFF4ImageBase(self, classname):
     """Run basic tests on a subclass of AFF4ImageBase."""
     path = "/C.12345/aff4image"
 
+    config_lib.CONFIG.Set("Server.max_unbound_read_size", 1)
```
destijl: put this down with the test code, so it's more obvious what's happening.

lib/aff4_test.py
```diff
@@ -471,6 +473,9 @@ def ExerciseAFF4ImageBase(self, classname):
     self.assertEqual(len(data), 300)
     self.assertTrue("XXXHello WorldXXX" in data)
     self.assertTrue("XXXYYY" in data)
+    fd.seek(0)
```
destijl: add a comment about what you're testing and also test the non-failure case that you get 1byte back from a read.

pidydx:
I shuffled the whole thing around and do both tests now.

lib/aff4_test.py
```diff
@@ -471,6 +471,25 @@ def ExerciseAFF4ImageBase(self, classname):
...
     self.assertTrue("XXXYYY" in data)
+    fd.Close()
+
+    # Set the max_unbound_read_size to last size of object at path
+    # before object creation for unbound read() tests.
+    config_lib.CONFIG.Set("Server.max_unbound_read_size", 10100)
+
+    fd = aff4.FACTORY.Create(path, classname, mode="rw", token=self.token)
+    fd.Set(fd.Schema._CHUNKSIZE(10))
+    # Write one byte and verify the unbound read returns 1 byte
```
grrrrrrrrr: Nit: Comment seems wrong.

Issue #158, opened 2015-04-17
grr_config_updater update_user can create a user


pidydx:
If you attempt to update_user with a username that doesn't exist it will create the user for you w/o a password.  This command should probably exit with an error if the user does not already exist rather than creating a new user and then updating it.

pidydx:
This fix is complete and waiting for merge.

PR #166, opened 2015-04-24
Separated AddUser and UpdateUser for proper behavior

pidydx:
This addresses this issue: https://github.com/google/grr/issues/158
It also fixes add_user so that you are not forced to type your password on the command line and extends add_user to allow creation with labels.
An existing user cannot be overridden by accident with a call to add_user as well.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/166)


tools/config_updater.py
```diff
@@ -100,15 +105,39 @@
...
+  with aff4.FACTORY.Create("aff4:/users/%s" % username,
+                           "GRRUser", mode="rw", token=token) as fd:
+    # Note this accepts blank passwords as valid.
+    if not password:
+      password = getpass.getpass(
+          prompt="Please enter password for user '%s': " % username)
+    fd.SetPassword(password)
+
+    # Use sets to dedup input.
+    if labels is not []:
```
grrrrrrrrr: This whole block can be done as
if labels:
  fd.AddLabels(*labels, owner="GRRR")

tools/config_updater.py
```diff
@@ -100,15 +105,39 @@
...
   """Implementation of the update_user command."""
-  print "Updating user %s" % username
-  with aff4.FACTORY.Create("aff4:/users/%s" % username,
+
+  with aff4.FACTORY.Open("aff4:/users/%s" % username,
                            "GRRUser", mode="rw", token=token) as fd:
+
     # Note this accepts blank passwords as valid.
-    if password is not None:
+    if password:
```
grrrrrrrrr: The comment above says it accepts blank passwords. I think we should allow "" so please change back to "is not None".

tools/config_updater.py
```diff
@@ -479,9 +510,7 @@ def Initialize(config=None, token=None):
   startup.ConfigInit()
 
   print "\nStep 3: Adding Admin User"
-  password = getpass.getpass(prompt="Please enter password for user 'admin': ")
-  UpdateUser("admin", password, ["admin"], token=token)
-  print "User admin added."
+  AddUser("admin", ["admin"], token=token)
```
grrrrrrrrr: This sets password to ["admin"]. Please call named arguments by name (labels=["admin"]) and I'd also suggest you add a default of None to the password parameter so you don't have to do AddUser("admin", None, labels=...)

tools/config_updater.py
```diff
@@ -553,26 +582,29 @@ def main(unused_argv):
...
     labels = []
     if not flags.FLAGS.noadmin:
       labels.append("admin")
-    UpdateUser(flags.FLAGS.username, password, labels, token=token)
+
+    if flags.FLAGS.labels:
+      labels.extend(flags.FLAGS.labels)
+
+    try:
+      if aff4.FACTORY.Open("aff4:/users/%s" % flags.FLAGS.username, "GRRUser",
```
grrrrrrrrr: Should we have this check (and also the one for above checking if the user doesn't exist) inside the AddUser/UpdateUser functions?

tools/config_updater.py
```diff
@@ -646,3 +678,6 @@ def main(unused_argv):
 
 if __name__ == "__main__":
   flags.StartMain(main)
+  # Wait for threads to finish inserting into the datastore
+  time.sleep(.5)
```
grrrrrrrrr: Please call data_store.DB.Flush() instead.

tools/config_updater.py
pidydx: I had them inside originally and wound up moving them out, but now I'm not entirely sure why. I think they can probably get moved back in now.

tools/config_updater.py
pidydx: I had originally planned on doing that, but what I noticed was that Flush() had already  been called so calling Flush() wouldn't solve the problem.  The flusher thread is already doing the database inserts when the interpreter starts shutting down and pulls the rug out of under it. The exceptions I was occasionally seeing with config_updater previously were in the datastore in Execute() when something went to None before it finished inserting.

tools/config_updater.py
pidydx: I had the checks outside to leave the with construct alone and allow initialize to always overwrite the admin user while preventing command line usage from doing overwrites with AddUser.  Would it be better to get rid of the with to do the try/except inside Add/UpdateUser?  What would be the best way to allow initialize to overwrite "admin" with AddUser while preventing it from command line usage?

tools/config_updater.py
pidydx: Good catch, thanks!

tools/config_updater.py
pidydx: The logic around this one changed a bit since I moved the prompting for a password.  
1.  UpdateUser does not force a password update so you can add/remove labels w/o resetting passwords.
2.  UpdateUser uses --password as a flag to trigger a password prompt rather than --password="Foo" so you don't have to type your password on the command line.  

So if the password value is true it prompts and if false it skips.

pidydx: I did change the behavior in AddUser as a result of this note though so AddUser should accept --password=""

grrrrrrrrr: Ok cool but the comment is definitely wrong now.

tools/config_updater.py
grrrrrrrrr: We should have the checks inside the functions and have them print meaningful errors. Would just give a better user experience. For the init, maybe we can just test if the user already exists? If we did, we could maybe even go a step further and ask if we should change the password, in most cases I don't really want to do that anyways. What do you think?

pidydx:
Checks have been moved inside the functions and added handling to prompt for admin reset if it already exists.

destijl:
@grrrrrrrrr can you finish this review?

grrrrrrrrr:
LGTM, thanks!!

PR #172, opened 2015-05-05
Lowered severity of PyInstaller import error

pidydx:
This isn't really an error for the majority of cases where build.py is getting loaded so lowering to a warning to clear up logs.  New users have also reported feelings of confusion and doubt when encountering this error.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/172)

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -22,7 +22,7 @@ import distorm3
 try:
   from PyInstaller import main as PyInstallerMain
 except ImportError:
-  logging.error("PyInstaller missing")
+  logging.warn("PyInstaller missing")
 
 from grr.lib import config_lib
 from grr.lib import rdfvalue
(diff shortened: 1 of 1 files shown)


lib/build.py
```diff
@@ -22,7 +22,7 @@
 try:
   from PyInstaller import main as PyInstallerMain
 except ImportError:
-  logging.error("PyInstaller missing")
+  logging.warn("PyInstaller missing")
```
destijl: I think we should just make this pass with no message, and let it blow up if someone actually tries to build a client without it.  Add a comment like:

```
# We ignore this failure since most people running the code don't build their own clients and printing an error message causes confusion.  Those building their own clients will need PyInstaller installed.
```

destijl:
LGTM

---

Reviewed files:
- lib/build.py @ r2

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/172)

Issue #170, opened 2015-05-01
MySQLAdvanced Datastore ConnectionPool hang


pidydx:
We are seeing workers that are still running, but appear to stop doing work.  The last error before they appear to drop off of the Server Load active thread count is:

Trace: Traceback (most recent call last):
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_6-py2.7.egg/grr/lib/flow_runner.py", line 548, in RunStateMethod
    responses=responses)
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_6-py2.7.egg/grr/lib/flow.py", line 312, in Decorated
    auth_required=auth_required)
  File "/opt/grr/lib64/python2.7/site-packages/grr-0.3.0_6-py2.7.egg/grr/lib/flow.py", line 149, in __init__
    raise FlowError("No valid Status message.")
FlowError: No valid Status message.

This appears to be associated primarily Interrogate hunts and 3.0.0.3 clients.

pidydx:
I put a worker in debug mode and let it run for about an hour or two while busy and it eventually just stopped with this message.
DEBUG:2015-05-08 17:41:09,951 stats_store:736] Writing stats to stats store.

It had written to the stats store multiple times before it got hung up here and it doesn't give any errors and the process doesn't die.  Looking through the only thing I see that might be related to an infinite hang is something going wrong in the loops in stats_store.py around line 169 in WriteStats()

pidydx:
I added a bunch of debug statements and it seems to hang here in stats_store.py:L243 in WriteStats.
    process_data = aff4.FACTORY.Create(self.urn.Add(process_id),
                                       "StatsStoreProcessData",
                                       mode="rw", token=self.token)

I put debugging statements in aff4.FACTORY.Create and it hasn't reproduced in 12+ hrs now.  I don't see anything special here so I suspect that GRR is occasionally hanging up on this call elsewhere, but the stats store one is what hangs the workers.

Until I can get more information deeper in the error I suspect this may be a rare timing issue in MySQLAdvanced datastore, but I'm not clear what could cause aff4.FACTORY.Create to permanently hang rather than raise and error somewhere.  Any thoughts on what could cause this or where to look would be greatly appreciated.  Thanks!

pidydx:
Good news.  I found the bug (mostly).  It is in the MySQLAdvanced connector, but thankfully it does not appear to be related to any of the MySQL stuff directly.

[...]

PR #176, opened 2015-05-13
Sql pool fix

pidydx:
This is to address this issue: https://github.com/google/grr/issues/170
The Queue object uses Lock instead of RLock so heavy usage can cause a situation where the ConnectionPool deadlocks and causes infinite hangs and timeouts for GRR servers using the MySQLAdvanced datastore.

I am waiting on final results to see if this is fixed for good, but early testing indicates that this solves the problem.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/176)


lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -382,11 +400,32 @@ def _BuildInserts(self, values):
...
     return [aff4_q, subjects_q, attributes_q]
 
-  def _ExecuteTransaction(self, transaction):
+  def _ExecuteQuery(self, query, args=None):
     """Get connection from pool and execute query."""
+    retries = 10
+    for attempt in range(1, retries + 1):
+      connection = self.pool.GetConnection()
+      try:
+        connection.cursor.execute(query, args)
```
destijl: This whole thing could do with lots of stats.STAT counters and timers added to enable proper monitoring and give you visibility of query execution times, error rates, retries, aborts etc.

pidydx: I will try to look into that when I get some time.  Right now we mostly rely on the mysql logs, but being able to see some of that from the GRR server side could be nice.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -26,23 +26,48 @@ class Error(data_store.Error):
   """Base class for all exceptions in this module."""
 # pylint: enable=nonstandard-exception
 
+class SafeQueue(Queue.Queue):
+  def __init__(self, maxsize=0):
```
destijl: you're missing a call to super here.

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -382,11 +400,32 @@ def _BuildInserts(self, values):
 
     return [aff4_q, subjects_q, attributes_q]
 
-  def _ExecuteTransaction(self, transaction):
+  def _ExecuteQuery(self, query, args=None):
```
destijl: Is it going to cause you problems if I change this to ExecuteQuery (drop the underscore?) It's incorrect as is because you access it from outside the class.  I'm going to fix this and the super call when I merge it to the internal code base.

lib/data_stores/mysql_advanced_data_store.py
pidydx: I don't think I need it do I? I am completely overriding __init__ (It is literally copy/pasted from the parent class with Lock changed to RLock).

lib/data_stores/mysql_advanced_data_store.py
```diff
@@ -26,23 +26,48 @@ class Error(data_store.Error):
   """Base class for all exceptions in this module."""
 # pylint: enable=nonstandard-exception
 
+class SafeQueue(Queue.Queue):
+  def __init__(self, maxsize=0):
+      self.maxsize = maxsize
+      self._init(maxsize)
```
destijl: Oh god, I just noticed this.  Why are you doing this instead of calling __init__ with the parameter?

lib/data_stores/mysql_advanced_data_store.py
destijl: I should have read more closely.  Is this what you're trying to do?

```
Queue.Queue.__init__(self, maxsize=maxsize)
self.mutex = threading.RLock()
```
If so, lets do that, copy pasting a whole method is almost never a good idea.

lib/data_stores/mysql_advanced_data_store.py
pidydx: "I'm" not.  If you look in the Python Queue module that is an exact copy paste of the Queue objects init with self.mutex=threading.Lock changed to self.mutex=threading.RLock.

pidydx: If you look at the other methods of Queue and the subclasses it makes more sense as to why they did it that way

lib/data_stores/mysql_advanced_data_store.py
pidydx: I don't think that will work be cause immediately after self.mutext=threading.RLock() it starts using self.mutex for other initialization stuff.  I think the right way to fix this is for Queue itself to be using RLock, but patching python to make GRR work isn't much of an option.

lib/data_stores/mysql_advanced_data_store.py
pidydx: I forgot about the Transaction piece.  As long as it all gets refactored with the new name then I don't see any problem.  Though the super call is probably not a good idea :)

lib/data_stores/mysql_advanced_data_store.py
pidydx: Unless I don't understand right, if I self.mutex=threading.RLock() before super, then it will just set self.mutex=threading.Lock() and then call these setup functions.
```self.not_empty = _threading.Condition(self.mutex)
self.not_full = _threading.Condition(self.mutex)
self.all_tasks_done = _threading.Condition(self.mutex)```
If I set self.mutex=threading.RLock() after calling super it will have already set those 3 up using a self.mutex using threading.Lock()

destijl: Yeah you're right. I came up with a few alternatives but none of them are particularly good.  Discussing with andy, we'll figure out what to do on monday.

pidydx: Yeah, I stared at this one for a while before deciding to just override the entire init.  This was a really awful bug to dig out and test because you have to put the system under heavy load with a bunch of extra debug logging so you can see when it hits the deadlock.  This can take anywhere between 15 minutes and a few hours.  This fix isn't pretty, but it works.

PR #177, opened 2015-05-14
Config datastore

pidydx:
This change prompts for datastore configuration during initialization and then does the full init after it has the necessary configuration to properly initialize on the datastore of choosing.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/177)


destijl:
Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.
Reviewed files:
- config/grr-server.yaml @ r1
- tools/config_updater.py @ r1

---

<sup>**[tools/config_updater.py, line 452 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKwJ_SBYznS04yRdY9)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L452)):</sup>
I don't think specifying this as a default makes sense. It will never be correct.  This is one thing where the user really needs to tell us.

---

<sup>**[tools/config_updater.py, line 583 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKyyuTRwE0Xp0o5r-R)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L583)):</sup>
Did you test this on a clean machine? Doesn't ConfigureBaseOptions need to run before ConfigInit will work?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

pidydx:
Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tools/config_updater.py, line 452 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKwJ_SBYznS04yRdY9)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L452)):</sup>
This is mostly true.  There are two reasons for doing it this way.

1. It makes it look consistent with all of the other questions so the default isn't a valid default value as much as it is an example value instead of doing "e.g. grr.example.com".

2. It actually is a valid default when button mashing through the install to test pieces further in. 

---

<sup>**[tools/config_updater.py, line 583 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKyyuTRwE0Xp0o5r-R)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L583)):</sup>
ConfigInit will load what it finds in grr-server.yaml and server.local.yaml if it exists (If doing reinit).  This allows ConfigureBaseOptions to any current configs as default values for the prompts.  After ConfigureBaseOptions I call startup.Init() which itself calls ConfigInit again updating with whatever was configured in ConfigureBaseOptions.  This was necessary to get it to properly load everything necessary without initializing the datastore until after you choose your datastore in ConfigureBaseOptions.

I was able to test the behavior of properly initializing on the datastore of your choice using SQLite and MySQLAdvanced, but a second set of eyes on testing certainly wouldn't hurt.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

destijl:
Review status: 0 of 2 files reviewed, 2 unresolved discussions, some commit checks broke.

---

<sup>**[tools/config_updater.py, line 452 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKwJ_SBYznS04yRdY9)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L452)):</sup>
I don't think we should optimize for testing here.  Pointing grr at some random domain is not a good idea.  If the user really wants that they can do asdflkjsdaf.com

---

<sup>**[tools/config_updater.py, line 584 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKyyuTRwE0Xp0o5r-R)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L584)):</sup>
OK I think this is fine to go in once grr.example.com is removed.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

pidydx:
Review status: 0 of 2 files reviewed, 2 unresolved discussions, some commit checks pending.

---

<sup>**[tools/config_updater.py, line 453 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKwJ_SBYznS04yRdY9)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L453)):</sup>
No worries, I changed the wording of the question to the original.  

---

<sup>**[tools/config_updater.py, line 584 \[r1\]](https://reviewable.io:443/reviews/google/grr/177#-JqKyyuTRwE0Xp0o5r-R)** ([raw file](https://github.com/google/grr/blob/f233327a7b0febde4ecdac584b6ad751cb33df39/tools/config_updater.py#L584)):</sup>
Done

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

destijl:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[scripts/install_script_ubuntu.sh, line 131 \[r4\]](https://reviewable.io:443/reviews/google/grr/177#-Jt05esNZ-Y0svphuj4f)** ([raw file](https://github.com/google/grr/blob/4307e8fc5f1013afe51a0cbfa9962670d1f188be/scripts/install_script_ubuntu.sh#L131)):</sup>
We should install python sqlite here somewhere or presumably the configure will fail. Did you test on a clean machine?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

pidydx:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[scripts/install_script_ubuntu.sh, line 131 \[r4\]](https://reviewable.io:443/reviews/google/grr/177#-Jt05esNZ-Y0svphuj4f)** ([raw file](https://github.com/google/grr/blob/4307e8fc5f1013afe51a0cbfa9962670d1f188be/scripts/install_script_ubuntu.sh#L131)):</sup>
I was unable to properly test due to how the script pulls down the old deb that has hard requirements on Mongo.  However, now that I have thought about it this should be unnecessary anyways.  Any necessary sqlite pieces for the datastore will already be present to handle the Firefox history pieces and I'm pretty sure that all comes default anyways.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

destijl:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[scripts/install_script_ubuntu.sh, line 131 \[r4\]](https://reviewable.io:443/reviews/google/grr/177#-Jt05esNZ-Y0svphuj4f)** ([raw file](https://github.com/google/grr/blob/4307e8fc5f1013afe51a0cbfa9962670d1f188be/scripts/install_script_ubuntu.sh#L131)):</sup>
As you say the script fails because mongo is still a dependency on the actual server package.   I tried

--ignore-depends=mongodb,python-pymongo

but this is bad because any subsequent apt operations will complain until mongo gets installed. So effectively you won't be able to update the server. So Mongo should stay here for now until we have a chance to release a new server package without the dependency.  Add a comment like:

# Mongo is actually unused in the default configuration (sqlite is the default). We install it here to satisfy the grr-server deb package dependency. Remove this once we have built a new grr-server.deb without the mongo dependency.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/177)

pidydx:
So let it install, but then remove the piece that activates mongo so you don't have unnecessary services running?

destijl:
Is this ready for me to take another look? I'd like to get it in tomorrow if possible. I'm doing some work on the install scripts and would like to have this in.

pidydx:
Yes?  I'm not sure why I didn't get a notification on this message or why the unit tests are suddenly failing, but it should all be working.

pidydx:
So the problem here is that I added a global to stop repeated calls to ConfigInit() because calling that during initialize resets the values to what they were before the new config was written.  Preventing ConfigInit() from being called repeatedly causes the unit tests to fail.

I would expect ConfigInit() to properly reload the newly written secondary config, but I'm not sure why it isn't doing that.

lib/startup.py
```diff
@@ -36,6 +36,7 @@ def ConfigInit():
   """Initialize the configuration manager from the command line arg."""
   # Initialize the config system from the command line options.
   config_lib.ParseConfigCommandLine()
+  CONFIG_INIT_RAN = True
```
destijl: looks like this is unused?

tools/config_updater.py
```diff
@@ -593,7 +689,12 @@ def main(unused_argv):
   token = GetToken()
   config_lib.CONFIG.AddContext("Commandline Context")
   config_lib.CONFIG.AddContext("ConfigUpdater Context")
-  startup.Init()
+
+  if flags.FLAGS.subparser_name == "initialize":
+    startup.ConfigInit()
+    Initialize(config_lib.CONFIG, token=token)
```
destijl: I think this should return here? The rest of this code shouldn't be run if we're initializing.

lib/startup.py
pidydx: Good catch, forgot the finish removing that.

tools/config_updater.py
pidydx: I don't think it matters since everything else is conditional, but it shouldn't hurt to add.  I will put that in and make sure it doesn't cause anything weird and push it up

pidydx:
Ok, so I wound up doing some extra cleanup to make configuration easier and look better.  I removed the mandatory public hostname setup and left documentation in explaining that it works best when public.  I think it is better to provide better defaults for an internal only configuration since many GRR installs will be testing and a public IP is not necessarily a good indicator of whether or not the hostname is publicly resolvable.

I also removed some configuration from grr-server.yaml since the defaults were providing the same values and being in grr-server.yaml caused it to always appear as existing configuration rather than forcing it to prompt on initial setup.

tools/config_updater.py
destijl: It's more maintainable and signals a clear intent that the code path should end. At the moment it's not obvious if there is further processing that needs to happen, you have to look at all the if's and think about what how they will be evaluated.

destijl:
LGTM after you fix config/client.py

Issue #178, opened 2015-05-15
Frontend Performance


pidydx:
It appears that the frontends are constantly updating metadata:last on aff:/.  Is this necessary?  Having so many threads trying to hammer the same subject/attribute is causing delays.  I haven't been able to find the code to see exactly why it is doing that.

destijl:
Talked with @scudette about this.  This code is intended to maintain indexes:
https://github.com/google/grr/blob/master/lib/aff4.py#L196

It *shouldn't* be updating the root very often because the root should be really hot and basically always in the cache.  You can see this in the code above.  It's possible that the cache is too small, or not functioning correctly for some reason.  This is likely more of a problem during interrogate than other hunts because interrogate creates lots more aff4 path components (due to all the flows) than most hunts.

You could try increasing the size, perhaps it is too small?
https://github.com/google/grr/blob/master/config/aff4.py#L19

We should probably also add some monitoring about cache hits and misses so we can see whether this is an issue for us too.

pidydx:
I bumped the cache size to 20k from 10k and the intermediate to 10k from 2k.  It seemed like that might have helped, but when I tried to relax the restriction on the SQL connection pool it quickly went back to hammering the database on subject aff4:/ attribute metadata:last.

Originally we had this problem because the code was telling it to delete aff4:/ metadata:last and then insert a new aff4:/ metadata:last.  This was why I had to modify the MySQLAdvanced connector to try to do UPDATE instead of DELETE/INSERT wherever possible.

pidydx:
I bumped again from 20k to 200k and 10k to 100k and expanded the connection pool for the frontends and this behavior continued.  I don't think the cache is the problem.  This issue only impacts the frontends, the workers don't cause this to happen.

scudette:
I think I found the problem. The offending code is:

```
        try:
          self.intermediate_cache.Get(urn)
          return
        except KeyError:
          data_store.DB.MultiSet(dirname, {
              AFF4Object.SchemaCls.LAST: [
                  rdfvalue.RDFDatetime().Now().SerializeToDataStore()],

              # This updates the directory index.
              "index:dir/%s" % utils.SmartStr(basename): [EMPTY_DATA],
          },
                                 token=token, replace=True, sync=False)

          self.intermediate_cache.Put(urn, 1)

          urn = dirname
```

Should be:

```
        try:
          self.intermediate_cache.Get(dirname)
          return
        except KeyError:
          data_store.DB.MultiSet(dirname, {
[...]

PR #186, opened 2015-05-28
AFF4 cache patch

pidydx:
Patch provided by @scudette for https://github.com/google/grr/issues/178.

I am still monitoring the behavior of this fix, but early indications are very good. Sending this PR just in case it is easier to merge in this way after some time observing.  If this is not easier feel free to close.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/186)


destijl:
@scudette can you review?

PR #202, opened 2015-06-19
Server 0.3 6

455thatuuy:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/202)


destijl:
Hey, so the gh-pages branch is only there for a website that isn't currently active (we have a redirect on the main page to the github site). I wonder if you got linked there from somewhere and decided to update it?  Do you remember where the link came from? If we can change it to point to a currently maintained location we should.

destijl:
I'm going to close this, since we're not using the branch, see comment above.

PR #214, opened 2015-06-29
Extended flow durations to 1d

pidydx:
These hunts can take longer than 6hrs to successfully run if there is a larger amount of data to get through.  I have had instances of both complete the deletes, but still report timeout errors.  1d seems like a reasonable limit for a weekly cron.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/214)


destijl:
LGTM

PR #215, opened 2015-06-29
Cleaned up Upstart scripts and added systemd scripts

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/215)


config/centos/default/grr-http-server
```diff
@@ -1,5 +0,0 @@
-# Configuration file for grr-http-server start-up script
```
destijl: So what's happening with these centos files? I think you removed them to use the ones from the os-agnostic upstart directory. But I don't see any corresponding build code change for rpms (only debs). How does the rpmbuilder know where to find them now?

pidydx: I actually put those files there in the first place back in Oct just as storage so people could find server startup scripts that would work with Centos/RHEL if they ran the servers on non-Ubuntu Linux.  The rpmbuilder is only dealing with the client stuff so it only cares about  grr-client.initd.in

config/upstart/grr-dataserver-master.conf
```diff
@@ -10,11 +10,15 @@ kill timeout 300
 
 pre-start script
 	mkdir -p /var/log/grr
+	mkdir -p /var/run/grr/tmp/dataserver
```
destijl: Lets add -master and -slave to all of these paths. It's possible (but not super useful) to run them on the same machine.

config/centos/default/grr-http-server
destijl: Ah OK thanks.

destijl:
LGTM

config/upstart/grr-dataserver-master.conf
pidydx: Done!

PR #217, opened 2015-06-29
Removed SingleServer operations

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/217)


destijl:
This LGTM, thanks, just waiting on tests to make sure they pass.

scripts/install_script_ubuntu.sh
```diff
@@ -211,17 +211,22 @@ fi
...
-if [ $IS_RUNNING = 0 ]; then
-  run_cmd_confirm service grr-single-server stop
-fi
-run_cmd_confirm service grr-single-server start
+header "Enable grr services to start automatically on boot"
+
+for SERVER in grr-http-server grr-worker grr-ui
+do
+  SERVER_DEFAULT=/etc/default/${SERVER}
+  run_cmd_confirm sed -i 's/START=\"no\"/START=\"yes\"/' ${SERVER_DEFAULT};
```
destijl: AFAICT there's no point having these set to "no" anymore, they should always be yes. I'll merge this as is, and fix it when I bring it in internally.

destijl: oh I see, will need new server release before we can do that. I'll add a comment.

pidydx: It may be better to leave them as No just for a sane default when installing additional GRR servers.  We install them a bit differently so this wouldn't directly impact how we do it, but just a thought since this script is for installing everything on a single host where GRR at scale is distributed.

destijl: ACK

PR #219, opened 2015-07-01
Datastore cleanup

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/219)


destijl:
@mbushkov this one is yours

lib/flows/cron/data_retention_test.py
```diff
@@ -183,6 +184,149 @@ def testDeletesFlowsOlderThanGivenAge(self):
...
+      flow.GRRFlow.StartFlow(
+          flow_name=data_retention.CleanTemp.__name__,
+          sync=True, token=self.token)
+
+    tmp_urns = list(aff4.FACTORY.Open("aff4:/tmp",
+                                        token=self.token).ListChildren())
+    self.assertEqual(len(tmp_urns), 10)
+
+  def testDeletesTempWithAgeOlderThanGivenAge(self):
+    config_lib.CONFIG.Set("DataRetention.tmp_ttl",
```
grrrrrrrrr: can you use a configoverrider context here? if you just set this, you will leak state. So if, for example, you run this test and afterwards the testDoesNothingIfAgeLimitNotSetInConfig above, the tmp_ttl will actually be set and the other test will fail. If you use a overrider context, it will restore the old settings after the test.

grrrrrrrrr:
This looks pretty good already, maybe you can fix that one small thing in the test so we can merge.

pidydx:
I moved the others over to ConfigOverriders as well

pidydx:
I altered this to use the client index to find client urns to delete instead of ListChildren, but I noticed taht the client index never gets cleaned up so it is another infinite growth point unless a way to remove a client from the index is added or the index is periodically rebuilt.

lib/flows/cron/data_retention.py
```diff
@@ -66,3 +65,65 @@ def Start(self):
...
+        "DataRetention.inactive_client_ttl_exception_label"]
+
+    index = aff4.FACTORY.Open(client_index.MAIN_INDEX, mode="r",
+                               token=self.token)
+    client_urns = index.LookupClients(["."])
+
+    deadline = rdfvalue.RDFDatetime().Now() - inactive_client_ttl
+
+    for client_urn in client_urns:
+      client = aff4.FACTORY.Open(client_urn, mode="r", token=self.token)
```
grrrrrrrrr: Opening clients one by one will not scale, please group them in batches and open using MultiOpen.
Example code that does this is here:
https://github.com/google/grr/blob/master/lib/hunts/implementation.py#L982

You can use a much bigger chunk size than we use there, maybe 10000 would be a good number.

Once you have all clients open, please collect the ones to delete and use aff4.FACTORY.MultiDelete. Should also be much faster.

lib/test_lib.py
```diff
@@ -571,7 +571,7 @@ def SetupClients(self, nr_clients):
         fd.Flush()
 
         index.AddClient(fd)
-
+    index.Close()
```
grrrrrrrrr: Hm do you find this necessary? I don't think this should do anything.

pidydx: I found this because I switched over to using the client index to find clients.  self.SetupClients(x) creates and adds the clients to the index, but it doesn't actually get written back until index.Close() since this isn't using "with" to create the index.  From what I saw other unit tests were either building a separate client index or otherwise updating the client index separately.

The problem was that in the unit test I would use SetupClients and then run the flow, but the index would be empty so it couldn't find any clients to delete and the test would fail.  After adding index.Close() the index would get updated and the flow would successfully find the clients and delete them.

grrrrrrrrr: Ah I see the issue, you are opening the index instead of creating so you rely on the object actually being there. If you change your code to use Create instead of Open this should work and be even a little bit faster. If you prefer, you can also just write the index object here but then I'd prefer a with statement.

pidydx: Wouldn't the index still be missing all the clients using Open instead of Create?

pidydx: I went ahead and switched SetupClients to use a with statement.  This seems like the more correct behavior for SetupClients either way.

grrrrrrrrr: No, the ClientIndex aff4 object writes all data directly to the DB unless you explicitly give it sync=False. The problem you encountered was not that the data wasn't there but that when you called Open on the MAIN_INDEX urn, there was no aff4 type there so you got back an AFF4Volume instead of an index and that would not know about all the functions you called afterwards. Just give it a try and see if it works?

lib/flows/cron/data_retention.py
pidydx: Done! There might be problems if it finds too many clients to delete in a chunk and takes too long in MultiDelete, but it seems like an unlikely problem and should only impact environments with large numbers of inactive clients on first run.

destijl:
@grrrrrrrrr can you finish this review?

grrrrrrrrr:
Review status: 0 of 5 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

<sup>**[lib/flows/cron/data_retention.py, line 93 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBC4OTh1tpQ-dJ9Kk4)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention.py#L93)):</sup>
It's more efficient to use MultiOpen here.

---

<sup>**[lib/flows/cron/data_retention.py, line 137 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBCTkGLJti453QxT_W)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention.py#L137)):</sup>
This seems wrong, it deletes after each client. Heartbeat here and MultiDelete after each batch?

---

<sup>**[lib/flows/cron/data_retention_test.py, line 294 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBCw4xHH5DyKEKnat6)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention_test.py#L294)):</sup>
nit: long line

---

<sup>**[lib/flows/cron/data_retention_test.py, line 322 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBD3tNfn42z6tEfBuw)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention_test.py#L322)):</sup>
same here

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/219)

pidydx:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[lib/flows/cron/data_retention.py, line 93 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBC4OTh1tpQ-dJ9Kk4)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention.py#L93)):</sup>
I switched out the code to use the grouping like the clients and then MultiOpen and MultiDelete

---

<sup>**[lib/flows/cron/data_retention.py, line 137 \[r6\]](https://reviewable.io:443/reviews/google/grr/219#-JvBCTkGLJti453QxT_W)** ([raw file](https://github.com/google/grr/blob/d34d6d8b3c979c19fad2fa6a8e575cdd23f8dc7c/lib/flows/cron/data_retention.py#L137)):</sup>
Oops, the MultiDelete and heartbeat are indented one too many levels.  It works, just not exactly efficient.  I will fix this.

---

<sup>**[lib/test_lib.py, line 573 \[r3\]](https://reviewable.io:443/reviews/google/grr/219#-JtGc0G8jp-4CtEUEF-h-r3-574)** ([raw file](https://github.com/google/grr/blob/69a0178a7cd2ecbe8705d16f2ccb432620aa2d2e/lib/test_lib.py#L573)):</sup>
I think this has already been resolved by using Create instead of Open, but it was showing as a new comment that hadn't been acknowledged.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/219)

PR #220, opened 2015-07-03
Interrogate fix

pidydx:
I left the Creates in the UI alone since they shouldn't ever be encountered at a rate that causes problems and it should ensure that the client index does get added to the root index so that nothing is fundamentally altered in how everything is getting stored.

I will be testing this soon against MySQLAdvanced with a much larger connection pool to ensure this solves the interrogate hunt performance issues.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/220)


grrrrrrrrr:
Looks good, thanks!

PR #227, opened 2015-07-22
Added debugging to cache usage

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/227)


destijl:
This looks like it might create a lot of logs....Fine to add something like this for debugging yourself but not sure we want it in the codebase?

pidydx:
Yeah, this doesn't need to be merged in at this time.  I set this up mostly for visibility into tracking the potential cache issues in the event it might evolve into a more robust cache hit/miss tracking.

PR #228, opened 2015-07-23
Add docker server image and modify install scripts.

destijl:
- Add a docker image. For now this runs all components in the same image.
  I'll break them out in the future, but there's still quite a bit of work
  involved.
- Add entrypoint script to handle running initialize and setting essential
  config.
- Handle missing /dev/log in docker containers gracefully.
- Use gift PPA to fulfill all non-pip-installable dependencies for travis.
- Modify install script to use the gift PPA.
- Add a noprompt option to config_updater to allow docker image initialization
  to work without prompts.
- Remove essentially all of the debian control file dependencies. It's weird
  having a partial list here that doesn't actually get you a working grr server.
  And having hard dependencies has burned us in the past (mongo and protobuf).
  If/when gift PPA has all the dependencies we need we can add them back in and
  have everything installed by apt.
- Make Sqlite the default database.
- Move travis scripts into their own directory and add a script to allow for
  easy testing with vagrant.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/228)


destijl:
@grrrrrrrrr Wanted to get this out for review but there are a few more things to do before I submit.

- merge sean's change
- build new package
- upload new package to drive
- modify Dockerfile to download it from there.

grrrrrrrrr:
Review status: 0 of 18 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[tools/config_updater.py, line 116 \[r2\]](https://reviewable.io:443/reviews/google/grr/228#-JvFPdfOxWsjeaG2QEqt)** ([raw file](https://github.com/google/grr/blob/ce105462fcfcbd2d6cdc2275b8e8ec237507bd89/tools/config_updater.py#L116)):</sup>
fix comment.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/228)

destijl:
Review status: 0 of 19 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[tools/config_updater.py, line 116 \[r2\]](https://reviewable.io:443/reviews/google/grr/228#-JvFPdfOxWsjeaG2QEqt)** ([raw file](https://github.com/google/grr/blob/ce105462fcfcbd2d6cdc2275b8e8ec237507bd89/tools/config_updater.py#L116)):</sup>
Thanks, done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/228)

PR #229, opened 2015-07-26
Increased intermediate cache size and optimized for heavily accessed …

pidydx:
…urns

[Review on Reviewable](https://reviewable.io/reviews/google/grr/229)

### DIFF
--- a/config/client.py
+++ b/config/client.py
@@ -48,7 +48,7 @@ config_lib.DEFINE_string(
     help="Where GRR stores cached Rekall profiles needed for memory analysis")
 
 config_lib.DEFINE_list("Client.control_urls",
-                       ["http://www.example.com/control"],
+                       ["http://localhost:8080/control"],
                        "List of URLs of the controlling server.")
 
 config_lib.DEFINE_string("Client.plist_path",
--- a/config/grr-server.yaml
+++ b/config/grr-server.yaml
@@ -6,18 +6,11 @@ AdminUI.bind: "::"
 
 AdminUI.django_debug: True
 
-AdminUI.port: 8000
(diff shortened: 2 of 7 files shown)


pidydx:
@grrrrrrrrr Can you sanity check this change for the expected behavior?  The overflow of the cache seems to be FIFO and Put says it removes the old object and puts the new one so the expectation is that it would be moving heavily accessed objects up the list while the less accessed objects are the ones that get pushed out during overflow.  I also expanded the intermediate cache by a large chunk so that there is more room for a worker processing a lot of new aff4 objects to keep moving heavily accessed ones to the front before they get pushed out.

lib/aff4.py
```diff
@@ -392,6 +392,10 @@ def _UpdateChildIndex(self, urn, token):
 
         try:
           self.intermediate_cache.Get(urn)
+          # When an object is retrieved from the cache bump it back to the
+          # front to ensure heavily accessed items remain in the cache for their
+          # full age.
+          self.intermediate_cache.Put(urn, 1)
```
grrrrrrrrr: This should not be necessary. The cache always moves all accessed items to the front of the list on access. Are you worried about the time expiring? That causes one write every 10 minutes in the current configuration, which should be ok, don't you think?

pidydx: Yeah, I just reread through the Gets and saw how it is already moving it to the front of the list in FastStore.  Once every 10 minutes should be fine, but when we are getting the collisions we are seeing it happen way more often.  I figured it was just getting pushed out due to lots of new aff4 objects being created when processing bulk collection file/registry results.

PR #230, opened 2015-07-26
Updated for syncronized Flush

pidydx:
[Review on Reviewable](https://reviewable.io/reviews/google/grr/230)

### DIFF
--- a/config/aff4.py
+++ b/config/aff4.py
@@ -16,7 +16,7 @@ config_lib.DEFINE_integer(
     "The number of seconds AFF4 urns live in index cache.")
 
 config_lib.DEFINE_integer(
-    "AFF4.intermediate_cache_max_size", 2000,
+    "AFF4.intermediate_cache_max_size", 20000,
     "Maximum size of the AFF4 index cache.")
 
 config_lib.DEFINE_integer(
--- a/lib/data_stores/mysql_advanced_data_store.py
+++ b/lib/data_stores/mysql_advanced_data_store.py
@@ -442,7 +442,7 @@ class MySQLAdvancedDataStore(data_store.DataStore):
           # This should indicate missing tables and raise immediately
           raise e
         else:
-          logging.warn("Datastore query failed with %s:", str(e))
+          logging.warn("Datastore query retrying after failed with %s.", str(e))
           # Most errors encountered here need a reasonable backoff time to
           # resolve.
           time.sleep(1)
@@ -480,7 +480,7 @@ class MySQLAdvancedDataStore(data_store.DataStore):
           # This should indicate missing tables and raise immediately
           raise e
         else:
-          logging.warn("Datastore query failed with %s:", str(e))
+          logging.warn("Datastore query retrying after failed with %s.", str(e))
           # Most errors encountered here need a reasonable backoff time to
           # resolve.
           time.sleep(1)
--- a/lib/data_stores/mysql_advanced_data_store.py
+++ b/lib/data_stores/mysql_advanced_data_store.py
@@ -152,6 +152,7 @@ class MySQLAdvancedDataStore(data_store.DataStore):
     try:
       self.ExecuteQuery("desc `aff4`")
     except MySQLdb.Error:
+      logging.debug("Recreating Tables")
       self.RecreateTables()
 
   def DropTables(self):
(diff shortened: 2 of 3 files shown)


pidydx:
@grrrrrrrrr Live testing of each of these changes will begin this week, but a sanity check on the changes would be appreciated, particularly for the reduction and delay of non-sync updates.

grrrrrrrrr:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/data_stores/mysql_advanced_data_store.py, line 376 \[r1\]](https://reviewable.io:443/reviews/google/grr/230#-JvBXPipZVWyl9G62XWm)** ([raw file](https://github.com/google/grr/blob/6abd7052016b82c5162c51140f8b5a786504219a/lib/data_stores/mysql_advanced_data_store.py#L376)):</sup>
You need to update updates[subject], not overwrite - it might already have some other attribute in the dict.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

grrrrrrrrr:
Looks really good except for that one small bug. Looking forward to seeing the results...

pidydx:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/data_stores/mysql_advanced_data_store.py, line 376 \[r1\]](https://reviewable.io:443/reviews/google/grr/230#-JvBXPipZVWyl9G62XWm)** ([raw file](https://github.com/google/grr/blob/6abd7052016b82c5162c51140f8b5a786504219a/lib/data_stores/mysql_advanced_data_store.py#L376)):</sup>
So I tested this a bit last night, but I was also drinking soooo...

What should happen here is if the updates dict already has a key of subject, then updates[subject][attribute] = X should either create the key for attribute and assign it, or overwrite an existing key's value.  It should only throw a KeyError exception if the updates dict doesn't have a key for subject so it can safely create it as a dict.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

grrrrrrrrr:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/data_stores/mysql_advanced_data_store.py, line 376 \[r1\]](https://reviewable.io:443/reviews/google/grr/230#-JvBXPipZVWyl9G62XWm)** ([raw file](https://github.com/google/grr/blob/6abd7052016b82c5162c51140f8b5a786504219a/lib/data_stores/mysql_advanced_data_store.py#L376)):</sup>
Here is the bug:

I call replace on subject, attribute1 and subject, attribute2.

in the loop you first get the key error for subject and do

updates[subject] = {attribute1: [...]}

but now you get subject, attribute2 and try to access

updates[subject][attribute2]

the first part here works but attribute2 is not in the second dict so you get a KeyError and go

updates[subject] = {attribute2: [...]}

and that overwrites the first update already there.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

pidydx:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/data_stores/mysql_advanced_data_store.py, line 376 \[r1\]](https://reviewable.io:443/reviews/google/grr/230#-JvBXPipZVWyl9G62XWm)** ([raw file](https://github.com/google/grr/blob/6abd7052016b82c5162c51140f8b5a786504219a/lib/data_stores/mysql_advanced_data_store.py#L376)):</sup>
updates[subject][attribute2] doesn't cause a KeyError, it just creates the missing key and assigns it.  Python can only do one level of create and assign like this.  This is actually one of my only real pet peeves with Python since Perl will automatically create all the keys necessary to do the assignment instead of just the last one. 

>>> values = [[1,1,1,1],[1,1,2,2],[1,2,4,4],[2,2,2,2],[2,3,3,3]]
>>> updates = {}
>>> for (subject, attribute, value, timestamp) in values:
...     try:
...         updates[subject][attribute] = [value, timestamp]
...     except KeyError:
...         updates[subject] = {attribute: [value, timestamp]}
... 
>>> for subject in updates:
...     for attribute in updates[subject]:
...         print str(subject)+str(attribute)+str(updates[subject][attribute][0])+str(updates[subject][attribute][1])
... 
1122
1244
2222
2333

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

scudette:
You can just use the setdefault() trick.

X.setdefault(x, {})[y] = z
Am 27.07.2015 01:24 schrieb "Sean Gillespie" <notifications@github.com>:

grrrrrrrrr:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[lib/data_stores/mysql_advanced_data_store.py, line 376 \[r1\]](https://reviewable.io:443/reviews/google/grr/230#-JvBXPipZVWyl9G62XWm)** ([raw file](https://github.com/google/grr/blob/6abd7052016b82c5162c51140f8b5a786504219a/lib/data_stores/mysql_advanced_data_store.py#L376)):</sup>
Oh my, I'm sorry. Completely read the code wrong, this is obviously fine. I'd second Michaels suggestion of using setdefault though.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

pidydx:
I was not aware of the setdefault() trick.  Thanks!  

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/google/grr/230)

PR #240, opened 2015-08-04
Converted MultiSet and Flush to use a single transaction to prevent p…

pidydx:
…otential race conditions and added locking to Flush.  Also made error handling for deadlocks, lost connections, and other errors more resilient. Fixed a possible connection pool leak.

[Review on Reviewable](https://reviewable.io/reviews/google/grr/240)


grrrrrrrrr:
LGTM

PR #315, opened 2015-12-11
Bumped to OpenSSL 1.0.2e

brian-olson:
This has been tested on Cent & Ubuntu. Windows is untested.


destijl:
Thanks for the PR. I already made the same change internally though so I'm going to save myself some time and just export mine. Appreciate the effort though.

PR #322, opened 2016-01-18
Add case to GetDistro() for CentOS linux.

andrewkrug:
Operating system detection for setting up server.yml non functional on CentOS to spite binary compatibility.  Added a case for this to GetDistro().

### DIFF
--- a/lib/distro_entry.py
+++ b/lib/distro_entry.py
@@ -27,9 +27,7 @@ def GetDistro():
     distribution = platform.linux_distribution()[0].lower()
     if distribution in ["ubuntu", "debian"]:
       return "debian"
-    if distribution in ["red hat enterprise linux server"]:
-      return "redhat"
-    if distribution in ["centos linux"]:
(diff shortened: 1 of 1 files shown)


lib/distro_entry.py
```diff
@@ -29,6 +29,8 @@ def GetDistro():
       return "debian"
     if distribution in ["red hat enterprise linux server"]:
       return "redhat"
+    if distribution in ["centos linux"]:
```
grrrrrrrrr: cool thanks! Do you mind putting the centos string in the list right above since we have it already?

andrewkrug: Not a problem.  Just made that change.

andrewkrug:
Looks like the travis build failed only due to some rekall tests that aren't currently working.  Let me know if that is indeed the case.

grrrrrrrrr:
Yes, the test failures are not related, they are caused by the outdated Rekall version. Thanks!

PR #325, opened 2016-01-22
Added dependencies for MySQL client connections and updated config_up…

pidydx:
…dater warning

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -42,3 +42,4 @@ wsgiref==0.1.2
 Werkzeug==0.11.3
 GRR-M2Crypto==0.22.4
 pytsk3==20160207.post1
+MySQL-python==1.2.5
--- a/scripts/install_script_ubuntu.sh
+++ b/scripts/install_script_ubuntu.sh
@@ -149,6 +149,7 @@ sudo apt-get install -y \
   libfreetype6-dev \
   libpng-dev \
   libprotobuf-dev \
+  libmysqlclient-dev \
   ncurses-dev \
   pkg-config \
   prelink \
--- a/tools/config_updater.py
+++ b/tools/config_updater.py
@@ -509,6 +509,18 @@ well.
     config.Set("Datastore.location", datastore_location)
 
   if datastore == "2":
+    print """\n\n***WARNING***
+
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -40,3 +40,4 @@ selenium==2.44.0
 urllib3==1.10
 wsgiref==0.1.2
 Werkzeug==0.11.3
+MySQL-python==1.2.5
```
destijl: This library is small, 110k. It needs the mysql libs to install, so it won't work on os x or windows, but we have separate client building requirements.txt for those. So I think this is OK to go in.

scripts/install_script_ubuntu.sh
```diff
@@ -149,6 +149,7 @@ sudo apt-get install -y \
   libfreetype6-dev \
   libpng-dev \
   libprotobuf-dev \
+  libmysqlclient-dev \
```
destijl: This also needs to go into https://github.com/google/grr/blob/master/travis/before_install.sh

tools/config_updater.py
```diff
@@ -501,6 +501,18 @@ def ConfigureDatastore(config):
...
+    print """\n\n***WARNING***
+
+Do not continue until a MySQL server, version 5.6 or greater, is running and a user
+with the ability to create the GRR database and tables has been created. You will need
+the server, username, password, and database name (if already created) to continue. If
+no database has been created this script will attempt to create the necessary database 
+and tables using the credentials provided.
+
+***WARNING***
+"""
```
destijl: Fix line breaks here.

pidydx: You mean use \n instead of literals?

destijl: Looks like it isn't wrapped at 80 correctly, there's breaks in the middle of sentences

pidydx: Does this look better?

scripts/install_script_ubuntu.sh
pidydx: Fixed

destijl:
LGTM this can go in on next sync, assuming travis builds OK.

PR #326, opened 2016-01-22
Added Server.username prompt for running as non-root and added output…

pidydx:
… of significant configuration options that do not get configured by default

### DIFF
--- a/config/server.py
+++ b/config/server.py
@@ -244,9 +244,8 @@ config_lib.DEFINE_string(
     " is only scheduled after results are received in the previous poll.")
 
 config_lib.DEFINE_string(
-    "Server.username", None,
-    "System account to run as after initialization for running the server as "
-    "non-root.")
+    "Server.username", 'root',
+    "System account for services to run as after initialization.")
 
 # Email Template Values
 config_lib.DEFINE_string(
--- a/tools/config_updater.py
+++ b/tools/config_updater.py
@@ -13,7 +13,7 @@ import readline  # pylint: disable=unused-import
 import socket
 import sys
 import urlparse
-
+import pwd
 
 # pylint: disable=unused-import,g-bad-import-order
 from grr.lib import server_plugins
@@ -134,7 +134,7 @@ def AddUser(username, password=None, labels=None, token=None):
 
   fd = aff4.FACTORY.Create("aff4:/users/%s" % username,
                            "GRRUser", mode="rw", token=token)
-    # Note this accepts blank passwords as valid.
+  # Note this accepts blank passwords as valid.
   if password is None:
     password = getpass.getpass(
         prompt="Please enter password for user '%s': " % username)
@@ -394,7 +394,7 @@ def GenerateKeys(config):
   if not hasattr(key_utils, "MakeCACert"):
     parser.error("Generate keys can only run with open source key_utils.")
   if (config.Get("PrivateKeys.server_key", default=None) and
-      not flags.FLAGS.overwrite):
+          not flags.FLAGS.overwrite):
     raise RuntimeError("Config %s already has keys, use --overwrite to "
                        "override." % config.parser)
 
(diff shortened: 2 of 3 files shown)


tools/config_updater.py
```diff
@@ -560,11 +560,55 @@ def ConfigureEmails(config):
...
+  username = RetryQuestion("GRR Server Username",
+                           "^[a-z0-9]+$",
+                           config_lib.CONFIG.Get("Server.username"))
+
+  try:
+    pwd.getpwnam(username)
+    print "User %s exists. Continuing..." % username
+    success = True
+  except KeyError:
+    user_create_error = os.system('useradd -r %s -M -d /nonexistant -s '
```
destijl: default is no homedir, which I think is less weird than setting a bad one.

tools/config_updater.py
```diff
@@ -560,11 +560,55 @@ def ConfigureEmails(config):
   config.Set("Monitoring.emergency_access_email", emergency_email)
 
 
+def ConfigureUser(config):
+  success = False
+  print
```
destijl: remove

tools/config_updater.py
```diff
@@ -696,6 +740,91 @@ def ManageBinaries(config=None, token=None):
   print "Please restart the service for it to take effect.\n\n"
 
 
+def SuggestedConfiguration():
```
destijl: This is good information, but I don't think putting it in code is the right thing to do. Lets add a documentation page and reference it in the "initialization complete message"

tools/config_updater.py
pidydx: /nonexistant is a Debian/Ubuntu convention for users without homedirs.  I'm not terribly opposed to changing it, but I figured I would go with what would look "normal" for the typical system.

tools/config_updater.py
pidydx: I was a little torn on that myself.  I went with code for a few reasons and I tried to limit this to the most significant.  I figured the documentation could be generated off the help messages to provide a comprehensive reference rather than just "Please be aware of these specific things before you hit the go button".  

1.  In the code it can display the current values.

2.  This ensures people actually at least see what their system is going to be doing before they start the system if they skip reading the docs.

3. Some of these configurations should/could become configuration prompts rather than just displays.

tools/config_updater.py
pidydx: This was just there to add a newline for visual niceness during config.  I'm not a fan of doing it this way, but wasn't sure where else to put it.  Should I just add a leading \n to the RetryQuestion() call?

tools/config_updater.py
destijl: ack, OK.

tools/config_updater.py
destijl: Yeah, but all of this is pre-production deployment config and can wait until after you've got it up and running and tried things out. The install process is long enough and full of tons of prompts. I think we should beg off anything that isn't essential to getting a working install until later. These are the sorts of config changes you should make once you have a config under version control and are working on getting it into production shape.

tools/config_updater.py
destijl: yeah, do that.

tools/config_updater.py
pidydx: Makes sense.  I will pull it out.

tools/config_updater.py
pidydx: Fixed

config/server.py
```diff
@@ -244,7 +244,7 @@
     " is only scheduled after results are received in the previous poll.")
 
 config_lib.DEFINE_string(
-    "Server.username", None,
+    "Server.username", 'root',
     "System account to run as after initialization for running the server as "
```
destijl: "System account to run as after initialization" since the rest doesn't really make sense now.

destijl:
LGTM, this can go in after the doco change and next sync.

config/server.py
pidydx: Fixed.

PR #329, opened 2016-01-30
MySQL cleanup

pidydx:


destijl:
@bgalehouse it would be good to get this in before we release the server. Can you take a look?

destijl:
Ping @bgalehouse

PR #376, opened 2016-05-18
Updated Steps in config_updater.py

jessicawilsonsc:
Noticed that it jumped from "Step 3" to "Step 5", so I fixed it.


jessicawilsonsc:
I signed it!

destijl:
Heh, it will still be wrong after this, update number 6 too :)

jessicawilsonsc:
Done!

Issue #377, opened 2016-05-19
Add data_server entry point


destijl:
from @scudette 

I think we forgot to add an entry point for data_server.py which is a
bug, thanks for reporting. Basically we just need to add another one
like:
https://github.com/google/grr/blob/master/grr/lib/distro_entry.py#L116

pointing at the tool
https://github.com/google/grr/blob/master/grr/server/data_server/data_server.py#L843

and specify it in the setup.py:
https://github.com/google/grr/blob/master/grr/config/grr-response-server/setup.py#L18

PR #378, opened 2016-05-19
Create Dataserver Entry Point

jessicawilsonsc:
Added in an entry point as per conversation in [GRR-Users group- Separation of Services Question]( https://groups.google.com/forum/#!topic/grr-users/Be6f0aig9iE ).


destijl:
Awesome! Although I think @scudette started on this and found there was a bit more to it...

https://github.com/google/grr/issues/377

destijl:
We'll put this in and add a new context (which is the missing piece) separately, thanks!

PR #380, opened 2016-05-25
Added option to define dmg package format

jessicawilsonsc:
Needed to add a package format option to build a custom darwin client. 

Per discussion https://groups.google.com/forum/#!topic/grr-users/scEpaarIiOI


destijl:
There's some conflicts that need to be resolved here but this change can go in. Since it's so small I'm inclined to just make the change internally and include it in the next export. Thanks for adding though.

Issue #279, opened 2015-10-16
Database disk image is malformed


cjdupreez:
I've gotten GRR installed and can access it via the web browser, but when I click on the "Cron Job Viewer" nothing is displayed and in an error message I'm presented with the following traceback:

Error: Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/renderers.py", line 318, in Decorated
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/views.py", line 156, in RenderGenericRenderer
    result = method(request, result) or result
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/plugins/cron_view.py", line 33, in Layout
    response = super(ManageCron, self).Layout(request, response)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/renderers.py", line 988, in Layout
    state=self.state.copy()).RawHTML(request)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/renderers.py", line 373, in RawHTML
    method(request, result, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/plugins/cron_view.py", line 161, in Layout
    response = super(CronTable, self).Layout(request, response)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/renderers.py", line 656, in Layout
    self.table_contents = delegate_renderer.RenderAjax(request, tmp).content
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/renderers.py", line 691, in RenderAjax
    limit_row + start_row, request)
  File "/usr/local/lib/python2.7/dist-packages/grr-0.3.0_7-py2.7-linux-x86_64.egg/grr/gui/plugins/cron_view.py", line 188, in BuildTable
    cron_jobs_urns = list(cronjobs.CRON_MANAGER.ListJobs(token=request.token))
[...]

PR #395, opened 2016-06-13
Prevent infinite loop when data store fails

chlunde:
If we get an empty or short response, the process will loop forever trying to read the rest of the data.  Prevent this by raising an exception if the response is not long enough.  I saw this issue on other nodes when a data server ran out of file descriptors.

An alternative would be to make sure _ReadReply returns None for a more graceful retry.  What do you think?

I guess this is not a common issue, because processes stuck at 100% CPU forever tends to stick out.

### DIFF
--- a/grr/lib/data_stores/http_data_store.py
+++ b/grr/lib/data_stores/http_data_store.py
@@ -83,21 +83,22 @@ class DataServerConnection(object):
     ret = ""
     left = n
     while left:
-      ret += self.sock.recv(left)
+      data = self.sock.recv(left)
+      if data == "":
+        raise IOError("Expected %d bytes, got EOF after %d" % (n, len(ret)))
+      ret += data
       left = n - len(ret)
     return ret
 
   def _ReadReply(self):
     try:
       replylen_str = self._ReadExactly(sutils.SIZE_PACKER.size)
-      if not replylen_str:
-        raise HTTPDataStoreError("Could not read reply from data server.")
       replylen = sutils.SIZE_PACKER.unpack(replylen_str)[0]
       reply = self._ReadExactly(replylen)
       response = rdf_data_store.DataStoreResponse(reply)
       CheckResponseStatus(response)
       return response
-    except (socket.error, socket.timeout) as e:
+    except (socket.error, socket.timeout, IOError) as e:
       logging.warning("Cannot read reply from server %s:%d : %s",
                       self.Address(), self.Port(), e)
       return None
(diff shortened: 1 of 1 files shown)


chlunde:
I believe there's also a similar potential problem two places in `grr/server/data_server/data_server.py`.

chlunde:
I signed it!

grr/lib/data_stores/http_data_store.py
```diff
@@ -83,7 +83,10 @@ def _ReadExactly(self, n):
     ret = ""
     left = n
     while left:
-      ret += self.sock.recv(left)
+      data = self.sock.recv(left)
+      if data == "":
+        raise IOError("Expected %d bytes, got EOF after %d", n, len(ret))
```
destijl: maybe it makes more sense to raise socket.error so it will return none in readreply. @grrrrrrrrr you know this code better than me, can you take a look?

chlunde: Note to self: I should also have used `% (n, len(ret))`.

grrrrrrrrr: I agree that catching the exception would be better but I also like IOError better than socket error - the socket probably works just fine. Can you maybe add IOError to the exceptions we catch in _Reconnect and _ReadReply?

destijl: And you're right about the format string, that needs fixing.

chlunde:
Thanks for the review! I have updated the PR. I have not tested it because of the missing clients/#394.  I also removed `if not replylen_str` because I don't think it is - or was - reachable.

destijl:
Thanks, can you give this a test now the clients are out? It's a fair bit of work for us to stand up a datastore just to test this.

chlunde:
Sorry for the delay, I've been busy with other tasks.

I believe I found the root cause: Either through out of memory or too many open files on the data nodes four C.nnnnn.sqlite files were malformed.  This made the data server return an empty response, which caused this infinite loop.

With this patch on 0.3.0.post7, the system behaved better but there were still many issues and restarts were required.  After I removed the four corrupt client databases everything seems much better.  I believe there might be a few issues still left if this everything should work "smoothly" with a corrupt file:

* In sqlite_data_store.py, Execute tries to log the file name for a malformed database (#279), but in my case, the error was actually thrown later, for example in fetchall(), so the file name was never logged.
* I've seen "bad filedescriptor" errors and spinning/repeating requests from the admin server

In conclusion, this patch makes the system behave better in error cases, but it does not solve the problem if a file is corrupt.  But at least you will not get workers or server spinning in an infinite loop if a database server is restarted, which was possible without this patch.

destijl:
LGTM. The downside of the sqlite datastore seems to be that it's fairly easy to corrupt one of the individual files. We should look at other ways we can make the problem more obvious, feel free to suggest any ideas you have.

destijl:
Thanks for your work. Let me know how you'd like to be listed in the ACK file:
https://github.com/google/grr/blob/master/ACKNOWLEDGEMENTS

PR #407, opened 2016-07-05
Create install_CentOS_virtualenv.sh

jolachea:
Created a script to install the server in CentOS 7 into a virtualenv.

### DIFF
--- a/scripts/install_CentOS7_virtualenv.sh
+++ b/scripts/install_CentOS7_virtualenv.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# Install dependencies.
+sudo yum -y update
+echo "Installing epel-release."
+sudo yum -y install epel-release
+sudo yum -y update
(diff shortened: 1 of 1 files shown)


jolachea:
Signed it.

destijl:
Thanks for sending this over. It's very similar to

https://github.com/google/grr/blob/master/vagrant/install_linux.sh

though and I don't think we really want to maintain this code in two places. More eyes (and use) on a single script means more chance it will actually work when people want to use it. I realise you're also automating the installfrompip instructions. We should perhaps make that a script too, then your centos install would be:

vagrant/install_linux.sh CentOS
script/install_from_pip.sh

jolachea:
Yes.  The CentOS 5.11 script breaks yum.  I will keep working on some things.  No need to do an install from pip script.  That's easy enough.

I am working on getting this to a prod state in CentOS 7.  Challenging but I wanted to get the virtualenv working first. Thanks.

Issue #429, opened 2016-09-06
 ArtifactDefinitionError: Artifact LinuxAtJobs missing from registry


giray:
Running grrdocker/grr:v3.1.0.2-latest, and I get the following error

`Validate raise ArtifactDefinitionError(e) ArtifactDefinitionError: Artifact LinuxAtJobs missing from registry. You may need to sync the artifact repo by running make in the artifact directory.`

detailed trace:
```
Traceback (most recent call last): File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow_runner.py", line 613, 
in RunStateMethod responses=responses) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow.py", line 353, 
in Decorated res = f(*args[:f.func_code.co_argcount]) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flows/general/discovery.py", line 153, 
in Platform next_state="ProcessKnowledgeBase") File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow.py", line 790, 
in CallFlow **kwargs) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow_runner.py", line 908, 
in CallFlow **kwargs) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow.py", line 933, 
in StartFlow flow_obj.Start() File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/flow.py", line 353, in Decorated res = f(*args[:f.func_code.co_argcount]) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/artifact.py", line 143, 
in Start first_flows = self.GetFirstFlowsForCollection() File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/artifact.py", line 407, 
in GetFirstFlowsForCollection artifact_registry.REGISTRY.GetArtifact(artifact_name) File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/artifact_registry.py", line 293, 
in GetArtifact self._CheckDirty() File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/artifact_registry.py", line 222, 
in _CheckDirty self._ReloadArtifacts() File "/usr/share/grr-server/local/lib/python2.7/site-packages/grr/lib/artifact_registry.py", line 201, 
[...]

PR #437, opened 2016-09-20
Set artifacts to 20160617 release for 3102 branch.

destijl:
- The 3102 server was released before a large artifact rename change so the
  server needs to use the older version.


destijl:
@scudette This fixes the tests on the release branch.

destijl:
https://github.com/google/grr/issues/429

PR #447, opened 2016-10-04
Add ListFlowResults to api_client FlowBase.

someguyiknow:


destijl:
This file has moved, but looks like ListFlows might still be necessary:
https://github.com/google/grr/blob/master/api_client/python/grr_api_client/flow.py

@someguyiknow do you still want to get this done?

mbushkov:
Greg, I think it's obsolete: we have ListFlows here: https://github.com/google/grr/blob/master/api_client/python/grr_api_client/client.py#L123
and ListFlowResults is called "ListResults" and is here:
https://github.com/google/grr/blob/master/api_client/python/grr_api_client/flow.py#L46

mbushkov:
Closing as the features are already implemented in the library in a slightly different way.

PR #457, opened 2016-11-21
Rename startup.py

jawilson0502:
Found a bug in `grr/gui/admin_ui.py` . It had `from grr.lib import server_startup`, however the corresponding file in `grr/lib` was named `startup.py`.  This caused a bug that did not allow `grr_server --component` to run.


jawilson0502:
I signed it

zbuc:
Can confirm, this change seems necessary to run GRR from the `master` branch right now.

jawilson0502:
Actually, this PR may not be perfect. `grr/server/data_server/data_server.py` still imports `startup.py`. So renaming this isn't the fix... I can change the code in `grr/gui/admin.py` to use only `startup.py` if that is how it should be done to save the devs some time :smile:

grrrrrrrrr:
Hey,

Thanks for this PR! This change actually went in too early - we are going to rename those files in a future commit but somehow this import change was prematurely committed. The correct fix is to change the import back (for now, until the big rename happens). I'll send a commit that does that.

Cheers,
-Andy

grrrrrrrrr:
Should be fixed in https://github.com/google/grr/commit/9edca58aa45b195a6648eed42493219d7421769c

PR #476, opened 2017-02-25
linux client interrogate fixes

hillu:
while playing with the RPM client package on CentOS 7 and SuSE Enterprise Linux machines, I noticed that the interrogation feature failed at several stages. None of the /etc/*release files exists on SuSE and /lost+found does not exist when the rootFS is either XFS or BTRFS.

Both patches are untested.

### DIFF
--- a/grr/client/client_actions/linux/linux.py
+++ b/grr/client/client_actions/linux/linux.py
@@ -169,7 +169,13 @@ class GetInstallDate(actions.ActionPlugin):
   out_rdfvalues = [rdf_protodict.DataBlob]
 
   def Run(self, unused_args):
-    self.SendReply(integer=int(os.stat("/lost+found").st_ctime))
+    ctime = 0
+    for path in ("/lost+found", "/"):
+      try:
+        ctime = os.stat(path).st_ctime
+      except OSError:
+        continue
+    self.SendReply(integer=int(ctime))
 
 
 class UtmpStruct(utils.Struct):
--- a/grr/parsers/linux_release_parser.py
+++ b/grr/parsers/linux_release_parser.py
@@ -87,6 +87,35 @@ class LsbReleaseParseHandler(ReleaseParseHandler):
     return complete, ParsedRelease(name, major, minor)
 
 
+class OsReleaseParseHandler(ReleaseParseHandler):
+  """Parse /etc/os-release file."""
+
+  def Parse(self):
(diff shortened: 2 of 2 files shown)


grr/client/client_actions/linux/linux.py
```diff
@@ -169,7 +169,13 @@ class GetInstallDate(actions.ActionPlugin):
   out_rdfvalues = [rdf_protodict.DataBlob]
 
   def Run(self, unused_args):
-    self.SendReply(integer=int(os.stat("/lost+found").st_ctime))
+    ctime = 0
+    for path in ("/lost+found", "/"):
```
destijl: there might be a better choice than / here, that's less likely to have had it's ctime changed after install. We should at least put a comment in about the edge conditions we know about (filesystems missing /lost+found).

hillu:
Hi Greg,

> there might be a better choice than / here, that's less likely to have
> had it's ctime changed after install.

Okay, how about using minimum of the ctimes of several files or
directories?

Here are some ideas for other candidates:

On "modern" Linux systems, /etc/machine-is created at or around
installation time and it is not supposed to be changed.

Installers for Linux distributions usually put log files on the
installed system. For the Debian installer, those are put into
/var/log/installer, on CentOS I have seen /var/log/anaconda. There is
/var/log/YaST2/ on SuSE.

Cheers,
-Hilko

grr/parsers/linux_release_parser.py
```diff
@@ -87,6 +87,35 @@ def Parse(self):
     return complete, ParsedRelease(name, major, minor)
 
 
+class OsReleaseParseHandler(ReleaseParseHandler):
```
destijl: I thought I put this comment in, but maybe it got lost. We definitely need a test case with an example file for this parser. I know you've copied the style of the one above but that one should probably have included a comment to the effect of "this is hacky and fragile, and needs to be rewritten properly" :) The parser is really tightly coupled to the exact current format of the file, if it changes slightly the code breaks, and then we probably need to support multiple formats somehow.

Since all we want to do here is pull out the name and version id I'd be inclined to just write regexes for those rather than attempting to parse the structure.

destijl:
Taking the minimum time of a few likely directories sounds reasonable.

destijl:
Handing the rest of this review over to @mbushkov

Issue #480, opened 2017-03-07
Creating an empty user ("") breaks the user database


hillu:
Starting off with an installation of the Debian package (grr-server 3.1.0.2+dfsg-1), configured to use the SQLite3 datastore and the default admin user as set up by `grr_config_updater`, let's check the user database:
```
$ grr_config_updater show_user
Using configuration <GrrConfigManager  file="/usr/share/grr-server/install_data/etc/grr-server.yaml"  file="/etc/grr/server.local.yaml" >

Username: admin
Labels: admin
Password: set
$ 
```
I create a user name called <empty string>:
```
$ grr_config_updater add_user ""
Using configuration <GrrConfigManager  file="/usr/share/grr-server/install_data/etc/grr-server.yaml"  file="/etc/grr/server.local.yaml" >
Please enter password for user '': 
Added user .
```
Now the user database is broken:
```
$ grr_config_updater show_user
Using configuration <GrrConfigManager  file="/usr/share/grr-server/install_data/etc/grr-server.yaml"  file="/etc/grr/server.local.yaml" >
Traceback (most recent call last):
  File "/usr/bin/grr_config_updater", line 11, in <module>
    load_entry_point('grr-response-server==3.1.0.post2', 'console_scripts', 'grr_config_updater')()
  File "/usr/lib/python2.7/dist-packages/grr/lib/distro_entry.py", line 75, in ConfigUpdater
    flags.StartMain(config_updater.main)
  File "/usr/lib/python2.7/dist-packages/grr/lib/flags.py", line 121, in StartMain
    main([sys.argv[0]])
  File "/usr/lib/python2.7/dist-packages/grr/tools/config_updater.py", line 925, in main
    ShowUser(flags.FLAGS.username, token=token)
  File "/usr/lib/python2.7/dist-packages/grr/tools/config_updater.py", line 265, in ShowUser
    for user in fd.OpenChildren():
AttributeError: 'GRRUser' object has no attribute 'OpenChildren'
```
I seem to be able to remove that <empty string> user:
```
$ grr_config_updater delete_user ""
Using configuration <GrrConfigManager  file="/usr/share/grr-server/install_data/etc/grr-server.yaml"  file="/etc/grr/server.local.yaml" >
User  has been deleted.
```
But wait! Now, my admin user is also gone:
```
$ grr_config_updater show_user
[...]

PR #481, opened 2017-03-08
maintenance_utils: Disallow empty user names

hillu:
This fixes the problem described #480.

### DIFF
--- a/grr/lib/maintenance_utils.py
+++ b/grr/lib/maintenance_utils.py
@@ -327,6 +327,9 @@ def ShowUser(username, token=None):
 
 def AddUser(username, password=None, labels=None, token=None):
   """Implementation of the add_user command."""
+  if not username:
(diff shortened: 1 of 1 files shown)


grr/lib/maintenance_utils.py
```diff
@@ -327,6 +327,9 @@ def ShowUser(username, token=None):
 
 def AddUser(username, password=None, labels=None, token=None):
   """Implementation of the add_user command."""
+  if username is None or username == "":
```
grrrrrrrrr: nit: You can combine those branches by testing like this:
if not username:

hillu:
* Andreas Moser:

> nit: You can combine those branches by testing like this:
> if not username:

Thanks. Updated.

Cheers,
-Hilko

destijl:
LGTM, the travis failure is unrelated. I'll merge this shortly.

Issue #490, opened 2017-04-24
Remote User not Recognised


hannahshort:
Hello, 

I've set up a GRR instance behind Apache as a reverse proxy with Shibboleth authentication. My META variables are being set correctly, including HTTP_REMOTE_USER, but I am still being prompted for local credentials after authenticating through Shibboleth. 

Reading https://github.com/google/grr-doc/blob/master/admin.adoc#authentication-to-the-admin-ui, I got the impression that it shouldn't be necessary to create a new WebAuth Manager. 

Am I missing something or is a new WebAuth Manager required? Happy to contribute code as needed.
Cheers,
Hannah

mbushkov:
By default GRR has BasicWebAuthManager enabled:
https://github.com/google/grr/blob/master/grr/gui/webauth.py#L53

This plugin expects "Authorization: Basic XXXXXX" header to be present and then checks that provided password is equal to the password set in the user object. This is likely not the behaviour you want when you're using external authentication system (Shibboleth in your case). I.e. you don't want GRR to do any password management, but just to pick up the user name that Shibboleth reports.

Seems like the way to go is to write a custom WebAuthManager that will pick the user name from the appropriate source (HTTP header, or env variable). It probably makes sense to use BasicWebAuthManager or FirebaseWebAuthManager as examples.

hannahshort:
Ok, that's the path I've gone down already. If I end up with something generic I'll try and contribute back. 
Cheers :)

mbushkov:
Sounds good, thanks! I'll close the issue for now. Feel free to open a new one in case you have issues.

PR #494, opened 2017-05-02
Http shib

hannahshort:
#490 
Adding web auth manager to accept remote user. 

This helps deployment recommendation from the admin guide: https://github.com/google/grr-doc/blob/master/admin.adoc#authentication-to-the-admin-ui

### DIFF
--- a/grr/gui/webauth.py
+++ b/grr/gui/webauth.py
@@ -101,8 +101,9 @@ class HTTPWebAuthManager(BaseWebAuthManager):
     request.user = request.META.get("HTTP_REMOTE_USER")
     
     if not request.user :
-      result = http.HttpResponse("Unauthorized", status=401)
-      result["WWW-Authenticate"] = "Basic realm='Secure Area'"
+      result = http.HttpResponse("Internal Server Error", status=500)
+      logging.info("Service misconfigured, authentication should be handled by a "
+         "reverse proxy before reaching GRR")
       return result
 
     # Modify this to implement additional checking (e.g. enforce SSL).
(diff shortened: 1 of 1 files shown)


hannahshort:
I signed it

grr/gui/webauth.py
```diff
@@ -88,6 +88,29 @@ def SecurityCheck(self, func, request, *args, **kwargs):
...
+  """Manager using external, HTTP based authention (e.g. 
+  shibboleth on apache as a reverse proxy in front of GRR)
+  This Manager assumes that access through the proxy is 
+  well controlled.
+  """
+
+  def SecurityCheck(self, func, request, *args, **kwargs):
+    """Wrapping function."""
+   
+    request.user = request.META.get("HTTP_REMOTE_USER")
```
mbushkov: I have doubts about this line: request is supposed to be a Werkzeug request object (see https://github.com/google/grr/blob/master/grr/gui/wsgiapp.py#L121) and consequently it shouldn't have the "META" attribute. "META" is part of Django's request objects, but we removed Django request a few months ago. Have you tested this patch with GRR HEAD?

grr/gui/webauth.py
```diff
@@ -88,6 +88,29 @@ def SecurityCheck(self, func, request, *args, **kwargs):
     return response
 
 
+class HTTPWebAuthManager(BaseWebAuthManager):
```
mbushkov: Do you mind adding a simple test for this manager to https://github.com/google/grr/blob/master/grr/gui/webauth_test.py?

PR #502, opened 2017-05-19
Add PATH var tweak to Dockerfile

### LINKED ISSUE #501: Latest Docker Image PATH
Not sure where to put this but the binaries for the tools for [documented stuff like repacking clients](https://github.com/google/grr-doc/blob/master/admin.adoc#repacking-the-client-with-a-new-configuration) isn't actually set properly in the latest Docker image. Probably a minor path issue or something. I'll look at the Dockerfile in the repo after this and if it's trivial I'll possibly issue a PR.

Pulled from `grr:latest`:
Entering with `docker exec -it $containerid /bin/bash`
```
root@013b037cef23:/# grr_config_updater repack_clients
bash: grr_config_updater: command not found
root@013b037cef23:/# /usr/share/grr-server/bin/grr_config_updater repack_clients                  
Using configuration <GrrConfigManager  file="/usr/share/grr-server/install_data/etc/grr-server.yaml"  file="/usr/share/grr-server/install_data/etc/server.local.yaml" >
```

mike-bailey:
Closes #501 
Issue is explained at the issue.

To justify why I did it like this:

- PATH precedence means if things get mucked up it can use, for example, pip in the bin directory but it won't select it by default unless it can't find it in the existing PATH
- Symlinking is a pain for this many tools and I don't know for sure which the maintainers would want accessible and which wouldn't
- It works in a default `sh` which is what the Dockerfile is run under

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,10 @@ pip install --find-links=/wheelhouse /sdists/core/*.tar.gz && \
 pip install --find-links=/wheelhouse /sdists/server/*.tar.gz && \
 pip install --find-links=/wheelhouse grr_response_templates
 
+# Add bin path to PATH so it can run commands like grr_config_updater or grr_console
+RUN export PATH=$PATH:/usr/share/grr-server/bin/
+RUN echo 'export PATH=$PATH:/usr/share/grr-server/bin/' >> /root/.profile
+
 COPY scripts/docker-entrypoint.sh /
 
 ENTRYPOINT ["/docker-entrypoint.sh"]


mbushkov:
Hi Mike! Thanks for the PR.

In Docker GRR gets installed into a virtualenv in /usr/share/grr-server. Given that, it seems that a cleaner way to setup the PATH correctly is:
```
source /usr/share/grr-server/bin/activate
```

I think that maybe updating grr-doc with a hint mentioning the command above is better than writing PATH into .profile. My biggest concern with the .profile approach is that, while it works, it's not a standard way of running binaries installed into a virtualenv.

What do you think?

mike-bailey:
I wholeheartedly agree. I felt like it was an issue but I didn't know how to actually patch best.

mbushkov:
I updated the docs: https://github.com/google/grr-doc/commit/49392483a2f1b9de6165a87e04d2665ee91950d5

Closing this PR.

mike-bailey:
Thanks!

PR #509, opened 2017-06-06
Bugfix frontend service

jawilson0502:
Found this bug when creating front end servers:
Line 15 is how GRR currently sets `max_port`, Line 16 is how I want to set it
```
In [15]: config_lib.CONFIG.Get("Frontend.port_max", config_lib.CONFIG["Frontend.bind_port"])
Out[15]: 4L

In [16]: config_lib.CONFIG["Frontend.port_max"] + config_lib.CONFIG["Frontend.bind_port"]
Out[16]: 8084L

In [17]: config_lib.CONFIG["Frontend.port_max"]
Out[17]: 4L

In [18]: config_lib.CONFIG["Frontend.bind_port"]
Out[18]: 8080L
```

### DIFF
--- a/grr/gui/admin_ui.py
+++ b/grr/gui/admin_ui.py
@@ -52,8 +52,8 @@ def main(_):
     # Address looks like an IPv4 address.
     ThreadedServer.address_family = socket.AF_INET
 
-  max_port = config_lib.CONFIG.Get("AdminUI.port_max",
-                                   config_lib.CONFIG["AdminUI.port"])
+  max_port = (config_lib.CONFIG["AdminUI.port_max"] +
+              config_lib.CONFIG["AdminUI.port"])
 
   for port in range(config_lib.CONFIG["AdminUI.port"], max_port + 1):
     # Make a simple reference implementation WSGI server
--- a/grr/tools/frontend.py
+++ b/grr/tools/frontend.py
@@ -281,8 +281,8 @@ class GRRHTTPServer(SocketServer.ThreadingMixIn, BaseHTTPServer.HTTPServer):
 
 def CreateServer(frontend=None):
   """Start frontend http server."""
-  max_port = config_lib.CONFIG.Get("Frontend.port_max",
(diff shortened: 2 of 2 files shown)


jawilson0502:
Went ahead and fixed the same bug in the admin ui as well.

mbushkov:
Jessica, thanks for the pull request!

Quick question about your example: why not just set Frontend.port_max to 8084 in the config file? IIUC, the point of Frontend.port_max is to set a hard limit on port numbers that can be used when Frontend.bind_port is not available. Why is it better to sum Frontend.port_max and Frontend.bind_port instead of using Frontend.port_max value directly?

jawilson0502:
When I read the documentation (https://github.com/google/grr-doc/blob/master/admin.adoc#running-multiple-copies-of-components) it sounds as if I'm supposed to set the number of ports I want to use (such as 4 ports), not as if I'm supposed to set the highest port. 

Perhaps I misunderstood the documentation, which caused me to set the `Frontend.port_max` incorrectly. I don't believe there is a reason that summing the two config values is better than using them directly, other than expectations set in the documentation.

mbushkov:
Jessica, thanks for your reply! Judging by the comment here https://github.com/google/grr/blob/b2a38b60b2ab70ea785660e3561f918b1d8b7f2c/grr/config/server.py#L145, Frontend.port_max is supposed to be the upper range.

The documentation is somewhat vague indeed, it says:

> then set Frontend.port_max so that you have a range of 10 ports from Frontend.bind_port

I guess this means that in the described case Frontend.port_max should be Frontend.bind_port + 10.

If you don't feel strongly about changing the current Frontend.port_max behavior, maybe it'd be better just to add a little example to the documentation to remove the ambiguity.

jawilson0502:
Added an example to the documentation in https://github.com/google/grr-doc/pull/77

mbushkov:
Thanks! I merged in the documentation example. Closing this PR.

Issue #530, opened 2017-07-17
API: Inconsistent use of `AnyValue`, `bytes` for generic message fields


hillu:
Among the API messages defined in the *.proto files of GRR 3.1.0.2, generic fields are usually defined using `AnyValue` messages which contain type information and a `bytes` buffer. This seems to be the case for all messages that travel from the GRR server to the client as part of `Api*Response` messages, but some message types use a simple `bytes` buffer with no space for type information.

For instance, the 

- `CreateFlow` (POST `/api/clients/<client_id>/flows`) expects `ApiCreateFlowArgs` -> `ApiFlow`: `args` is an `AnyValue`.

but:

- `CreateHunt` (POST `/api/hunts`) expects `ApiCreateHuntArgs` -> `GenericHuntArgs`: `flow_args` is a `bytes`

Shouldn't those be changed to `AnyValue`?

Other examples (not sure how they are used):
- `CreateCronJobFlowArgs`: `flow_args` is a `bytes`
- `FlowRequest` `args` is a `bytes`.
- `ClientActionArgs`: `action_args` is a `bytes`
- `ConsoleDebugFlowArgs`: `flow_args` is a `bytes`.

mbushkov:
Hi Hilko! Thanks for asking - this is an interesting question.

Historically, we used to store dynamic data (i.e. arbitrary protobufs) as bytes. We were annotating corresponding protobuf fields with sem_type {dynamic_type = "..." } annotation. "dynamic_type" was pointing to a Python function that had to be called to find out the type of the data stored in the field.

As soon as we started working on the API, limitations of such approach became apparent. It was impossible to read such protobufs not in Python, and even when using Python one had to depend on GRR source code to do the parsing. By that time a google.protobuf.Any (https://github.com/google/protobuf/blob/master/src/google/protobuf/any.proto#L112) was introduced, so in the API layer instead of storing dynamic fields in bytes, we started to store them as Any messages (in the beginning we were using AnyValue definition, which was a copy of google.protobuf.Any).

As a result, all the API-related data structures have dynamic fields defined with google.protobuf.Any. But whenever you have to deal with the older non-API-related protobufs, you hit the issue of dynamic fields being stored in bytes instead of google.protobuf.Any/AnyValue. 

These older data structures can be updated to use google.protobuf.Any, but this change would be backwards-incompatible and has to be done with care. I'll look into deprecating the "bytes" approach and using google.protobuf.Any for everything.

[...]

PR #536, opened 2017-08-30
Use google.protobuf.Any for argument values

hillu:
This is my attempt at taking care of #530 before 3.2.0 is released. Unfortunately, I don't know if this is enough and I don't really have a way to test. If this is not enough, please give me directions so I can work on completing this in the next few days.

Thanks,
-Hilko

### DIFF
--- a/grr/proto/flows.proto
+++ b/grr/proto/flows.proto
@@ -4,6 +4,8 @@
 
 syntax = "proto2";
 
+import "google/protobuf/any.proto";
+
 import "grr/proto/jobs.proto";
 import "grr/proto/knowledge_base.proto";
 import "grr/proto/output_plugin.proto";
@@ -416,7 +418,7 @@ message FlowRequest {
       description: "The clients to start the flow on.",
     }];
 
-  optional bytes args = 5 [(sem_type) = {
+  optional google.protobuf.Any args = 5 [(sem_type) = {
       dynamic_type: "GetFlowArgsClass",
       description: "Args to the flow. These depend on the flow_name parameter."
       "It is expected that this field contains an instance of flow's arg_type.",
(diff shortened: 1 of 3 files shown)


mbushkov:
Hilko, thanks for the PR.

The changes you made are ok, but only if backwards compatibility is not a concern. If the changes are done in a way they're done in this PR, you won't be able to use new GRR server with any of the older clients and most likely you'll have troubles inspecting older hunts and cron jobs.

While datastore compatibility is a lesser concern, breaking compatibility with older clients is something that we generally tend to avoid.

The proper way to solve the issue would be to add a new field to each protobuf in question. This field should have the same name as the old one. The old one should be renamed to DEPRECATED_[field_name]. 

After that changes in https://github.com/google/grr/blob/master/grr/lib/rdfvalues/structs.py#L988 have to be made so that a new field will take the data from the deprecated field if the data are passed from the older clients. Similarly, when a value is assigned to a new field it has to duplicate it into the DEPRECATED_[field_name] field to maintain backwards compatibility. Most likely we have to add a kind of "legacy_field" annotation to dynamic fields that we have to migrate, so that they know that they have to look into DEPRECATED_[field_name] field for data.

All these changes are not too complicated but doing them without GRR development experience may take time, since you'd have to touch the internals of key GRR components (protos and RDFValues).

As far as I understand, you want to make sure that the proto changes get into the release, so that you can use the API to inspect the data. Good news is that we have changed our release process significantly and have mostly automated it. We intend to do releases periodically (every 3 months), but we also have an opportunity now to cut an out-of-band release if needed.

We can probably do the changes to migrate legacy dynamic fields to google.protobuf.Any within a few weeks. We can then cut a new release candidate and make a new release. Would that be ok for you?

hillu:
Thank you for your detailed explanation and advice. To me, datastore compatibility is just as big a concern as API compatibility. I'll try to implement the the compatibility layer and update this PR in the next few days.

mbushkov:
Hilko, you're welcome! Please feel free to ask questions about the current implementation and about technologies involved (i.e. protobufs). We'll be happy to answer and to help.

mbushkov:
Hi Hilko! Jfyi: we released a 3.2.0.1 GRR version (https://github.com/google/grr/releases/tag/v3.2.0.1). As I said earlier, no worries about the google.protobuf.Any changes not being included into this release - we'll cut a new release as soon as they're in.

hillu:
I haven't been interested in this PR for quite some time, so I am closing it.

PR #620, opened 2018-08-27
Adding config-variable "smtp_mailuser"

b2az:
Adding the config-variable "smtp_mailuser", because in some environments it's not always possible, that the alerting-mail sender is 'grr-noreply'


grr/server/grr_response_server/output_plugins/email_plugin.py
```diff
@@ -89,10 +89,15 @@ def ProcessResponse(self, response):
         signature=config.CONFIG["Email.signature"],
         hostname=utils.SmartUnicode(hostname),
         creator=utils.SmartUnicode(self.token.username))
-
+    
+    if config.CONFIG["Worker.smtp_mailuser"]:
```
mrhania: Instead of this `if`-statement you should specify a default value when you define config option. It also helps to avoid code duplication with the other file.

PR #648, opened 2018-11-22
Make grr-server components a part of main grr-server.service unit

mback2k:
This makes the instantiation of components more explicit.

Instead of implicitly launching components via systemctl they
can be created as symlinks to or copies of the template on disk.

### DIFF
--- a/debian/grr-server.service
+++ b/debian/grr-server.service
@@ -3,15 +3,16 @@
 
 [Unit]
 Description=GRR Service
-After=syslog.target network.target
+After=network.target
+After=syslog.target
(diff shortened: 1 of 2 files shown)


grrrrrrrrr:
Hey Marc,
Thanks for your PRs! We can only accept patches from developers who sign the CLA, would that be possible for you?

mback2k:
@grrrrrrrrr Unfortunately I think I would only be able to sign an individual CLA which won't apply to these (simple) changes since they originated within https://github.com/TKCERT.

PR #655, opened 2019-01-01
Fix mysql database fail to start due to permissions issues.

iwm911:
GRR container fail to start because of a MySQL permissions error on '/var/lib/mysql' (the MySQL fail to start).
mounting the path as a volume fix's the problem.

inspired by https://github.com/geerlingguy/drupal-vm/issues/1497

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,6 +76,6 @@ EXPOSE 8080
 # Directories used by GRR at runtime, which can be mounted from the host's
 # filesystem. Note that volumes can be mounted even if they do not appear in
 # this list.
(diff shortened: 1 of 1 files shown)


iwm911:
I signed it!

demonchild2112:
Did you try starting the container with the instructions provided in [GRR documentation](https://grr-doc.readthedocs.io/en/v3.2.4/installing-grr-server/via-docker.html)?

It is not always the intent to mount `/var/lib/mysql` from the host when running the GRR container. Does using the `--volume` flag instead work for you?

PR #662, opened 2019-01-16
[IAP] Add IAP Web Auth Support

4ndygu:
This PR Adds Authentication support from Google's IAP service. The extension pulls the x-google-iap-jwt-assertion header from IAP and generates a new user for that header via the 'sub' claim. Authorization is delegated via IAP.

### DIFF
--- a/grr/core/grr_response_core/config/gui.py
+++ b/grr/core/grr_response_core/config/gui.py
@@ -142,11 +142,11 @@ config_lib.DEFINE_bool(
 
 # Configuration requirements for Cloud IAP Setup
 config_lib.DEFINE_string(
-    "AdminUI.cloud_project_id", None,
+    "AdminUI.google_cloud_project_id", None,
     "Cloud Project ID for IAP. This must be set if the IAPWebAuthManager is used."
 )
 
 config_lib.DEFINE_string(
-    "AdminUI.cloud_backend_service_id", None,
+    "AdminUI.google_cloud_backend_service_id", None,
     "GCP Cloud Backend Service ID for IAP. This must be set if the IAPWebAuthManager is used."
 )
--- a/grr/core/grr_response_core/config/gui.py
+++ b/grr/core/grr_response_core/config/gui.py
@@ -139,3 +139,14 @@ config_lib.DEFINE_bool(
     "When running in headless mode, AdminUI ignores checks for JS/CSS compiled "
     "bundles being present. AdminUI.headless=True should be used to run "
     "the AdminUI as an API endpoint only.")
+
+# Configuration requirements for Cloud IAP Setup
+config_lib.DEFINE_string(
+    "AdminUI.cloud_project_id", None,
+    "Cloud Project ID for IAP. This must be set if the IAPWebAuthManager is used."
+)
+
+config_lib.DEFINE_string(
+    "AdminUI.cloud_backend_service_id", None,
+    "GCP Cloud Backend Service ID for IAP. This must be set if the IAPWebAuthManager is used."
+)
--- a/grr/server/grr_response_server/gui/validate_iap.py
+++ b/grr/server/grr_response_server/gui/validate_iap.py
@@ -1,17 +1,3 @@
-# Copyright 2016 Google Inc. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 """Sample showing how to validate the Identity-Aware Proxy (IAP) JWT.
 
 This code should be used by applications in Google Compute Engine-based
@@ -22,7 +8,6 @@ of assurance that a request was authorized by IAP.
 For applications running in the App Engine standard environment, use
 App Engine's Users API instead.
 """
-# [START iap_validate_jwt]
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import unicode_literals
@@ -94,5 +79,3 @@ def GetIapKey(key_id):
         if not key:
             raise Exception('Key {!r} not found'.format(key_id))
     return key
-
-# [END iap_validate_jwt]
--- a/grr/server/grr_response_server/gui/validate_iap.py
+++ b/grr/server/grr_response_server/gui/validate_iap.py
@@ -1,0 +1,93 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Sample showing how to validate the Identity-Aware Proxy (IAP) JWT.
+
+This code should be used by applications in Google Compute Engine-based
+environments (such as Google App Engine flexible environment, Google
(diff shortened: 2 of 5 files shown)


grr/server/grr_response_server/gui/webauth.py
```diff
@@ -51,6 +53,57 @@ def RedirectBase(self):
...
+                         "not set. Please set your AdminUI.cloud_project_id"
+                         "or AdminUI.cloud_backend_service_id keys.")
+
+    self.cloud_project_id = config.CONFIG["AdminUI.cloud_project_id"]
+    self.backend_service_id = config.CONFIG["AdminUI.cloud_backend_service_id"]
+    self.iap_header = "x-goog-iap-jwt-assertion"
+
+  def SecurityCheck(self, func, request, *args, **kwargs):
+    """Wrapping function."""
+    request.user = u""
```
mbushkov: I don't think this is needed. If the user can't be authenticated, request.user can be simply left unset.

grr/server/grr_response_server/gui/webauth.py
```diff
@@ -51,6 +53,57 @@ def RedirectBase(self):
...
+      # Return failure if IAP is not decoded correctly
+      return werkzeug_wrappers.Response("Unauthorized", status=401)
+    else:
+      # Generate a new user if not created, else authenticate current
+      try:
+        grr_user = config_updater_util.GetUserSummary(user_id)
+      except config_updater_util.UserNotFoundError:
+        grr_user = None
+
+      if grr_user:
```
mbushkov: I think just doing `request.user = user_id` should be enough here. GRR will create the user automatically, you don't have to do it in the webauth manager.

4ndygu: I seeded each user with a long random password so a user who is logged in can't switch users to another user who has logged in via IAP, thus obfuscating logging. Do you think this will be an issue if I resolve this comment?

grr/core/grr_response_core/config/gui.py
```diff
@@ -139,3 +139,14 @@
     "When running in headless mode, AdminUI ignores checks for JS/CSS compiled "
     "bundles being present. AdminUI.headless=True should be used to run "
     "the AdminUI as an API endpoint only.")
+
+# Configuration requirements for Cloud IAP Setup
+config_lib.DEFINE_string(
+    "AdminUI.cloud_project_id", None,
```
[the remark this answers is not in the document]
4ndygu: 👍

grr/core/grr_response_core/config/gui.py
```diff
@@ -139,3 +139,14 @@
...
+    "Cloud Project ID for IAP. This must be set if the IAPWebAuthManager is used."
+)
+
+config_lib.DEFINE_string(
+    "AdminUI.cloud_backend_service_id", None,
```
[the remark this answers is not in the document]
4ndygu: 👍

grr/server/grr_response_server/gui/validate_iap.py
```diff
@@ -0,0 +1,93 @@
...
+    except (jwt.exceptions.InvalidTokenError,
+            requests.exceptions.RequestException) as e:
+        return (None, None, '**ERROR: JWT validation error {}**'.format(e))
+
+
+def GetIapKey(key_id):
+    """Retrieves a public key from the list published by Identity-Aware Proxy,
+    re-fetching the key file if necessary.
+    """
+    key_cache = GetIapKey.key_cache
```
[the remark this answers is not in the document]
4ndygu: Makes sense to me. For my own learning, is there a standard that makes the module-level variable a better pattern than the current implementation?

grr/server/grr_response_server/gui/webauth.py
4ndygu: 👍

grr/server/grr_response_server/gui/validate_iap.py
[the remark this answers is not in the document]
mbushkov: I don't think there's a standard, but rather a convention.
* If you use a module-level variable, things are very explicit. There's a singleton variable \_KEY_CACHE (by convention global variables/constants are named in uppercase), it's defined with a default value and it's intended to be used in the module only (because of the "_" prefix).
* If you use the GetApiKey.key_cache approach, the mechanics under the hood are somewhat more complicated. Effectively, you're assigning an attribute to a function object (you can do this because everything is an object in Python). When I see GetIapKey.key_cache, I first assume that GetIapKey is a class (because of an attribute access) and it's a bit surprising to see that it's actually a function. I need to do a few hops in the source code to understand what's going on. Mostly because assigning custom attributes to a function is something that you rarely (if ever) need to do a Python.

grr/server/grr_response_server/gui/webauth.py
mbushkov: Ok, I think I understand the scenario. Quick question - do you plan to have 2 AdminUI's running, one for basic http-based auth and one for IAP? Because at the moment GRR AdminUI can have only one WebAuth manager set, meaning that a single AdminUI instance would be able to run with *either* IAP or HTTP Basic auth.

grr/server/grr_response_server/gui/validate_iap.py
```diff
@@ -0,0 +1,93 @@
...
+"""Sample showing how to validate the Identity-Aware Proxy (IAP) JWT.
+
+This code should be used by applications in Google Compute Engine-based
+environments (such as Google App Engine flexible environment, Google
+Compute Engine, or Google Container Engine) to provide an extra layer
+of assurance that a request was authorized by IAP.
+
+For applications running in the App Engine standard environment, use
+App Engine's Users API instead.
+"""
```
mrhania: I have just a minor request: we are trying to make our codebase compatible with Python 3. One of the things we do to ensure that is adding [`__future__` imports](https://docs.python.org/2/library/__future__.html). In all new files we try to at have for `absolute_import`, `division` and `unicode_literals` (for example, see [this module](https://github.com/google/grr/blob/8c94f4f836b3944c7cebc08a8416cbd0c2a0ad21/grr/core/grr_response_core/lib/util/compatibility.py#L3)).

This of course does not of fully enforce Python 3 compatiblity, but at least adds some level of convidence and brings us one step closer to our the goal.

grr/server/grr_response_server/gui/webauth.py
mrhania: Just a tip for the future: this module has `from __future__ import unicode_literals` import which means that unicode literals are enabled by default - the `u` prefix in this empty string is redundant.

grr/server/grr_response_server/gui/validate_iap.py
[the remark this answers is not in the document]
4ndygu: 👍

grr/server/grr_response_server/gui/webauth.py
4ndygu: Makes sense to me!

grr/server/grr_response_server/gui/webauth.py
4ndygu: Ah, that makes sense. We have two admin UIs loaded in our implementation, but they are ostensibly going to deploy the same Docker image. I think what you point out here means that we wouldn't need to generate a new seeded password each time. I'll do some testing of new account creation to validate, and push your suggestion.

grr/server/grr_response_server/gui/validate_iap.py
4ndygu: 👍

grr/server/grr_response_server/gui/validate_iap.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
mbushkov: One question: GRR is already coverted by the Apache license (https://github.com/google/grr/blob/master/LICENSE). Is there any specific reason to insert the license header into this file?

4ndygu: This may be a misinterpretation on my part, but this mini-library of functions is taken from the Google sample code on validating IAP headers. I placed this here to emphasize its distinct status, but I may be violating some standards here. Do you think we should replace this blob?

mbushkov: Given that GRR is also a Google project completely covered by the same license, I think it's safe to assume that an explicit copy of the license is not needed here.

4ndygu:
Thanks for the edits! We'd like to deploy IAP in our network without having to maintain a fork built from source. Is there any way we can rebuild and cut a new release image?

mbushkov:
@OmarDarwish , apparently it was a transient issue with PIP. I triggered rebuilt and it has succeeded:
https://hub.docker.com/r/grrdocker/grr/builds

PR #663, opened 2019-01-24
Travis CI: 'sudo' tag is now deprecated

cclauss:
[Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration).


demonchild2112:
Thanks for doing this!

PR #675, opened 2019-03-17
FDE done with some existing limitations

ehossam:


mbushkov:
Hi @ehossam ! I'm not sure what's the intended purpose of this Pull Request is as it lacks the description. However, there are a few major issues about it that I can see right away:

- it includes thousands of files from gui/static/node_modules
- it includes prebuilt GRR binaries
- it includes an existing server configuration (with keys)

None of the above should be committed to GRR Git repo, because these things are either build leftovers or user/deployment-specific files. I'll close this pull request for now.

PR #718, opened 2019-07-01
Removing old README

kshithijiyer:


demonchild2112:
This is referenced by the server deb: https://github.com/google/grr/blob/7286fc8d3eebcaf3ace66d06c132c6ce2da998c0/debian/grr-server.docs#L2

This PR will be reverted by a future GRR commit.

kshithijiyer:
Should I resubmit this pr again?

demonchild2112:
@kshithijiyer Since the server deb depends on the README file, an acceptable solution would be to move it to the server deb's build script or `debian/rules` file. At the moment though, having the partially redundant README file next to the mardown README around is just a minor distraction/annoyance.

kshithijiyer:
Could you explain what would be changes needed if we would want to move it to the Debian folder apart from moving the file itself?

demonchild2112:
[dh_installdocs](https://manpages.debian.org/testing/debhelper/dh_installdocs.1.en.html) is supposed to copy the files in [grr-server.docs](https://github.com/google/grr/blob/98acb284be10a5c780d6ae170f86ac18b8064b9a/debian/grr-server.docs#L1) to the target system when the server deb is installed. The server's [debian/rules](https://github.com/google/grr/blob/98acb284be10a5c780d6ae170f86ac18b8064b9a/debian/rules#L28) file however doesn't seem to invoke the debhelper command, so it may not make a difference if the [docs](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#docs) file is deleted, along with the README. I'll see if the server deb builds and installs successfully with the files deleted, and if it does, I will remove the files from the repo myself.

kshithijiyer:
Okay cool! Thanks for sharing the knowledge with me.

PR #744, opened 2019-10-31
Line 159 webauth.py - added source IP of the authentication attempt t…

dotalchemy:
To assist with troubleshooting remote auth issues - in my case, we were having issues with the F5's sending through requests in an odd format of ::ffff:10.x.x.x, whereas pcaps were showing the IPv4 10.x.x.x address. Adding this to our webauth.py so the source request address was displayed in the error let us figure out what was going on pretty quick.

This change isn't exposing trusted addresses for GRR - anyone able to connect directly to GRR would just see their own IP shown as source.

### DIFF
--- a/grr/server/grr_response_server/gui/webauth.py
+++ b/grr/server/grr_response_server/gui/webauth.py
@@ -156,7 +156,7 @@ class RemoteUserWebAuthManager(BaseWebAuthManager):
   def SecurityCheck(self, func, request, *args, **kwargs):
     if request.remote_addr not in self.trusted_ips:
       return self.AuthError("Request sent from an IP not in "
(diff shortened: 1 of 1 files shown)


dotalchemy:
@googlebot I signed it!

mbushkov:
This change apparently required some test fixes. Unfortunately, the tests weren't triggered by this PR (something we should fix). Submitted the suggested change and fixed the tests in  https://github.com/google/grr/commit/49345353d301a4c5190df8b1c1b8a5544929ae6e. See https://github.com/google/grr/blob/49345353d301a4c5190df8b1c1b8a5544929ae6e/grr/server/grr_response_server/gui/webauth.py#L160

PR #783, opened 2020-06-15
Recreated Grafana dashboards with grafanalib

tsehori:
Instead of adding [Grafana dashboards](https://grafana.com/docs/grafana/latest/features/dashboard/dashboards/) in primitive JSON files, which are hard to read and modify, we will introduce sample dashboards implemented in Python that use the module [grafanalib](https://github.com/weaveworks/grafanalib). This module "compiles" the dashboards written in Python to JSON files that can be later imported natively into Grafana.

- Created a new package for the sample dashboards, [grr-grafanalib-dashboards](monitoring/grafana/setup.py) which embodies the Grafana dashboards and the logic behind them, that can be reused for enhancements and new dashboards and graphs. The configurations for the package can be found in the corresponding [setup](monitoring/grafana/setup.py) file.
- The currently implemented dashboards, [frontends.dashboard.py](monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py) and [workers.dashboard.py](monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/workers.dashboard.py), have some dashboards shared between them that is implemented in [reusable_panels.py](monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py). The choice of which panels should be reused across all dashboards resides in the variable `PANELS` in [config.py](monitoring/grafana/grr_grafanalib_dashboards/config.py).
- Also, the existing dashboards were moved to the folder `dashboards_to_generate`.
- The dashboards share common attributes like a data source, which can be found in [util.py](monitoring/grafana/grr_grafanalib_dashboards/util.py).
- Added a [README](monitoring/grafana/README.md) with steps for a quick start regarding how to set up these sample dashboards in a GRR deployment.
- Re-generated and added the sample dashboards in JSON format, so that if a user wants to quickly import and the use the given dashboards, they can do so without having to install `grr-grafanalib-dashboards` and re-generating the dashboards.


monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
```diff
@@ -0,0 +1,50 @@
+from grafanalib.core import (
+    Dashboard, Graph, Row, Target
+)
```
mbushkov, line 3: Minor style thing: please use the one line syntax:
from grafanalib.core import Dashboard, Graph, Row, Target

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
```diff
@@ -0,0 +1,50 @@
...
+from grr_grafanalib_dashboards.util import add_data_source
+from grr_grafanalib_dashboards.config import PANELS
+
+GRR_COMPONENT = "frontend"
+
+dashboard = Dashboard(
+    title="{}s Dashboard".format(GRR_COMPONENT).title(),
```
mbushkov, line 10: Please use 2 spaces indent - this is the default throughout the project.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
```diff
@@ -0,0 +1,50 @@
...
+            panel(GRR_COMPONENT) for panel in PANELS
+        ]
+        ),
+        Row(panels=[
+          Graph(
+            title="QPS",
+            targets=[
+                Target(
+                    expr='sum(rate({}_request_count_total[1m]))'.format(GRR_COMPONENT),
+                    legendFormat="Requests",
```
mbushkov, line 22: I'm not sure it makes sense format GRR_COMPONENT inside the metric name. We have frontend_request_count_total metric, but is there worker_request_count_total or admin_ui_request_count_total?

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
```diff
@@ -0,0 +1,50 @@
...
+                ),
+            ],
+            ),
+            Graph(
+            title="Request Latency Rate",
+            targets=[
+                Target(
+                    expr='sum(rate({0}_request_latency_sum[5m])) / sum(rate({0}_request_latency_count[5m]))'.format(GRR_COMPONENT),
```
mbushkov, line 30: Please see my comment above. I think it makes sense to name the metric literally.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
```diff
@@ -0,0 +1,50 @@
...
+                ),
+            ],
+            ),
+        ]),
+        Row(panels=[
+            Graph(
+                title="Active Tasks Count",
+                targets=[
+                    Target(
+                        expr='sum({}_active_count)'.format(GRR_COMPONENT),
```
mbushkov, line 41: Same here: we know that we're inside the frontend code and the code is not meant to be reused for other components. Naming the metric literally would make this code easier to read, I think.

Basically, it makes sense to put a string literal into a constant if it's used in multiple places. This way it's easy to change it. But metrics names are fixed and so changing GRR_COMPONENT to a different value (i.e. "GRR Frontend") would just break them.

monitoring/grafana/grr_grafanalib_dashboards/config.py
```diff
@@ -0,0 +1,5 @@
+from grr_grafanalib_dashboards import reusable_panels
+
+DATA_SOURCE = "grr-server"
+ACTIVE_PROCESSES_CONDITION = 1
```
mbushkov, line 4: Please add a comment documenting what this value means. Same for "DATA_SOURCE".

monitoring/grafana/grr_grafanalib_dashboards/config.py
```diff
@@ -0,0 +1,5 @@
...
+PANELS = [reusable_panels.number_of_active_processes_graph, reusable_panels.avg_cpu_usage_percentage]
```
mbushkov, line 5: Would it make sense to just put this constant into reusable_panels module? What do we gain by having it "config"?

monitoring/grafana/grr_grafanalib_dashboards/util.py
```diff
@@ -0,0 +1,7 @@
+from grr_grafanalib_dashboards.config import DATA_SOURCE
+
+def add_data_source(dashboard):
```
mbushkov, line 3: I think it would be cleaner to pass the data source as an argument. This way the utility function would be more narrow-scoped. You can import the DATA_SOURCE in a place where it's used and pass it as an argument.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/frontends.dashboard.py
tsehori: You're right; there aren't. Will get rid of that.

monitoring/grafana/grr_grafanalib_dashboards/config.py
tsehori: I agree that this can go into `reusable_panels.py` straight away, but I think that the variables `DATA_SOURCE` and `ACTIVE_PROCESSES_CONDITION` are not related to the general panels whatsoever, and this is why I kept them in different files.

mbushkov: Totally agreed. Sorry for not being clear. My comment was related to PANELS only.

monitoring/grafana/grr_grafanalib_dashboards/config.py
tsehori: If it's ok, I think comments will clutter it a bit; I'll try to make the variables' names more descriptive, as they are pretty straightforward. Let me know what you think.

monitoring/grafana/grr_grafanalib_dashboards/util.py
tsehori: As of now, `DATA_SOURCE` is used nowhere else in the module; we can pass it as an argument from each `<grr_component>.dashboard.py` file, but then we would have to import `DATA_SOURCE` to there, achieving the same result.
What do you think?

mbushkov: Yes, this is what I'd suggest. Then add_data_source would be a narrow-scoped utility function. Importing config.py in each dashboard also seems like a nice idea. Then for the reader of the dashboard code it'd be immediately clear where the config options are coming from.

monitoring/grafana/grr_grafanalib_dashboards/config.py
mbushkov: ACTIVE_PROCESSES_ALERTING_CONDITION is a better name but it's still confusing. Is this a number of processes? Will it alert if there's one process or less than one process? Comments would be helpful.
Also for DATA_SOURCE, having a comment would help - where does the value is coming from, am I supposed to change it? Understanding this from the comment would be great.

In general, I'd say that if comments clutter the code, that's great - because most of the code is not cluttered by comments and is sometimes hard to read precisely because of that :)

tsehori: You're right; I elaborated in the comments, and also added a relevant line in the monitoring documentation, in [this commit](https://github.com/google/grr-doc/pull/117/commits/b1507d3a95a8b405c3e3db95888e71d9c2eb210b), as the definition of `DATA_SOURCE` depends on how exactly did the user set up Grafana during installation.

monitoring/grafana/grr_grafanalib_dashboards/config.py
```diff
@@ -0,0 +1,12 @@
+from grr_grafanalib_dashboards import reusable_panels
+
+# The data source name is specified after Grafana is set up
+# and it can be visited at localhost:3000.
+# In GRR Monitoring docs, we suggest naming it "grr-server", but if it's
```
mbushkov, line 5: Nit: would it make sense to put a link to the docs you mention here?

monitoring/grafana/grr_grafanalib_dashboards/util.py
```diff
@@ -0,0 +1,5 @@
+def add_data_source(dashboard, datasource):
```
mbushkov, line 1: Please add a short comment explaining what the function does.

monitoring/grafana/grr_grafanalib_dashboards/config.py
tsehori: I will; just note that the documentation exists, but this bit is a part of a [PR](https://github.com/google/grr-doc/pull/117) to grr-doc which isn't merged at the moment.

tsehori:
ptal

monitoring/grafana/grr_grafanalib_dashboards/config.py
mbushkov: Ah, my bad. Merged.

PR #787, opened 2020-06-23
Extended existing dashboards and added new dashboard for AdminUI

tsehori:
This PR extends #777 and #783 with additional panels and graphs for the existing dashboards (for Frontends and Workers components) and an additional new dashboard and panels for the AdminUI components.

- Created graphs that are related to database queries and operations and resource usage (memory and CPU) across all GRR server components.
- Moved some panels from specific GRR server components to `reusable_panels.py` for general use across all components.
- [AdminUI dashboard](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/admin_ui.dashboard.py) added, which consists of graphs related to HTTP API calls along with the general graphs that exist in all dashboards.
- The general panels in `reusable_panels.py` are [re-structured](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py#L115) as apparently, there is an issue in the grafanalib module that implicitly prevents having more than 4 panels per row in a Grafana dashboard. With this new structure, a user may choose which panels to include in which row in the dashboard to be generated, and in which exact order. Also added [a comment](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py#L114) to prevent a modification that will raise the issue by grafanalib.

Please note of the following:
- The [Database Operation Errors panel](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py#L65) does not show data for all GRR server components in my tests; it does for Workers and AdminUI dashboards but not for Frontends. The data for Workers and AdminUI dashboards is triggered by crashing the MySQL instance during the run of [end-to-end tests](https://github.com/google/grr/blob/master/grr/test/grr_response_test/run_end_to_end_tests.py).
- The [GRR Client Crashes panel](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py#L103) shows a flat line in my tests, even when a client does crash. I suspect that the error is not with the query, but that the crash is not registered in the Stats handler. Thus, it will addressed in an issue.
- The [API Calls Count Rate by other statuses (not SUCCESS) panel](https://github.com/tsehori/grr/blob/b8cd73652f9251f54fe059cebc42ca26bd601184/monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/admin_ui.dashboard.py#L34) will *not* show a "spike" in the graph given a single API call (as the first call). This is thanks to the way the [rate() function](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) of Prometheus works. This edge case is documented in the [README](https://github.com/tsehori/grr/blob/4359450331370f1b22bb816fb488a82fc3833c09/monitoring/grafana/README.md) of the grr_grafanalib_dashboards module.


monitoring/grafana/README.md
```diff
@@ -4,6 +4,17 @@ that will help you get started quickly introducing monitoring capabilities to yo
 
 Quick Start
 -----------
+Before generating the dashboards on your own, note that you may find the sample
```
mbushkov, line 7: I'm not a native English speaker, so not sure about this one, but: would it be better to use "can" instead of "may" here? I.e. "you can find" instead of "you may find".

monitoring/grafana/README.md
```diff
@@ -4,6 +4,17 @@ that will help you get started quickly introducing monitoring capabilities to yo
...
 -----------
+Before generating the dashboards on your own, note that you may find the sample
+dashboard in importable JSON format inside the folder
+`grr/monitoring/grafana/grr_grafanalib_dashboards/dashboards_for_use`. From
+there you may import the sample dashboards without re-generating them to your
+Grafana instance. Only if you wish to generate the dashboards, proceed.
+
+**Please note the following edge case**: the panel "API Calls Count Rate by other statuses (not
+  SUCCESS)" inside the AdminUI dashboard will *not* show a spike in the graph
+  given a single HTTP API call, where the status of the call is not SUCCESS.
```
mbushkov, line 15: I'd phrase it slightly differently: "given a single HTTP API call" -> "when there's only one data point present (only a single API call of a given method was made)".

monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py
```diff
@@ -37,6 +37,92 @@ def avg_cpu_usage_percentage(grr_component):
...
+        expr='sum(threadpool_outstanding_tasks{{job="grr_{}"}})'.format(grr_component),
+        legendFormat="Outstanding Tasks",
+      ),
+      Target(
+        expr='sum(threadpool_threads{{job="grr_{}"}})'.format(grr_component),
+        legendFormat="Threads",
+      ),
+    ])
+
+def threadpool_cpu_usa(grr_component):
```
mbushkov, line 89: Typo: usage.

monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py
```diff
@@ -37,6 +37,92 @@ def avg_cpu_usage_percentage(grr_component):
...
+        expr='avg(rate(threadpool_cpu_use{{job="grr_{}"}}[30s])) * 100'.format(grr_component),
+        legendFormat="Average Process CPU Usage in % (over all jobs & pools)",
+      ),
+    ],
+    yAxes=YAxes(
+      left=YAxis(max=105)
+    ),
+  )
+
+def client_crashes(grr_component):
```
mbushkov, line 103: Why is client_crashes reusable?

tsehori: The reason is that this metric is exposed by all GRR server components (Frontends, Workers and AdminUI). Since at the moment I can't manage to trigger the metric, I am not sure which one of them would be triggered (if any).

mbushkov: Got it. Let's expose it on the frontend only - this seems like the most logical place to show crashes graph. I wouldn't show it on other dashboards.

monitoring/grafana/README.md
tsehori: I'm not a native English speaker as well, but it seems that [you're right](https://writingexplained.org/can-vs-may-difference). 😄

tsehori:
ptal

monitoring/grafana/README.md
```diff
@@ -4,6 +4,18 @@ that will help you get started quickly introducing monitoring capabilities to yo
 
 Quick Start
 -----------
+Before generating the dashboards on your own, note that you can find the sample
+dashboard in importable JSON format inside the folder
+`grr/monitoring/grafana/grr_grafanalib_dashboards/dashboards_for_use`. From
+there you may import the sample dashboards without re-generating them to your
```
mbushkov, line 10: One "may" survived :)

monitoring/grafana/README.md
```diff
@@ -4,6 +4,18 @@ that will help you get started quickly introducing monitoring capabilities to yo
 
 Quick Start
 -----------
+Before generating the dashboards on your own, note that you can find the sample
+dashboard in importable JSON format inside the folder
+`grr/monitoring/grafana/grr_grafanalib_dashboards/dashboards_for_use`. From
+there you may import the sample dashboards without re-generating them to your
+Grafana instance. Only if you wish to generate the dashboards, proceed.
```
mbushkov, line 11: I'd make "Only if you wish to generate the dashboards, proceed." more clear: "Follow the instructions below if you want to rebuild or customise the dashboards".

Issue #791, opened 2020-06-25
Timestamp rendering in UI v2


danielberbece:
UI v2 displays timestamps inconsistently and not very readable (e.g. "06/25/2020 11:17:16 UTC" in the client overview panel)

Would be more useful to have a component just for the timestamp, and use it across all the app. Then, if any updates will come, updating the component will also update all references. Also, finding a more elegant way of displaying the timestamp is a priority.

All the current timestamps in the UI should be converted to the new component.

I will prepare a pull-request for this issue

PR #793, opened 2020-07-02
New timestamp module for UIv2

danielberbece:
Added a new timestamp module, as specified in #791


grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (timeDifference < this.oneMin / 2) {
+      return "Just now";
+    }
+
+    if (timeDifference < this.oneHour) {
+      return super.transform(relativeDate, "m'min ago'");
+    }
+
+    if (timeDifference < this.oneDay) {
+      return super.transform(relativeDate, "H'h'm'min ago'");
```
max-vogler, line 37: `relativeDate` is not actually a time difference, but `timeDifference` added onto the unix epoch. Per default, DatePipe displays any Date in the computer's local timezone. Thus, a timeDifference of 1 hour (`timeDifference = 3600000`) can translate to a `relativeDate` of `1970-01-01 02:00:00 GMT+0100`, which in turn would be displayed as `2h0min ago` to the user.

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp_test.ts
```diff
@@ -0,0 +1,106 @@
...
+      .compileComponents();
+  }));
+
+  it('is created successfully', () => {
+    const fixture = TestBed.createComponent(Timestamp);
+    const componentInstance = fixture.componentInstance;
+    expect(componentInstance).toBeTruthy();
+  });
+
+  it('shows "Just now" for dates less than 30s ago', () => {
```
max-vogler, line 31: In tests, generally test for edge cases. For your time handling logic, please test the biggest value below the boundary (29.999s) and exactly the boundary (30s). For added fun, test some very small numbers (e.g. 0), very big numbers (e.g. 1000 years ago), and fractional numbers (e.g. 44.444s).

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp_test.ts
```diff
@@ -0,0 +1,106 @@
...
+    date.setMinutes(date.getMinutes() - 3);
+    date.setHours(date.getHours() - 24);
+
+    componentInstance.date = date;
+    fixture.detectChanges();
+    expect(fixture.nativeElement.querySelector('div').innerText)
+      .toEqual('yesterday at ' + new DatePipe('en').transform(date, 'HH:mm'));
+  });
+
+  it('shows absolute timstamp for dates older than 48 hours', () => {
```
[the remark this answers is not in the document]
danielberbece: Nice catch

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp.ts
```diff
@@ -0,0 +1,16 @@
+import {Component, Input, ViewEncapsulation} from '@angular/core';
+
+/**
+ * Shows a formatted timestamp, based on the date received as parameter
+ */
+@Component({
+  selector: 'timestamp',
+  templateUrl: './timestamp.ng.html',
+  styleUrls: ['./timestamp.scss'],
+  encapsulation: ViewEncapsulation.None
```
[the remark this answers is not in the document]
danielberbece: Not needed actually, I'll remove it

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp.ts
```diff
@@ -0,0 +1,16 @@
+import {Component, Input, ViewEncapsulation} from '@angular/core';
+
+/**
+ * Shows a formatted timestamp, based on the date received as parameter
```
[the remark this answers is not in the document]
danielberbece: Got it.

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp.scss
```diff
@@ -0,0 +1,9 @@
+.timestamp {
+  position: relative;
+  display: inline-block;
+  // border-bottom: 1px dotted rgb(126, 126, 126);
```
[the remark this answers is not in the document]
danielberbece: Sorry, missed this one.

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+  transform(date: Date, absoluteOnly: any): any {
+    if (date === undefined) {
+      return "-";
+    }
+
+    if (absoluteOnly) {
+      return this.getAbsoluteTimestamp(date);
+    }
+
+    var currentDate = new Date();
```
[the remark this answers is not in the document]
danielberbece: I didn't know the differences between `var` and `let`, thanks Max!

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
danielberbece: Sorry about this, you're right!

grr/server/grr_response_server/gui/ui/components/home/search_box.ng.html
```diff
@@ -16,7 +16,7 @@
         <mat-icon *ngIf="!isOnline(client.lastSeenAt)"
                   class="offline-icon">highlight_off</mat-icon>
         <span>{{ client.knowledgeBase.fqdn }}</span> |
-        <small>{{ client.lastSeenAt ? (client.lastSeenAt | date:'short':'UTC') : '-' }}</small>
+        <small><timestamp [date]="client.lastSeenAt" tooltipDisabled="true"></timestamp></small>
```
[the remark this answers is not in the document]
danielberbece: I was inspired by the Angular Material convention, but I think `tooltipEnabled` is actually a better way for our use case.

grr/server/grr_response_server/gui/ui/components/client/module.ts
```diff
@@ -22,6 +23,7 @@ import {ClientRoutingModule} from './routing';
     ApprovalModule,
     FlowFormModule,
     FlowListModule,
+    TimestampModule
```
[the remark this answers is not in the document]
danielberbece: I didn't think that far into the future. Good to know 👍

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,50 @@
...
+  name: 'relativeTimestamp'
+})
+export class RelativeTimestampPipe extends
+  DatePipe implements PipeTransform {
+
+  readonly oneDay = 86400000; // One day in millis
+  readonly oneHour = 3600000; // One hour in millis
+  readonly oneMin = 60000; // One minute in millis
+
+  transform(date: Date, absoluteOnly: any): any {
```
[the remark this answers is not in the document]
danielberbece: I tried using `boolean` instead of `any`, but because the method overrides the one from `DatePipe` class, it didn't allow me. I will find a workaround.

grr/server/grr_response_server/gui/ui/components/client_search/module.ts
```diff
@@ -5,6 +5,7 @@ import {RouterModule} from '@angular/router';
 
 import {ClientSearch} from './client_search';
 import {ClientSearchRoutingModule} from './routing';
+import { TimestampModule } from '../timestamp/module';
```
max-vogler, line 8: Does your formatter format with spaces around import literals `{` and `}`? If so, let's chat about formatter configuration.

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+@Pipe({
+  name: 'relativeTimestamp'
+})
+export class RelativeTimestampPipe implements PipeTransform {
+  readonly oneDay = 86400000; // One day in millis
+  readonly oneHour = 3600000; // One hour in millis
+  readonly oneMin = 60000; // One minute in millis
+  readonly oneSecond = 1000; // One second in millis
+
+  transform(jsDate: Date, absoluteOnly: boolean): any {
```
max-vogler, line 13: Is it possible to change the return type from any to string?

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,54 @@
+import {Pipe, PipeTransform} from '@angular/core';
+import {DateTime} from 'luxon';
+
+@Pipe({
+  name: 'relativeTimestamp'
+})
+export class RelativeTimestampPipe implements PipeTransform {
+  readonly oneDay = 86400000; // One day in millis
```
max-vogler, line 8: These are de-facto constants. Please us `static` for constants - they are not changing per instance of RelativeTimestampPipe. 

Additionally, if class members (variables or methods) are not used outside of the class, prefer to make them private to prevent outside classes using/depending on them. 

Finally, you could move the information that the comment conveys ("in millis") into the variable name. This removes the need for the reader to read the comment, because the variable conveys all information:

```
  private static readonly ONE_DAY_IN_MILLIS = 86400000;
```

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const now = DateTime.local();
+    const timeElapsed = now.diff(date, "milliseconds").as("milliseconds");
+
+    if (timeElapsed < this.oneMin / 2) {
+      return "Just now";
+    }
+
+    if (timeElapsed < this.oneMin) {
+      return Math.floor(timeElapsed / this.oneSecond) + " seconds ago";
```
max-vogler, line 32: Use template strings instead of string concatenation:
```
return `${Math.floor(timeElapsed / this.oneSecond)} seconds ago`;
```

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    if (timeElapsed < this.oneHour) {
+      return Math.floor(timeElapsed / this.oneMin) + "min ago";
+    }
+
+    if (timeElapsed < this.oneDay) {
+      return Math.floor(timeElapsed / this.oneHour) + "h" +
+        Math.floor((timeElapsed % this.oneHour) / this.oneMin) + "min ago";
```
max-vogler, line 41: Luxon has logic to transform a time difference into units of choice. This is from their [docs](https://moment.github.io/luxon/docs/manual/math.html#duration-math):

```
var end = DateTime.fromISO('2017-03-13');
var start = DateTime.fromISO('2017-02-15');
end.diff(start, ['months', 'days']) //=> { months: 1, days: 2 }
```

I suggest using their diffing math over division and modulo calculations.

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    if (now.startOf("day").diff(date, "milliseconds").as("milliseconds") < this.oneDay) {
```
max-vogler, line 44: This is fairly complicated and could lead to displaying "yesterday at xx" when in fact the `date` was the day before yesterday in the user's local timezone. While relative times ("xx hours ago") can be calculated independently of the users local timezone, an absolute wall clock time like "yesterday at xx:xx" is in the user's local timezone. The only way the be sure if `date` is yesterday in the user's timezone, is to get the user's current datetime (in their timezone), then subtract 1 day and verify if the date part is equal.

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp.ng.html
```diff
@@ -0,0 +1,8 @@
+<div class="timestamp"
+  matTooltip="{{date | date:'yyyy-MM-dd HH:mm:ss':'utc'}} UTC"
+  matTooltipPosition="above"
+  matTooltipClass="tooltip"
+  matTooltipDisabled="{{!tooltipEnabled}}"
+  [style.border-bottom]="tooltipEnabled ? '1px dotted rgb(126, 126, 126)' : '0px'">
```
max-vogler, line 6: Use `none` instead of `'0px'` to hide the bottom border.

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp_test.ts
```diff
@@ -0,0 +1,210 @@
...
+    jasmine.clock().uninstall();
+  });
+
+  it('is created successfully', () => {
+    const fixture = TestBed.createComponent(Timestamp);
+    const componentInstance = fixture.componentInstance;
+    expect(componentInstance).toBeTruthy();
+  });
+
+  it('shows "Just now" for dates less than 1min ago', () => {
```
max-vogler, line 42: Please adjust test description to test behavior (30s ago).

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp_test.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  it('shows absolute timestamp for dates older than yesterday beginning of day', () => {
+    const fixture = TestBed.createComponent(Timestamp);
+    const componentInstance = fixture.componentInstance;
+
+    let date = new Date('2020-06-29T23:59:59.999');
+    componentInstance.date = date;
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText)
+      .toEqual(DateTime.fromJSDate(date).toLocaleString(DateTime.DATETIME_MED));
```
max-vogler, line 161: As a reader, I don't know what output to expect and need to read up on the luxon documentation. If possible, prefer to test against a string literal (`.toEqual('yesterday at 00:00')`) and not a function output (`toLocaleString(DateTime.DATETIME_MED)`).

grr/server/grr_response_server/gui/ui/components/client_search/module.ts
danielberbece: This was added before updating the formatter, that's why it has the spaces.

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
danielberbece: Sure!

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
danielberbece: Good idea!

grr/server/grr_response_server/gui/ui/components/timestamp/timestamp_test.ts
danielberbece: I was thinking that it may be different across localizations, but for test purposes I suppose it won't be a problem.

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
danielberbece: Didn't know about that, thanks!

grr/server/grr_response_server/gui/ui/components/timestamp/relative_timestamp_pipe.ts
danielberbece: This implementation uses local timezone of the user, because Luxon uses local timezone by default, so it shouldn't be a problem. Will change to make it more readable.

PR #796, opened 2020-07-07
Created a new dashboard for Fleetspeak

tsehori:
GRR deployments that incorporate [Fleetspeak](https://github.com/google/fleetspeak) can now use this dashboard along with the [existing dashboards](https://github.com/google/grr/tree/master/monitoring/grafana) for the different GRR server components. For quick import and use in [Grafana](https://grafana.com/), the dashboard is already generated in folder [dashboards_for_use/](https://github.com/google/grr/tree/master/monitoring/grafana/grr_grafanalib_dashboards/dashboards_for_use).
Just for context, this PR heavily relies on google/fleetspeak#258 that exposed many of the metrics used within this new dashboard.

Here's a snippet of the dashboard:
Screenshot from 2020-07-07 11-30-16

Some things to note regarding the dashboard:
- This dashboard belongs here and not in the Fleetspeak repository since nowadays Fleetspeak is almost entirely used within GRR. Hopefully this will be changed over time, but for now putting it here makes more sense.
- Unlike the GRR components' dashboards, the Fleetspeak dashboard does not use the general dashboards in [reusable_panels.py](https://github.com/google/grr/blob/master/monitoring/grafana/grr_grafanalib_dashboards/reusable_panels.py), thus some code is duplicated. This is intended by design so that the logic of this dashboard is not linked or dependent on GRR dashboard configurations, where if some of GRR's stats names change for example, it wouldn't break this new dashboard's code.
- The dashboard is independent of GRR server, that is; it can be imported and used without a running GRR deployment (just Fleetspeak).


monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+              timeRange=TimeRange("10s", "now"),
+              evaluator=LowerThan(ACTIVE_PROCESSES_ALERTING_CONDITION),
+              operator=OP_AND,
+              reducerType=RTYPE_SUM
+            )
+          ],
+        )
+      ),
+      Graph(
+        title="Sum of Process Memory Bytes (across all instances)",
```
mbushkov, line 35: I'd stick to a single convention regarding where we indicate the units used in the graph. I.e. here this is indicated in the graph's title, whereas in "CPU Usage" graph below it's indicated in the legend.
Do I understand it correctly, that we can also explicitly specify the units used in the Y axis? Then we won't need the units description to be present in the title or legend.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+            legendFormat="Resident Memory",
+          ),
+        ]
+      ),
+      Graph(
+        title="CPU Usage",
+        targets=[
+          Target(
+            expr='avg(rate(process_cpu_seconds_total{job="fleetspeak"}[30s])) * 100',
+            legendFormat="Average Process CPU Usage in %",
```
mbushkov, line 48: I wonder if we can tell Grafana explicitly that we're dealing with percents on Y axis? I.e. not to put it into the legend, but make Grafana aware of the unit (IIRC that was possible to do through Grafana UI).

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+        ],
+        yAxes=YAxes(
+          left=YAxis(max=105)
+        ),
+      ),
+    ]
+    ),
+    Row(panels=[
+      Graph(
+        title="Datastore Latency per Operation",
```
mbushkov, line 59: It's not clear whether latency is expressed in seconds or in something else. Let's indicate the unit of the Y axis somehow (preferably - not by mentioning it in title/legend, but by explicitly configuring the Y axis presentation ).

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+        targets=[
+          Target(
+            expr='sum by (operation) (rate(fleetspeak_server_datastore_operations_completed_latency_sum[10m]) / rate(fleetspeak_server_datastore_operations_completed_latency_count[10m]))',
+            legendFormat="{{operation}}",
+          ),
+        ]
+      ),
+      Heatmap(
+        title="Datastore Latency Distribution",
```
mbushkov, line 68: Same here - it should be clear, which unit is used in the graph.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+        title="Client Polls Errors Rate per Status",
+        targets=[
+          Target(
+            expr='sum by (http_status_code) (rate(fleetspeak_server_client_polls_total{http_status_code="^(4|5).*"}[10m]))',
+            legendFormat="{{http_status_code}}",
+          ),
+        ]
+      ),
+      Heatmap(
+        title="Client Poll Latency Distribution",
```
mbushkov, line 133: Same thing - not clear which unit is used for latency.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+            expr='sum(rate(fleetspeak_server_client_polls_operation_time_latency_bucket[10m])) by (le)',
+          ),
+        ],
+        legend={'show': True},
+      ),
+      ]
+    ),
+    Row(panels=[
+      Graph(
+        title="Client Poll Latency Quantiles",
```
mbushkov, line 145: Same question: unit is not clear.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+            legendFormat="90th percentile",
+          ),
+          Target(
+            expr='histogram_quantile(0.99, sum by (le) (rate(fleetspeak_server_client_polls_operation_time_latency_bucket[10m])))',
+            legendFormat="99th percentile",
+          ),
+        ]
+      ),
+      Graph(
+        title="Client Poll Operation Time Latency per Poll Type",
```
mbushkov, line 162: Ditto.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+        targets=[
+          Target(
+            expr='sum by (poll_type) (rate(fleetspeak_server_client_polls_operation_time_latency_sum[10m]) / rate(fleetspeak_server_client_polls_operation_time_latency_count[10m]))',
+            legendFormat="{{poll_type}}",
+          ),
+        ]
+      ),
+      Graph(
+        title="Client Polls Rate per Cache",
```
mbushkov, line 171: I'd phrase it more explicitly: Cache -> Cache Status

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+            expr='sum by (destination_service) (rate(fleetspeak_messages_ingested_total[10m]))',
+            legendFormat="{{destination_service}}",
+          ),
+        ]
+      ),
+      ]
+    ),
+    Row(panels=[
+      Heatmap(
+        title="Message Processing Latency Distribution",
```
mbushkov, line 192: Same thing: should be clear, which unit is used.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
```diff
@@ -0,0 +1,239 @@
...
+        targets=[
+          Target(
+            expr='sum(rate(fleetspeak_server_messages_processed_latency_bucket[10m])) by (le)',
+          ),
+        ],
+        legend={'show': True},
+      ),
+      Graph(
+        title="Message Processing Latency Quantiles per Service",
```
mbushkov, line 201: Same thing: should be clear, which unit is used.

monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py
tsehori: Yes, Grafana allows editing the Y-axis legend to reflect measurement units. Will work on that.

tsehori:
All comments are resolved.
On another note; the heatmaps show magnitudes of data which are very different from the graphs; I think it is due to us summing up the latencies in the graphs. I think averaging the data rather than summing it would be preferable; what do you think?

mbushkov:
> On another note; the heatmaps show magnitudes of data which are very different from the graphs; I think it is due to us summing up the latencies in the graphs. I think averaging the data rather than summing it would be preferable; what do you think?

Yes, if this will make it easier to correlate heatmaps with other graphs, let's use averages instead of sums.

tsehori:
I added the bytes where applicable, as discussed.
I tried averaging, but it doesn't look like it helps correlating with the other graphs.
By the way, what should we do about the rest of the graphs with no Y-axis units? Leave them as is for the time being or figure them out?

mbushkov:
> I tried averaging, but it doesn't look like it helps correlating with the other graphs.

We can merge it as is and then figure out the histograms graphs behavior.

> By the way, what should we do about the rest of the graphs with no Y-axis units? Leave them as is for the time being or figure them out?

Yes, we should specify Y-axis units everywhere where we deal with something other than counters. Saying this, let's do this in a separate PR.

PR #798, opened 2020-07-07
Add format argument for hunt timelines collection

alexandrucosminmihai:
- Update the ApiGetCollectedHuntTimelinesArgs protobuf message to hold the format
- Update the API router and handler
- Update the Python API client to accept the format argument
- Update/Add tests

### DIFF
--- a/grr/server/grr_response_server/gui/api_call_router.py
+++ b/grr/server/grr_response_server/gui/api_call_router.py
@@ -948,13 +948,13 @@ class ApiCallRouterStub(ApiCallRouter):
   @Category("Hunts")
   @ArgsType(api_timeline.ApiGetCollectedHuntTimelinesArgs)
   @ResultBinaryStream()
-  @Http("GET", "/api/hunts/<hunt_id>/timelines")
+  @Http("GET", "/api/hunts/<hunt_id>/timelines/<format>")
   def GetCollectedHuntTimelines(
       self,
       args: api_timeline.ApiGetCollectedHuntTimelinesArgs,
       token: Optional[access_control.ACLToken] = None,
   ) -> api_timeline.ApiGetCollectedHuntTimelinesHandler:
-    """Exports results of a timeline hunt as a ZIP archive."""
+    """Exports results of a timeline hunt using the format as a ZIP archive."""
     raise NotImplementedError()
 
   # Stats metrics methods.
--- a/grr/server/grr_response_server/gui/api_call_router.py
+++ b/grr/server/grr_response_server/gui/api_call_router.py
@@ -954,7 +954,11 @@ class ApiCallRouterStub(ApiCallRouter):
       args: api_timeline.ApiGetCollectedHuntTimelinesArgs,
       token: Optional[access_control.ACLToken] = None,
   ) -> api_timeline.ApiGetCollectedHuntTimelinesHandler:
-    """Exports results of a timeline hunt using the format as a ZIP archive."""
+    """Exports results of a timeline hunt.
+
+    The results are exported as a ZIP archive whose files follow the specified
+    format.
+    """
     raise NotImplementedError()
 
   # Stats metrics methods.
--- a/grr/server/grr_response_server/gui/api_integration_tests/hunt_test.py
+++ b/grr/server/grr_response_server/gui/api_integration_tests/hunt_test.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 
 import csv
 import io
+import stat
 import zipfile
 
 from absl import app
@@ -331,7 +332,7 @@ class ApiClientLibHuntTest(
 
         self.assertEqual(rows[0][1], "/bar/baz/quux")
         self.assertEqual(rows[0][2], "5926273453")
-        self.assertEqual(rows[0][3], "?rw-rw-r--")
+        self.assertEqual(rows[0][3], stat.filemode(0o664))
         self.assertEqual(rows[0][6], "13373")
         self.assertEqual(rows[0][7], "111")
         self.assertEqual(rows[0][8], "222")
@@ -339,7 +340,7 @@ class ApiClientLibHuntTest(
 
         self.assertEqual(rows[1][1], "/bar/baz/quuz")
         self.assertEqual(rows[1][2], "6037384564")
-        self.assertEqual(rows[1][3], "?rwxrwxrwx")
+        self.assertEqual(rows[1][3], stat.filemode(0o777))
         self.assertEqual(rows[1][6], "13374")
         self.assertEqual(rows[1][7], "777")
         self.assertEqual(rows[1][8], "888")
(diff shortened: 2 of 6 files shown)


grr/server/grr_response_server/gui/api_plugins/timeline.py
```diff
@@ -112,17 +112,22 @@ def Handle(
...
     filename = f"timelines_{hunt_id}.zip"
-    content = self._Generate(hunt_id)
+    content = self._Generate(hunt_id, args.format)
     return api_call_handler_base.ApiBinaryStream(filename, content)
 
-  def _Generate(self, hunt_id: Text) -> Iterator[bytes]:
+  def _Generate(
+      self,
+      hunt_id: Text,
+      format: ApiGetCollectedTimelineArgs.Format
```
panhania, line 121: Try to use trailing commas in such cases—I am pretty sure that our internall formatter would move the paren to this line without it. Also, it makes the diffs nicer if you add more arguments.

grr/server/grr_response_server/gui/api_plugins/timeline.py
```diff
@@ -152,15 +158,22 @@ def _GenerateTimeline(
...
     args = ApiGetCollectedTimelineArgs()
     args.client_id = client_id
     args.flow_id = flow_id
-    # TODO(hanuszczak): Add support for other formats.
-    args.format = ApiGetCollectedTimelineArgs.Format.RAW_GZCHUNKED  # pytype: disable=attribute-error
+    args.format = format
+
+    if format == ApiGetCollectedTimelineArgs.Format.RAW_GZCHUNKED:
+      filename = f"{client_id}_{fqdn}.gzchunked"
+    elif format == ApiGetCollectedTimelineArgs.Format.BODY:  # pytype: disable=attribute-error
```
panhania, line 171: Interesting, why `pytype` directive? We don't have any `pytype` in our open source tree, do we? Also, why is it needed here but not two lines above?

grr/server/grr_response_server/gui/api_plugins/timeline_test.py
```diff
@@ -206,6 +206,83 @@ def testRaisesOnIncorrectFlowType(self):
     with self.assertRaises(ValueError):
       self.handler.Handle(args)
 
+  def testBodyMultipleClients(self):
```
panhania, line 209: Considering how similar is this test to the previous one, is there a reason why you omitted the mode?

grr/server/grr_response_server/gui/api_plugins/timeline_test.py
```diff
@@ -206,6 +206,83 @@ def testRaisesOnIncorrectFlowType(self):
     with self.assertRaises(ValueError):
       self.handler.Handle(args)
```
panhania, line 208: There should also be a test case with incorrect format.

grr/server/grr_response_server/gui/api_call_router.py
```diff
@@ -948,13 +948,13 @@ def GetHuntFile(self, args, token=None):
...
   @ResultBinaryStream()
-  @Http("GET", "/api/hunts/<hunt_id>/timelines")
+  @Http("GET", "/api/hunts/<hunt_id>/timelines/<format>")
   def GetCollectedHuntTimelines(
       self,
       args: api_timeline.ApiGetCollectedHuntTimelinesArgs,
       token: Optional[access_control.ACLToken] = None,
   ) -> api_timeline.ApiGetCollectedHuntTimelinesHandler:
-    """Exports results of a timeline hunt as a ZIP archive."""
+    """Exports results of a timeline hunt using the format as a ZIP archive."""
```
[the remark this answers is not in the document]
alexandrucosminmihai: I'll fix it. Thank you!

grr/server/grr_response_server/gui/api_integration_tests/hunt_test.py
```diff
@@ -255,7 +258,94 @@ def testExportedResults(self):
...
+    with zipfile.ZipFile(buffer, mode="r") as archive:
+      with archive.open(f"{client_id}_{fqdn}.body", mode="r") as file:
+        content_file = file.read().decode("utf-8")
+
+        rows = list(csv.reader(io.StringIO(content_file), delimiter="|"))
+        self.assertLen(rows, 2)
+
+        self.assertEqual(rows[0][1], "/bar/baz/quux")
+        self.assertEqual(rows[0][2], "5926273453")
+        self.assertEqual(rows[0][3], "?rw-rw-r--")
```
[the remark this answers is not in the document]
alexandrucosminmihai: Oh, yeah, it's much more readable using filemode. Thank you!

grr/server/grr_response_server/gui/api_plugins/timeline.py
alexandrucosminmihai: Makes sense. Thank you!

grr/server/grr_response_server/gui/api_plugins/timeline.py
alexandrucosminmihai: Oups! I reused this line as it was from the previous version of the file, but forgot about the comment. Thanks!

grr/server/grr_response_server/gui/api_plugins/timeline_test.py
alexandrucosminmihai: I've tried to be as less intrusive as possible to the existing codebase. So because initially there wasn't a filemode check in this test, I omitted it as well. I will add it. Thank you!

grr/server/grr_response_server/gui/api_plugins/timeline_test.py
alexandrucosminmihai: I'll add it. Thank you!

grr/server/grr_response_server/gui/api_plugins/timeline.py
```diff
@@ -168,10 +173,11 @@ def _GenerateTimeline(
 
     if format == ApiGetCollectedTimelineArgs.Format.RAW_GZCHUNKED:
       filename = f"{client_id}_{fqdn}.gzchunked"
-    elif format == ApiGetCollectedTimelineArgs.Format.BODY:  # pytype: disable=attribute-error
+    elif format == ApiGetCollectedTimelineArgs.Format.BODY:
       filename = f"{client_id}_{fqdn}.body"
     else:
-      message = "Incorrect timeline export format: {}".format(format)
+      message = "Unknown file extension for timeline export format: {}".format(
+          format)
```
panhania, line 180: Nit: if you use f-strings, you can avoid an ugly line break here.

alexandrucosminmihai: Fixed it. Thank you!

panhania: Hm, actually, now thinking about it: the text is also not correct. It is not the file extension that is unknown, it is the format that has been specified (the format implies the file extension, if known, but the API user has no control over the extension itself).

alexandrucosminmihai: You are right, but the format is already checked in the top-level [Handle method](https://github.com/alexandrucosminmihai/grr/blob/e45d4e12ef92682243f68c03285826d4c30dba78/grr/server/grr_response_server/gui/api_plugins/timeline.py#L114) and I thought that it would be useful for the developer adding new export formats to be notified in this "private" method if they forgot to specify the file extension to be used by the new format. Please let me know if you want me to change it.

grr/server/grr_response_server/gui/api_plugins/timeline.py
```diff
@@ -111,18 +111,28 @@ def Handle(
       message = f"Hunt '{hunt_id}' is not a timeline hunt"
       raise ValueError(message)
 
+    if (args.format != ApiGetCollectedTimelineArgs.Format.RAW_GZCHUNKED) and (
+        args.format != ApiGetCollectedTimelineArgs.Format.BODY):
```
panhania, line 115: No need for the internal parens, `!=` binds stronger than `and`. External ones are needed because of the linebreak, though.

grr/server/grr_response_server/gui/api_plugins/timeline.py
panhania: Oh, you are right. In this case, I would not even write any exception message, just raise `AssertionError` without anything (as this is just an equivalent of "the impossible happened").

grr/server/grr_response_server/gui/api_plugins/timeline.py
alexandrucosminmihai: Thank you! I updated it

grr/server/grr_response_server/gui/api_plugins/timeline.py
alexandrucosminmihai: Changed it. Thank you!

PR #802, opened 2020-07-08
Add status chip in client overview for UIv2

### LINKED ISSUE #797: No status indicator in UIv2
In the new UI, we need a way to show the status of a client. The current UI has this option, but not the new one.

To achieve this, a status chip should be added in the client overview, which will showcase whether the client is `Online` or `Offline`. The chip should be suggestively colored, depending on the state of the client.

danielberbece:
This pull-request adds a status chip to the client overview panel, which reflects the status of the client.

The status is computed based on the timestamp the client was last seen active. If he was active until 15 minutes ago, the status chip will be green, with an "Online" text. Otherwise, the status chip will be gray, with an "Offline" text.

The status chip is a standalone component in the client module, and it updates every time the timestamp is updated and every one second. This means that the status chip will be updated in both directions (from "Offline" to "Online" and vice versa) immediately, in a reactive way.

This pull-request closes #797


grr/server/grr_response_server/gui/ui/components/client/client.ng.html
```diff
@@ -6,6 +6,7 @@ <h1 class="mat-h1">{{client.knowledgeBase.fqdn}}</h1>
       </div>
       <div class="chips">
         <mat-chip-list>
+          <status-chip [lastSeen]="client.lastSeenAt"></status-chip>
```
max-vogler, line 9: We also want to use this component outside of <Client />. For unambiguity, please rename the component to  `online-chip`, `client-online-chip`, or something of your choice.

grr/server/grr_response_server/gui/ui/components/client/module.ts
```diff
@@ -27,6 +28,7 @@ import {TimestampModule} from '../timestamp/module';
   ],
   declarations: [
     Client,
+    StatusChip,
```
max-vogler, line 31: Please move StatusChip with all its files in a dedicated directory, e.g. `[...]/ui/components/<component-name>/` and create a dedicated Module only for StatusChip.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ng.html
```diff
@@ -0,0 +1,6 @@
+<div *ngIf="(status$ | async) === 'online'; else offlineChip">
```
max-vogler, line 1: Is the wrapping `<div>` needed? When online, this Component would render `<div><mat-chip>Online</mat-chip></div>`, but when offline it would only render `<mat-chip>Offline</mat-chip>`.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
```diff
@@ -0,0 +1,44 @@
...
+  templateUrl: './status_chip.ng.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class StatusChip implements OnChanges {
+  private static readonly STATUS_OFFLINE = 'offline';
+  private static readonly STATUS_ONLINE = 'online';
+  private static readonly ONLINE_TRESHOLD_MINUTES = 15;
+
+  @Input() lastSeen?: Date;
+  private lastSeenChange$ = new Subject<void>();
```
max-vogler, line 20: Please add `readonly` because the variable reference is never changed.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // status observable that updates every second and when lastSeen changes
+  status$ = merge(interval(1000), this.lastSeenChange$).pipe(
+    map(_ => this.getStatus())
+  );
+
+  ngOnChanges(_changes: SimpleChanges): void {
+    this.lastSeenChange$.next()
```
max-vogler, line 28: Missing semicolon.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
```diff
@@ -0,0 +1,44 @@
...
+  private static readonly STATUS_OFFLINE = 'offline';
+  private static readonly STATUS_ONLINE = 'online';
+  private static readonly ONLINE_TRESHOLD_MINUTES = 15;
+
+  @Input() lastSeen?: Date;
+  private lastSeenChange$ = new Subject<void>();
+
+  // status observable that updates every second and when lastSeen changes
+  status$ = merge(interval(1000), this.lastSeenChange$).pipe(
+    map(_ => this.getStatus())
```
max-vogler, line 24: `_` is unused. Instead, define a lambda function without arguments: `() => this.getStatus()`.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  ngOnChanges(_changes: SimpleChanges): void {
```
max-vogler, line 27: Use `changes` over `_changes`. Avoid leading underscores.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.lastSeenChange$.next()
+  }
+
+  getStatus(): string {
+    if (this.lastSeen === undefined) {
+      return StatusChip.STATUS_OFFLINE;
+    }
+
+    const timeDiff = DateTime.local()
```
max-vogler, line 36: This code contains some [accidental complexity](https://en.wikipedia.org/wiki/No_Silver_Bullet). As a reader I wonder what `DateTime.local()` means (is it the same as now?). Also, why the need to specify `'minutes'` twice?

I suggest to use a [Duration](https://moment.github.io/luxon/docs/class/src/duration.js~Duration.html) object as `ONLINE_THRESHOLD` instead of an integer as `ONLINE_THRESHOLD_MINUTES`. This way, the unit is encoded in the threshold itself and not only in its variable name. Also, you can use `this.lastSeen.diffNow()` to calculate the elapsed time.

grr/server/grr_response_server/gui/ui/components/client/status_chip_test.ts
```diff
@@ -0,0 +1,120 @@
...
+    jasmine.clock().uninstall();
+  });
+
+  it('is created successfully', () => {
+    const fixture = TestBed.createComponent(TestHostComponent);
+    const componentInstance = fixture.componentInstance;
+    expect(componentInstance).toBeTruthy();
+  });
+
+  it('shows "Online" for users last seen up to 15 minutes ago', () => {
```
max-vogler, line 53: Replace `users` with `clients`. Same in the other test descriptions.

grr/server/grr_response_server/gui/ui/components/client/status_chip_test.ts
```diff
@@ -0,0 +1,120 @@
...
+    componentInstance.lastSeen = new Date('2020-07-01T12:40:15');
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText).toEqual(STATUS_OFFLINE);
+
+    componentInstance.lastSeen = new Date('1623-07-01T12:59:00.001');
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText).toEqual(STATUS_OFFLINE);
+  });
+
+  it('updates the status from "Online" to "Offline" as time passes by', () => {
```
max-vogler, line 93: `as time passes by` is ambiguous. Please use: when time crosses the threshold.

grr/server/grr_response_server/gui/ui/components/client/client.ng.html
danielberbece: Yes, no problem

grr/server/grr_response_server/gui/ui/components/client/module.ts
danielberbece: Alright!

grr/server/grr_response_server/gui/ui/components/client/status_chip.ts
danielberbece: That is a great idea! The second `'minutes'` was used because luxon has the ability to convert a `Duration` to the unit of choosing. Indeed, the first one was not actually necessary as I thought.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ng.html
danielberbece: This one is a little tricky, because if I try and remove the div, to integrate `ngIf` directive into the `<mat-chip>`, for some reason it won't work. Trying to enclose the offline chip in a `<div>` also doesn't work. Only other way that will work is to have a `<mat-chip-list>` enclosing the whole thing. What do you think?

grr/server/grr_response_server/gui/ui/components/online_chip/online_chip.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+@Component({
+  selector: 'online-chip',
+  templateUrl: './online_chip.ng.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OnlineChip implements OnChanges {
+  private static readonly STATUS_OFFLINE = 'offline';
+  private static readonly STATUS_ONLINE = 'online';
+  private static readonly ONLINE_TRESHOLD = Duration.fromMillis(1000 * 60 * 15); // 15 minutes
```
max-vogler, line 17: Use `Duration.fromObject({minutes: 15});` instead - then no math and comment is needed.

danielberbece: That's a lot cleaner, thanks!

grr/server/grr_response_server/gui/ui/components/online_chip/online_chip.ng.html
```diff
@@ -0,0 +1,6 @@
+<div *ngIf="(status$ | async) === 'online'; else offlineChip">
+  <mat-chip color="accent" selected>Online</mat-chip>
+</div>
+<ng-template #offlineChip>
+  <mat-chip>Offline</mat-chip>
```
max-vogler, line 5: Please also wrap this mat-chip in a div to produce the same element hierarchy in the if and else branch. Because div is `display: block` per default, the display could change when online status changes.

grr/server/grr_response_server/gui/ui/components/client/status_chip.ng.html
max-vogler: Please also wrap this mat-chip in a div to produce the same element hierarchy in the if and else branch. Because div is display: block per default, the display could change when online status changes.

max-vogler:
Awesome!

PR #803, opened 2020-07-09
Update timestamp in UIv2 to show absolute timestamp for short format

danielberbece:
Added some changes to the timestamp component. Now, it shows the absolute format as a base format, and the tooltip contains a relative timestamp, showing how much time passed since then. Also, I added a long version, which doesn't show the relative timestamp in a tooltip, but inline instead. The long version also has a copy icon, for easy and fast copy.


danielberbece:
This is how the short and long formats look like: 
image
image

danielberbece:
This is how the short format looks like:
image
And this is how the long format looks like:
image

max-vogler:
Looks good!

PR #806, opened 2020-07-14
Add label to client

danielberbece:
Added a button to use for adding a label to a client

### DIFF
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -1,11 +1,12 @@
-<h1 mat-dialog-title>Adding a label</h1>
-<div mat-dialog-content>
+<div mat-dialog-title class="title">Adding a label</div>
+<div mat-dialog-content class="content">
   <p>You can add a new label or select from existing labels.</p>
-  <mat-form-field>
-    <input matInput [(ngModel)]="label">
+  <mat-form-field class="label-field" appearance="outline">
+    <mat-label>Start typing label name</mat-label>
+    <input matInput [(ngModel)]="label" cdkFocusInitial (keyup.enter)="onAddClick()">
   </mat-form-field>
 </div>
-<div mat-dialog-actions>
-  <button mat-button (click)="onCancelClick()">Cancel</button>
-  <button mat-button [mat-dialog-close]="label" cdkFocusInitial>Add label</button>
+<div mat-dialog-actions class="actions">
+  <button mat-button color="primary" (click)="onCancelClick()">Cancel</button>
+  <button mat-button color="primary" (click)="onAddClick()">Add label</button>
 </div>
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -1,0 +1,11 @@
+<h1 mat-dialog-title>Adding a label</h1>
+<div mat-dialog-content>
+  <p>You can add a new label or select from existing labels.</p>
+  <mat-form-field>
+    <input matInput [(ngModel)]="label">
+  </mat-form-field>
+</div>
+<div mat-dialog-actions>
+  <button mat-button (click)="onCancelClick()">Cancel</button>
+  <button mat-button [mat-dialog-close]="label" cdkFocusInitial>Add label</button>
+</div>
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -6,21 +6,28 @@
     <input matInput
       cdkFocusInitial
       (keyup.enter)="onAddClick()"
-      [formControl]="newLabelInputControl"
+      [formControl]="labelInputControl"
       [matAutocomplete]="auto">
 
     <mat-autocomplete #auto="matAutocomplete">
-      <mat-option *ngFor="let option of filteredLabels | async" [value]="option">
-        {{option}}
-        <ng-template [ngIf]="clientHasLabel(option)">already present</ng-template>
+      <mat-option *ngFor="let label of suggestedLabels$ | async" [value]="label">
+        <small>{{label}}</small>
       </mat-option>
-    </mat-autocomplete>
+      <!-- If the input label already exists, announce to the user! -->
+      <mat-option *ngIf="labelInputControl.hasError('alreadyPresentLabel')" [value]="labelInputControl.value">
+        <small>{{labelInputControl.value}} already present!</small>
+      </mat-option>
+      <!-- Suggest creating a new label, if so (default option) -->
+      <mat-option *ngIf="isNewLabel$ | async" [value]="labelInputControl.value">
+        <small>Add new label "{{labelInputControl.value.trim()}}"</small>
+      </mat-option>
+     </mat-autocomplete>
 
-    <mat-error *ngIf="newLabelInputControl.hasError('alreadyPresentLabel')">This label is already present!</mat-error>
+    <mat-error *ngIf="labelInputControl.hasError('alreadyPresentLabel')">This label is already present!</mat-error>
   </mat-form-field>
 </div>
 <div mat-dialog-actions class="actions">
   <button id="cancel" mat-button color="primary" (click)="onCancelClick()">Cancel</button>
   <button id="add" mat-button color="primary" (click)="onAddClick()"
-    [disabled]="newLabelInputControl.invalid">Add label</button>
+    [disabled]="labelInputControl.invalid">Add label</button>
 </div>
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -3,7 +3,19 @@
   <p>You can add a new label or select from existing labels.</p>
   <mat-form-field class="label-field" appearance="outline">
     <mat-label>Start typing label name</mat-label>
-    <input matInput [(ngModel)]="label" cdkFocusInitial (keyup.enter)="onAddClick()">
+    <input matInput
+      cdkFocusInitial
+      (keyup.enter)="onAddClick()"
+      [formControl]="newLabelInputControl"
+      [matAutocomplete]="auto">
+
+    <mat-autocomplete #auto="matAutocomplete">
+      <mat-option *ngFor="let option of filteredCurrentLabels | async" [value]="option">
+        {{option}}
+      </mat-option>
+    </mat-autocomplete>
+
+    <mat-error *ngIf="newLabelInputControl.invalid">This label is already present!</mat-error>
   </mat-form-field>
 </div>
 <div mat-dialog-actions class="actions">
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -10,8 +10,9 @@
       [matAutocomplete]="auto">
 
     <mat-autocomplete #auto="matAutocomplete">
-      <mat-option *ngFor="let option of filteredCurrentLabels | async" [value]="option">
+      <mat-option *ngFor="let option of filteredLabels | async" [value]="option">
         {{option}}
+        <ng-template [ngIf]="clientHasLabel(option)">already present</ng-template>
       </mat-option>
     </mat-autocomplete>
 
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -15,10 +15,10 @@
       </mat-option>
       <!-- If the input label already exists, announce to the user! -->
       <mat-option *ngIf="labelInputControl.hasError('alreadyPresentLabel')" [value]="labelInputControl.value">
-        <small>{{labelInputControl.value}} already present!</small>
+        <small>Label "{{labelInputControl.value.trim()}}" already present!</small>
       </mat-option>
       <!-- Suggest creating a new label, if so (default option) -->
-      <mat-option *ngIf="isNewLabel$ | async" [value]="labelInputControl.value">
+      <mat-option *ngIf="isNewLabel$ | async" [value]="labelInputControl.value.trim()">
         <small>Add new label "{{labelInputControl.value.trim()}}"</small>
       </mat-option>
      </mat-autocomplete>
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -15,10 +15,11 @@
       </mat-option>
     </mat-autocomplete>
 
-    <mat-error *ngIf="newLabelInputControl.invalid">This label is already present!</mat-error>
+    <mat-error *ngIf="newLabelInputControl.hasError('alreadyPresentLabel')">This label is already present!</mat-error>
   </mat-form-field>
 </div>
 <div mat-dialog-actions class="actions">
   <button id="cancel" mat-button color="primary" (click)="onCancelClick()">Cancel</button>
-  <button id="add" mat-button color="primary" (click)="onAddClick()">Add label</button>
+  <button id="add" mat-button color="primary" (click)="onAddClick()"
+    [disabled]="newLabelInputControl.invalid">Add label</button>
 </div>
--- a/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
+++ b/grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
@@ -19,6 +19,6 @@
   </mat-form-field>
 </div>
 <div mat-dialog-actions class="actions">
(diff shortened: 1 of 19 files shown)


grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -380,25 +380,34 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
...
+      tap(flows => {
+        this.updateFlows(flows);
+      }),
+    ));
+
+  // An effect to add a label to the selected client
+  readonly addClientLabel = this.effect<string>(
+    obs$ => obs$.pipe(
+      withLatestFrom(this.selectedClientId$),
+      map(([label, clientId]) => {
```
max-vogler, line 407: In your code, `map` returns an Observable that would send the HTTP POST request, but this inner observable is not subscribed to. Try `switchMap` instead of `map`. Further reading: https://medium.com/@luukgruijs/understanding-rxjs-map-mergemap-switchmap-and-concatmap-833fc1fb09ff

danielberbece: Thanks, that was the problem!

danielberbece:
I updated the pull-request. I implemented an `Add label` button, which upon click shows a dialog where the user can enter the new label. When a label is added, the client page is updated and the new label appears. 
The tests for this feature are still a work in progress.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
```diff
@@ -0,0 +1,24 @@
...
+      [formControl]="newLabelInputControl"
+      [matAutocomplete]="auto">
+
+    <mat-autocomplete #auto="matAutocomplete">
+      <mat-option *ngFor="let option of filteredCurrentLabels | async" [value]="option">
+        {{option}}
+      </mat-option>
+    </mat-autocomplete>
+
+    <mat-error *ngIf="newLabelInputControl.invalid">This label is already present!</mat-error>
```
max-vogler, line 18: Use `hasError('alreadyPresentLabel')` instead of `invalid` to only show this error message for the validator that should trigger it.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
```diff
@@ -0,0 +1,24 @@
...
+  </mat-form-field>
+</div>
+<div mat-dialog-actions class="actions">
+  <button mat-button color="primary" (click)="onCancelClick()">Cancel</button>
+  <button mat-button color="primary" (click)="onAddClick()">Add label</button>
```
max-vogler, line 23: Since the onclick-handler does nothing when the form is invalid, this button should be disabled when the form is disabled. Since adding a label can be regarded as an idempotent action (the collection of client labels has the same state, even if you add a label twice), I'd also be fine with not handling this case at all and allowing the user to try to add a label twice.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
```diff
@@ -0,0 +1,24 @@
...
+  <mat-form-field class="label-field" appearance="outline">
+    <mat-label>Start typing label name</mat-label>
+    <input matInput
+      cdkFocusInitial
+      (keyup.enter)="onAddClick()"
+      [formControl]="newLabelInputControl"
+      [matAutocomplete]="auto">
+
+    <mat-autocomplete #auto="matAutocomplete">
+      <mat-option *ngFor="let option of filteredCurrentLabels | async" [value]="option">
```
max-vogler, line 13: Is this autocomplete only suggesting already pre-existing labels of the client?

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
```diff
@@ -0,0 +1,55 @@
...
+      .map(clientLabel => clientLabel.name)
+      .filter(label => label.toLowerCase().includes(filterValue));
+  }
+
+  labelPresentValidator(): ValidatorFn {
+    return (control: AbstractControl): {[key: string]: any} | null => {
+      const input = control.value.toLowerCase();
+      const matchingLabels = this.clientLabels
+        .map(clientLabel => clientLabel.name)
+        .filter(label => label.toLowerCase() === input);
```
max-vogler, line 37: Use `find()` over() `filter` to  reduce the average runtime.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
danielberbece: Good idea!

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
danielberbece: Exactly!

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
danielberbece: Good idea!

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
max-vogler: We want to suggest all labels in the autocomplete, not only labels that are already assigned to the client.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
danielberbece: I think that disabling the button would be a nice touch!

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
danielberbece: All labels from all clients? Is there an API for this?

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -36,6 +36,14 @@ export declare interface ApiClientLabel {
   readonly name?: string;
 }
 
+/**
+ * AddClientLabel proto mapping.
+ */
+export declare interface ApiAddClientLabel {
```
max-vogler, line 42: Please rename to the correct `ApiAddClientsLabelsArgs`.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ng.html
max-vogler: Yes, use `ApiListClientsLabelsHandler` via `GET /api/v2/clients/labels`.

To clarify: Clients and Labels have a many-to-many relationship. Any Client can have `[0,n]` labels assigned and any label be assigned to many clients. As an implementation detail, if a label is no longer assigned to any client, it is also deleted from the database, but that does not need to concern you here.

When adding a label to a Client, the user can type in an arbitrary string to be used as label. The autocomplete should suggest any pre-existing labels, not only labels that are already assigned to the current client.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
```diff
@@ -0,0 +1,97 @@
...
+
+@Component({
+  selector: 'client-add-label-dialog',
+  templateUrl: './client_add_label_dialog.ng.html',
+  styleUrls: ['./client_add_label_dialog.scss'],
+})
+export class ClientAddLabelDialog {
+  constructor(
+    readonly dialogRef: MatDialogRef<ClientAddLabelDialog>,
+    @Inject(MAT_DIALOG_DATA) private clientLabels: ReadonlyArray<ClientLabel>,
```
max-vogler, line 17: Is it possible to change both to `private readonly`?

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  /** An internal observable, which emits the following combined:
+   *    * a string containing the trimmed input
+   *    * a string[] containing all labels used in GRR
+   */
+  private readonly inputAndAllLabels$ = this.labelInputControl.valueChanges
+    .pipe(
+      filter((input): input is string | String => input !== undefined),
+      map(input => input.trim()),
+      combineLatest(this.allClientsLabels$.pipe(
```
max-vogler, line 32: I don't think combineLatest works as you intended here. The [docs](https://www.learnrxjs.io/learn-rxjs/operators/combination/combinelatest) do not show it used in a `pipe`. Instead, I think you want to use `withLatestFrom`.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
```diff
@@ -0,0 +1,97 @@
...
+    );
+
+  /** An observable emitting true when the input label is not exactly matching
+   *  any of the existing labels in GRR, and false otherwise.
+   */
+  readonly isNewLabel$: Observable<boolean> = this.inputAndAllLabels$
+    .pipe(
+      map(([input, allLabels]) => {
+        if (input === '') return false;
+        return allLabels.find(label => input === label) === undefined;
```
max-vogler, line 44: Use `allLabels.includes()`, which better expresses what you try to achieve.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
```diff
@@ -0,0 +1,97 @@
...
+  /** An observable emitting an array of possible labels, for autocompletion.
+   *  The array doesn't contain labels already assigned to the client.
+   */
+  readonly suggestedLabels$: Observable<string[]> = this.inputAndAllLabels$
+    .pipe(
+      map(([input, allLabels]) => {
+        if (input === '') return [];
+        return allLabels.filter(label => label.includes(input));
+      }),
+      map(filteredLabels => filteredLabels.filter((label) => this.clientHasLabel(label) === false))
```
max-vogler, line 57: Use `!` instead of `=== false`.

grr/server/grr_response_server/gui/ui/components/client_add_label_dialog/client_add_label_dialog.ts
danielberbece: Totally

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -36,6 +36,17 @@ export declare interface ApiClientLabel {
   readonly name?: string;
 }
 
+export declare interface ApiClientsLabels {
```
max-vogler, line 39: Please use the exact names from the Protobuf files (`ApiListClientsLabelsResult` see proto/grr_response_proto/api/client.proto). This allows us to switch from the manual definitions to auto-generated ones in the near future.

max-vogler: Also, please add a one-line JSDoc. I know it doesn't really contain information, but it is a necessary evil to keep the linter happy until we have code generation in place.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -36,6 +36,17 @@ export declare interface ApiClientLabel {
...
+  readonly items?: ApiClientLabel[];
+}
+/**
+ * AddClientsLabelsArgs proto mapping.
+ */
+export declare interface ApiAddClientsLabelsArgs {
+  readonly client_ids: string[];
```
max-vogler, line 46: The API expects lowerCamelCase names. What is `client_ids` in the Protobuf should be called `clientIds` in TypeScript.

max-vogler: Please use optional ReadonlyArray for all repeated Protobuf fields, e.g. `readonly clientIds?: ReadonlyArray<string>`. Generally, we try to keep data structures mostly immutable to leverage advantages of function programming (e.g. no unexpected modifications through side-effects, easy to understand code).

grr/server/grr_response_server/gui/ui/lib/api/http_api_service.ts
```diff
@@ -215,6 +215,17 @@ export class HttpApiService {
     return this.http.post<ApiExplainGlobExpressionResult>(url, args).pipe(
         map(result => result.components ?? []));
   }
+
+  addClientLabel(clientId: string, label: string): Observable<ApiAddClientsLabelsArgs> {
+    const url = `${URL_PREFIX}/clients/labels/add`;
+    return this.http.post<ApiAddClientsLabelsArgs>(url, {client_ids: [clientId], labels: [label]});
```
max-vogler, line 221: The generic type parameter in angle brackets specifies the return type, not the argument type. This API handler does not return a real payload (it only returns `{status: 'ok'}`), so you can use `{}` to specify that the HTTP response contains an object with unknown properties.

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
```diff
@@ -14,6 +14,15 @@ function createKnowledgeBase(kb: ApiKnowledgeBase): KnowledgeBase {
   };
 }
 
+export function translateClientLabel(label: ApiClientLabel): ClientLabel {
```
max-vogler, line 17: Please add a one-liner JSDoc to all exported symbols (functions, constants, ...).

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
```diff
@@ -14,6 +14,15 @@ function createKnowledgeBase(kb: ApiKnowledgeBase): KnowledgeBase {
...
+  let owner = '';
+
+  if (!label.name) throw new Error('name attribute is missing.');
+  if (label.owner) owner = label.owner;
```
max-vogler, line 21: The purpose of functions in api_translation is to convert API data structures that are heavily influenced by the Protobuf language to pure, plain-old TypeScript objects. In Proto best practices, all fields should be `optional`, because `required` fields can NEVER be changed to `optional` in the future.

Assigning the empty string to `label.owner` is a hack to satisfy the type system, but makes using ClientLabel hard. With an empty string as replacement for `undefined`, code now needs to check for the special value of `''` to really be sure that the ClientLabel has an owner.

Please instead, throw an Error if label.owner is false-y.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
danielberbece: Got it

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
danielberbece: Is there an exception to this rule?

grr/server/grr_response_server/gui/ui/lib/api/http_api_service.ts
danielberbece: That was a novice mistake, I must say

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
danielberbece: I tried throwing an error when the `owner` property is undefined, but the result of this was an empty autocomplete list. What do you think could be the issue and a possible solution?

max-vogler: Is the error occurring in tests or when you live test it in the browser? You can walk me through in a VC

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
max-vogler: Yeah, we are not 100 % strict about the immutability. Sometimes the type system does not permit, e.g. NgRx expects a mutable `string[]`.

danielberbece:
I updated the pull request. The API doesn't send the owner when requesting all clients labels, so I addressed this issue without affecting the current implementation of the ClientLabel. The PR is ready for review now.

grr/server/grr_response_server/gui/ui/store/config_facade.ts
```diff
@@ -55,6 +62,14 @@ export class ConfigStore extends ComponentStore<ConfigState> {
           }),
           ));
 
+  private readonly fetchClientsLabels = this.effect<void>(
+      obs$ => obs$.pipe(
+          switchMapTo(this.httpApiService.fetchAllClientsLabels()),
+          map(apiClientsLabels =>
+              apiClientsLabels.map(getApiClientLabelName)),
```
max-vogler, line 69: Please add a short comment why you only get the label name (stating that the API includes `owner` inconsistently).

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -54,9 +54,9 @@ interface ClientPageState {
 })
 export class ClientPageStore extends ComponentStore<ClientPageState> {
   constructor(
-      private readonly httpApiService: HttpApiService,
-      private readonly configService: ConfigService,
-      private readonly configFacade: ConfigFacade,
+    private readonly httpApiService: HttpApiService,
```
max-vogler, line 57: Your PR has hundreds of lines reformatted with whitespace changes. This pollutes the `git blame` view and makes it hard to track actual changes. Also, our internal formatter will override it again.

Please use [clang format](https://github.com/xaverh/vscode-clang-format-provider) to format your code. Use `"clang-format.style": "{BasedOnStyle: google}",` in the settings. If this does not work for you, please try https://github.com/google/gts.

PR #811, opened 2020-07-27
OpenAPI description generation for the GRR API

alexandrucosminmihai:
- Implement a new GRR API method that generates and returns the OpenAPI description of the GRR API
- Add associated Python client library method
- Write unit tests and an integration test

EDIT:
Please note that I use the ":+1:" emoji to mark comments that I've addressed in a yet to be created/pushed commit at the time of the emoji usage.

### DIFF
--- a/api_client/python/grr_api_client/api.py
+++ b/api_client/python/grr_api_client/api.py
@@ -80,6 +80,7 @@ class GrrApi(object):
     return self._context.username
 
   def GetOpenApiDescription(self) -> Dict[str, Any]:
+    """Returns the OpenAPI description of the GRR API as a dictionary."""
     return metadata.GetOpenApiDescription(context=self._context)
 
 
--- a/api_client/python/grr_api_client/api.py
+++ b/api_client/python/grr_api_client/api.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import unicode_literals
 
-from typing import Text
+from typing import Text, Dict, Any
 
 from grr_api_client import artifact
 from grr_api_client import client
@@ -79,7 +79,7 @@ class GrrApi(object):
   def username(self):
     return self._context.username
 
-  def GetOpenApiDescription(self):
+  def GetOpenApiDescription(self) -> Dict[str, Any]:
     return metadata.GetOpenApiDescription(context=self._context)
 
 
--- a/api_client/python/grr_api_client/api.py
+++ b/api_client/python/grr_api_client/api.py
@@ -79,8 +79,8 @@ class GrrApi(object):
   def username(self):
     return self._context.username
 
-  def GetOpenApiDescription(self, as_string=False):
-    return metadata.GetOpenApiDescription(as_string, context=self._context)
+  def GetOpenApiDescription(self):
+    return metadata.GetOpenApiDescription(context=self._context)
 
 
 def InitHttp(api_endpoint=None,
--- a/api_client/python/grr_api_client/api.py
+++ b/api_client/python/grr_api_client/api.py
@@ -15,6 +15,7 @@ from grr_api_client import root
 from grr_api_client import types
 from grr_api_client import user
 from grr_api_client import yara
+from grr_api_client import metadata
 from grr_api_client.connectors import http_connector
 
 
@@ -77,6 +78,9 @@ class GrrApi(object):
   @property
   def username(self):
     return self._context.username
+
+  def GetOpenApiDescription(self, as_string=False):
+    return metadata.GetOpenApiDescription(as_string, context=self._context)
 
 
 def InitHttp(api_endpoint=None,
--- a/api_client/python/grr_api_client/metadata.py
+++ b/api_client/python/grr_api_client/metadata.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python
+"""Metadata-related part of GRR API client library."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import json
+
+
+def GetOpenApiDescription(as_string=False, context=None):
+  if not context:
+    raise ValueError("context can't be empty")
+
+  openapi_proto = context.SendRequest("GetOpenApiDescription", None)
+  openapi_json = openapi_proto.openapi_description
+
+  if as_string:
+    return openapi_json
+
+  return json.loads(openapi_json)
--- a/api_client/python/grr_api_client/metadata.py
+++ b/api_client/python/grr_api_client/metadata.py
@@ -17,7 +17,7 @@ def GetOpenApiDescription(
   if not context:
     raise ValueError("context can't be empty")
 
-  open_api_proto = context.SendRequest("GetOpenApiDescription", None)
-  open_api_json = open_api_proto.open_api_description
+  openapi_proto = context.SendRequest("GetOpenApiDescription", None)
+  openapi_json = openapi_proto.openapi_description
 
-  return json.loads(open_api_json)
+  return json.loads(openapi_json)
--- a/api_client/python/grr_api_client/metadata.py
+++ b/api_client/python/grr_api_client/metadata.py
@@ -6,8 +6,13 @@ from __future__ import unicode_literals
 
 import json
 
+from typing import Dict, Any
+from grr_api_client.context import GrrApiContext
 
-def GetOpenApiDescription(context=None):
+
+def GetOpenApiDescription(
+    context: GrrApiContext = None
+) -> Dict[str, Any]:
   """Returns the OpenAPI description of the GRR API as a dictionary."""
   if not context:
     raise ValueError("context can't be empty")
--- a/api_client/python/grr_api_client/metadata.py
+++ b/api_client/python/grr_api_client/metadata.py
@@ -7,14 +7,12 @@ from __future__ import unicode_literals
 import json
 
 
-def GetOpenApiDescription(as_string=False, context=None):
+def GetOpenApiDescription(context=None):
+  """Returns the OpenAPI description of the GRR API as a dictionary."""
   if not context:
     raise ValueError("context can't be empty")
 
-  openapi_proto = context.SendRequest("GetOpenApiDescription", None)
-  openapi_json = openapi_proto.openapi_description
+  open_api_proto = context.SendRequest("GetOpenApiDescription", None)
+  open_api_json = open_api_proto.open_api_description
 
-  if as_string:
-    return openapi_json
-
-  return json.loads(openapi_json)
+  return json.loads(open_api_json)
--- a/grr/proto/grr_response_proto/api/metadata.proto
+++ b/grr/proto/grr_response_proto/api/metadata.proto
@@ -14,5 +14,5 @@ message ApiGetGrrVersionResult {
 
 // A message type representing the OpenAPI description of the entire API.
 message ApiGetOpenApiDescriptionResult {
-  optional string open_api_description = 1;
+  optional string openapi_description = 1;
 }
--- a/grr/proto/grr_response_proto/api/metadata.proto
+++ b/grr/proto/grr_response_proto/api/metadata.proto
@@ -14,5 +14,5 @@ message ApiGetGrrVersionResult {
 
 // A message type representing the OpenAPI description of the entire API.
 message ApiGetOpenApiDescriptionResult {
-  optional string placeholder = 1;
+  optional string openapi_description = 1;
 }
--- a/grr/proto/grr_response_proto/api/metadata.proto
+++ b/grr/proto/grr_response_proto/api/metadata.proto
@@ -14,5 +14,5 @@ message ApiGetGrrVersionResult {
 
 // A message type representing the OpenAPI description of the entire API.
 message ApiGetOpenApiDescriptionResult {
-  optional string openapi_description = 1;
+  optional string open_api_description = 1;
 }
--- a/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
+++ b/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
@@ -25,13 +25,10 @@ class ApiClientLibMetadataTest(api_integration_test_lib.ApiIntegrationTest):
   )
   def testOpenApiDescriptionValidity(self):
     import openapi_spec_validator
-    openapi_dict = self.api.GetOpenApiDescription(as_string=False)
-    try:
-      errors = openapi_spec_validator.validate_spec(openapi_dict)
-    except:
-      errors = True
+    open_api_dict = self.api.GetOpenApiDescription()
 
-    self.assertIsNone(errors)
+    # Will raise Exceptions when the OpenAPI specification is invalid.
+    openapi_spec_validator.validate_spec(open_api_dict)
 
 
 def main(argv):
--- a/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
+++ b/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python
 # Lint as: python3
-# -*- encoding: utf-8 -*-
 """Tests for API client and metadata-related API calls."""
 from __future__ import absolute_import
 from __future__ import division
@@ -23,11 +21,13 @@ class ApiClientLibMetadataTest(api_integration_test_lib.ApiIntegrationTest):
     "The openapi-spec-validator module used for validating the OpenAPI "
     "specification is not installed."
   )
-  def testOpenApiDescriptionValidity(self):
+  def testGeneratedOpenApiDescriptionIsValid(self):
+    # TODO(alexandrucosminmihai): Move this import to the top as soon as GitHub
+    # issue #813 (https://github.com/google/grr/issues/813) is resolved.
     import openapi_spec_validator
     open_api_dict = self.api.GetOpenApiDescription()
 
-    # Will raise Exceptions when the OpenAPI specification is invalid.
+    # Will raise exceptions when the OpenAPI specification is invalid.
     openapi_spec_validator.validate_spec(open_api_dict)
 
 
--- a/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
+++ b/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
@@ -25,10 +25,10 @@ class ApiClientLibMetadataTest(api_integration_test_lib.ApiIntegrationTest):
     # TODO(alexandrucosminmihai): Move this import to the top as soon as GitHub
     # issue #813 (https://github.com/google/grr/issues/813) is resolved.
     import openapi_spec_validator
-    open_api_dict = self.api.GetOpenApiDescription()
+    openapi_dict = self.api.GetOpenApiDescription()
 
     # Will raise exceptions when the OpenAPI specification is invalid.
-    openapi_spec_validator.validate_spec(open_api_dict)
+    openapi_spec_validator.validate_spec(openapi_dict)
 
 
 def main(argv):
--- a/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
+++ b/grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python
+# Lint as: python3
+# -*- encoding: utf-8 -*-
+"""Tests for API client and metadata-related API calls."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import pkg_resources
+
+from absl import app
+
+from grr_response_server.gui import api_integration_test_lib
+from grr.test_lib import skip
+from grr.test_lib import test_lib
+
+
+class ApiClientLibMetadataTest(api_integration_test_lib.ApiIntegrationTest):
+  """Tests metadata-related part of GRR Python API client library."""
+
+  @skip.If(
+    "openapi-spec-validator" not in {p.key for p in pkg_resources.working_set},
+    "The openapi-spec-validator module used for validating the OpenAPI "
+    "specification is not installed."
+  )
+  def testOpenApiDescriptionValidity(self):
+    import openapi_spec_validator
+    openapi_dict = self.api.GetOpenApiDescription(as_string=False)
+    try:
+      errors = openapi_spec_validator.validate_spec(openapi_dict)
+    except:
+      errors = True
+
+    self.assertIsNone(errors)
+
+
+def main(argv):
+  test_lib.main(argv)
+
+
+if __name__ == "__main__":
+  app.run(main)
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -636,11 +636,9 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       schemas_obj[type_name] = self.schema_objs[type_name]
 
     # The Components Object "components" of the root OpenAPI object.
-    components_obj = {
+    return {
       "schemas": cast(Dict[str, Union[EnumSchema, MessageSchema]], schemas_obj)
     }
-
-    return components_obj
 
   def _SeparateFieldsIntoParams(
       self,
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -28,28 +28,79 @@ MessageSchema = Dict[
              Dict[str, Union[SchemaReference, PrimitiveSchema, ArraySchema]]]
 ]
 Schema = Union[PrimitiveSchema, EnumSchema, MessageSchema, ArraySchema]
+PrimitiveDescription = Dict[str, Union[str, PrimitiveSchema]]
 
-proto_primitive_types_names: Dict[int, str] = {
-  protobuf2.TYPE_DOUBLE: "protobuf2.TYPE_DOUBLE",
-  protobuf2.TYPE_FLOAT: "protobuf2.TYPE_FLOAT",
-  protobuf2.TYPE_INT64: "protobuf2.TYPE_INT64",
-  protobuf2.TYPE_UINT64: "protobuf2.TYPE_UINT64",
-  protobuf2.TYPE_INT32: "protobuf2.TYPE_INT32",
-  protobuf2.TYPE_FIXED64: "protobuf2.TYPE_FIXED64",
-  protobuf2.TYPE_FIXED32: "protobuf2.TYPE_FIXED32",
-  protobuf2.TYPE_BOOL: "protobuf2.TYPE_BOOL",
-  protobuf2.TYPE_STRING: "protobuf2.TYPE_STRING",
-  protobuf2.TYPE_BYTES: "protobuf2.TYPE_BYTES",
-  protobuf2.TYPE_UINT32: "protobuf2.TYPE_UINT32",
-  protobuf2.TYPE_SFIXED32: "protobuf2.TYPE_SFIXED32",
-  protobuf2.TYPE_SFIXED64: "protobuf2.TYPE_SFIXED64",
-  protobuf2.TYPE_SINT32: "protobuf2.TYPE_SINT32",
-  protobuf2.TYPE_SINT64: "protobuf2.TYPE_SINT64",
+primitive_types: Dict[Union[int, str], PrimitiveDescription] = {
+  protobuf2.TYPE_DOUBLE: {
+    "name": "protobuf2.TYPE_DOUBLE",
+    "schema": {"type": "number", "format": "double"},
+  },
+  protobuf2.TYPE_FLOAT: {
+    "name": "protobuf2.TYPE_FLOAT",
+    "schema": {"type": "number", "format": "float"},
+  },
+  protobuf2.TYPE_INT64: {
+    "name": "protobuf2.TYPE_INT64",
+    "schema": {"type": "integer", "format":"int64"},
+  },
+  protobuf2.TYPE_UINT64: {
+    "name": "protobuf2.TYPE_UINT64",
+    "schema": {"type": "integer", "format":"uint64"},
+  },
+  protobuf2.TYPE_INT32: {
+    "name": "protobuf2.TYPE_INT32",
+    "schema": {"type": "integer", "format": "int32"},
+  },
+  protobuf2.TYPE_FIXED64: {
+    "name": "protobuf2.TYPE_FIXED64",
+    "schema": {"type": "integer", "format": "uint64"}
+  },
+  protobuf2.TYPE_FIXED32: {
+    "name": "protobuf2.TYPE_FIXED32",
+    "schema": {"type": "integer", "format": "uint32"},
+  },
+  protobuf2.TYPE_BOOL: {
+    "name": "protobuf2.TYPE_BOOL",
+    "schema": {"type": "boolean"},
+  },
+  protobuf2.TYPE_STRING: {
+    "name": "protobuf2.TYPE_STRING",
+    "schema": {"type": "string"},
+  },
+  protobuf2.TYPE_BYTES: {
+    "name": "protobuf2.TYPE_BYTES",
+    # TODO: Here, "byte" (base64) might be used ?
+    "schema": {"type": "string", "format": "binary"},
+  },
+  protobuf2.TYPE_UINT32: {
+    "name": "protobuf2.TYPE_UINT32",
+    "schema": {"type": "integer", "format": "uint32"},
+  },
+  protobuf2.TYPE_SFIXED32: {
+    "name": "protobuf2.TYPE_SFIXED32",
+    "schema": {"type": "integer", "format": "int32"},
+  },
+  protobuf2.TYPE_SFIXED64: {
+    "name": "protobuf2.TYPE_SFIXED64",
+    "schema": {"type": "integer","format": "int64"},
+  },
+  protobuf2.TYPE_SINT32: {
+    "name": "protobuf2.TYPE_SINT32",
+    "schema": {"type": "integer", "format": "int32"},
+  },
+  protobuf2.TYPE_SINT64: {
+    "name": "protobuf2.TYPE_SINT64",
+    "schema": {"type": "integer","format": "int64"},
+  },
+  "BinaryStream": {
+    "name": "BinaryStream",
+    "schema": {"type": "string", "format": "binary"},
+  }
 }
 
-primitive_types_names: Set[str] = (
-    set(proto_primitive_types_names.values()) | {"BinaryStream"}
-)
+primitive_types_names: Set[str] = {
+  primitive_type["name"] for primitive_type in primitive_types.values()
+}
 
 class ApiGetGrrVersionResult(rdf_structs.RDFProtoStruct):
   """An RDF wrapper for result of the API method for getting GRR version."""
@@ -101,20 +152,20 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     self.open_api_obj: Optional[Dict[str, Union[str, List, Dict]]] = None
     self.schema_objs: Optional[Dict[str, Schema]] = None
 
-  def _SimplifyPathNode(self, node: str) -> str:
+  def _NormalizePathComponent(self, component: str) -> str:
     """Normalize the given path component to be used in a valid OpenAPI path."""
-    if len(node) > 0 and node[0] == '<' and node[-1] == '>':
-      node = node[1:-1]
-      node = node.split(":")[-1]
-      node = f"{{{node}}}"
+    if len(component) > 0 and component[0] == '<' and component[-1] == '>':
+      component = component[1:-1]
+      component = component.split(":")[-1]
+      component = f"{{{component}}}"
 
-    return node
+    return component
 
-  def _SimplifyPath(self, path: str) -> str:
+  def _NormalizePath(self, path: str) -> str:
     """Keep only fixed parts and parameter names from Werkzeug URL patterns.
 
     The OpenAPI specification requires that parameters are surrounded by { }
-    which are added in _SimplifyPathNode.
+    which are added in _NormalizePathComponent.
 
     Args:
       path: The path whose representation will be normalized.
@@ -123,23 +174,25 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       The normalized version of the path argument, which is a valid OpenAPI
       path, with curly brackets around path arguments.
     """
-    nodes = path.split("/")
-    simple_nodes = [self._SimplifyPathNode(node) for node in nodes]
+    components = path.split("/")
+    normalized_components = [
+      self._NormalizePathComponent(component) for component in components
+    ]
 
-    simple_path = '/'.join(simple_nodes)
+    normalized_path = '/'.join(normalized_components)
 
-    return simple_path
+    return normalized_path
 
   def _GetPathArgsFromPath(self, path: str) -> List[str]:
     """Extract path parameters from a Werkzeug Rule URL."""
     path_args = []
 
-    nodes = path.split("/")
-    for node in nodes:
-      if len(node) > 0 and node[0] == '<' and node[-1] == '>':
-        simple_node = self._SimplifyPathNode(node)
-        simple_node = simple_node[1:-1]
-        path_args.append(simple_node)
+    components = path.split("/")
+    for component in components:
+      if len(component) > 0 and component[0] == '<' and component[-1] == '>':
+        normalized_component = self._NormalizePathComponent(component)
+        normalized_component = normalized_component[1:-1]
+        path_args.append(normalized_component)
 
     return path_args
 
@@ -166,7 +219,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       return cls.__name__
 
     if isinstance(cls, int):  # It's a protobuf.Descriptor.type value.
-      return proto_primitive_types_names[cls] # TODO: Delete this. I could use a Dict[proto_int_const, proto_detail].
+      return primitive_types[cls]["name"]
 
     return str(cls)  # Cover "BinaryStream" and None.
 
@@ -213,75 +266,15 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     self.open_api_obj.update(metadata_dict)
 
   def _AddPrimitiveTypesSchemas(self) -> None:
-    """Creates OpenAPI schemas for Protobuf primitives and BinaryStream."""
+    """Adds the OpenAPI schemas for Protobuf primitives and BinaryStream."""
     if self.schema_objs is None:  # Check required by mypy.
       raise AssertionError(
         "The container of OpenAPI type schemas is not initialized."
       )
 
     primitive_types_schemas = {
-      proto_primitive_types_names[protobuf2.TYPE_DOUBLE]: {
-        "type": "number",
-        "format": "double"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_FLOAT]: {
-        "type": "number",
-        "format": "float"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_INT64]: {
-        "type": "integer",
-        "format": "int64"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_UINT64]: {
-        "type": "integer",
-        "format": "uint64"  # Undefined by the OpenAPI Specification (OAS).
-      },
-      proto_primitive_types_names[protobuf2.TYPE_INT32]: {
-        "type": "integer",
-        "format": "int32"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_FIXED64]: {
-        "type": "integer",
-        "format": "uint64"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_FIXED32]: {
-        "type": "integer",
-        "format": "uint32"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_BOOL]: {
-        "type": "boolean"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_STRING]: {
-        "type": "string"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_BYTES]: {
-        "type": "string",
-        "format": "binary"  # TODO: Here "byte" (base64) might be used?
-      },
-      proto_primitive_types_names[protobuf2.TYPE_UINT32]: {
-        "type": "integer",
-        "format": "uint32"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_SFIXED32]: {
-        "type": "integer",
-        "format": "int32"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_SFIXED64]: {
-        "type": "integer",
-        "format": "int64"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_SINT32]: {
-        "type": "integer",
-        "format": "int32"
-      },
-      proto_primitive_types_names[protobuf2.TYPE_SINT64]: {
-        "type": "integer",
-        "format": "int64"
-      },
-      "BinaryStream": {
-        "type": "string",
-        "format": "binary"
-      }
+      primitive_type["name"]: primitive_type["schema"]
+      for primitive_type in primitive_types.values()
     }
 
     self.schema_objs.update(primitive_types_schemas)
@@ -649,14 +642,14 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     for router_method_name in router_methods:
       router_method = router_methods[router_method_name]
       for http_method, path, strip_root_types in router_method.http_methods:
-        simple_path = self._SimplifyPath(path)
+        normalized_path = self._NormalizePath(path)
         path_args = set(self._GetPathArgsFromPath(path))
 
-        if simple_path not in paths_obj:
-          paths_obj[simple_path] = dict()
+        if normalized_path not in paths_obj:
+          paths_obj[normalized_path] = dict()
 
         # The Path Object associated with the current path.
-        path_obj = paths_obj[simple_path]
+        path_obj = paths_obj[normalized_path]
 
         url_path = (
           path
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -93,9 +93,9 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
   result_type = ApiGetOpenApiDescriptionResult
 
   def __init__(self, router: Any) -> None:
-    # TODO(hanuszczak): Break dependency cycle between this module
-    # and `api_call_router.py` and then use ApiCallRouter as self.router's and
-    # the argument's type.
+    # TODO(hanuszczak): Break dependency cycle between this module and
+    # `api_call_router.py` and then use ApiCallRouter as self.router's and the
+    # argument's type.
     self.router: Any = router
     # The main OpenAPI description object.
     self.open_api_obj: Optional[Dict[str, Union[str, List, Dict]]] = None
@@ -166,7 +166,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       return cls.__name__
 
     if isinstance(cls, int):  # It's a protobuf.Descriptor.type value.
-      return proto_primitive_types_names[cls]
+      return proto_primitive_types_names[cls] # TODO: Delete this. I could use a Dict[proto_int_const, proto_detail].
 
     return str(cls)  # Cover "BinaryStream" and None.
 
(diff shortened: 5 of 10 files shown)


grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
```diff
@@ -0,0 +1,42 @@
...
+  )
+  def testOpenApiDescriptionValidity(self):
+    import openapi_spec_validator
+    openapi_dict = self.api.GetOpenApiDescription(as_string=False)
+    try:
+      errors = openapi_spec_validator.validate_spec(openapi_dict)
+    except:
+      errors = True
+
+    self.assertIsNone(errors)
```
mbushkov, line 34: I wouldn't put the validator call into try/except. Currently, if the test fails, it's not possible to distinguish cases when it reported errors or actually raised.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+  def Method6TypeReferences(self, args, token=None):
+    """Method 6 description."""
+
+
+class ApiGetOpenApiDescriptionHandlerTest(api_test_lib.ApiCallHandlerTest):
+  """Test for ApiGetOpenApiDescriptionHandler."""
+  def setUp(self):
+    super(ApiGetOpenApiDescriptionHandlerTest, self).setUp()
+    self.router = DummyApiCallRouter()
+    self.router_methods = self.router.__class__.GetAnnotatedMethods()
```
mbushkov, line 109: Please add an assertion that would check that we have all the expected methods in self.router_methods.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    "specification is not installed."
+  )
+  def testRendersValidOpenApiDescription(self):
+    import openapi_spec_validator
+    try:
+      errors = openapi_spec_validator.validate_spec(self.openapi_desc_dict)
+    except:
+      errors = True
+
+    self.assertIsNone(errors)
```
mbushkov, line 129: Same as one of the above comments - I'm not sure we gain much by catching all the exceptions.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    self.openapi_desc = result.openapi_description
+    self.openapi_desc_dict = json.loads(self.openapi_desc)
+
+  @skip.If(
+    "openapi-spec-validator" not in {p.key for p in pkg_resources.working_set},
+    "The openapi-spec-validator module used for validating the OpenAPI "
+    "specification is not installed."
+  )
+  def testRendersValidOpenApiDescription(self):
+    import openapi_spec_validator
```
mbushkov, line 123: Please open a Github issue to make the openapi spec validator a GRR test package dependency and add a TODO here to move this import to the top as soon as this issue is resolved.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+
+        request_body_obj = operation_obj.get("requestBody")
+        if request_body_obj:
+          schema_obj = request_body_obj["content"]["application/json"]["schema"]
+          for param_name in schema_obj["properties"]:
+            openapi_body_params.add(param_name)
+
+        self.assertSetEqual(path_params, openapi_path_params)
+        self.assertSetEqual(query_params, openapi_query_params)
+        self.assertSetEqual(body_params, openapi_body_params)
```
mbushkov, line 199: I think this test might be improved. The main problem: the test contains loops and a lot of conditional logic. This means that it's somewhat hard to read, but also that it might have hard to spot bugs.

What this test seems to be doing is checking that openapi spec path, query and body params are figured out correctly. Instead of using a generic testing logic, I'd refactor this test to be as simple as possible: something like a series of assertEqual calls (note that assertEqual works fine on complex nested data structures). Here the dummy router is used, meaning that we know the API methods we work with and the expected results for each method.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+      for http_method, path, strip_root_types in router_method.http_methods:
+        simple_path = self.handler._SimplifyPath(path)
+        http_method = http_method.lower()
+
+        operation_obj = \
+          self.openapi_desc_dict["paths"][simple_path][http_method]
+
+        self.assertIsNotNone(operation_obj)
+
+  def testRouteArgsSeparationInPathQueryBody(self):
```
mbushkov, line 142: Please use "[test subject] [verb] [goal]" naming scheme. I.e. testRouteArgsAreCorrectlySeparated.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+        if request_body_obj:
+          schema_obj = request_body_obj["content"]["application/json"]["schema"]
+          for param_name in schema_obj["properties"]:
+            openapi_body_params.add(param_name)
+
+        self.assertSetEqual(path_params, openapi_path_params)
+        self.assertSetEqual(query_params, openapi_query_params)
+        self.assertSetEqual(body_params, openapi_body_params)
+
+  def testRouteResultOpenApiDescription(self):
```
mbushkov, line 201: Same comment about the test name. Maybe: `testRouteResultsArePresentInOpenApiSpec` or something along these lines?

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    expected["field_sfixed64"] = expected_sfixed64
+
+    expected_sint32 = {"type": "integer", "format": "int32"}
+    expected["field_sint32"] = expected_sint32
+
+    expected_sint64 = {"type": "integer", "format": "int64"}
+    expected["field_sint64"] = expected_sint64
+
+    expected_binarystream = {"type": "string", "format": "binary"}
+    expected["BinaryStream"] = expected_binarystream
```
mbushkov, line 291: Would it be better to just build a Python object like:
```
expected = {
  ...
  "BinaryStream": {"type": "string", "format": "binary"}
}
```
?

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+
+    self.assertDictEqual(expected, openapi_primitives)
+
+  def testRepetitiveField(self):
```
mbushkov, line 295: In protobufs `repeated` is the term used for multiple-values fields. Also, please indicate the goal of the test. I.e. `testRepeatedFieldsAreCorrectlyPresentInOpenApiSpec`.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    # Test the GET route. The repeated field should be a query parameter.
+    params = (self.openapi_desc_dict.get("paths").get("/metadata_test/method4")
+              .get("get").get("parameters"))
+    self.assertEqual(len(params), 1)
```
mbushkov, line 299: Please use assertLen.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+
+    param = params[0]
+    self.assertEqual(param["in"], "query")
+    schema = param["schema"]
+    self.assertEqual(schema["type"], "array")
+    self.assertEqual(schema["items"]["type"], "integer")
+    self.assertEqual(schema["items"]["format"], "int64")
```
mbushkov, line 306: Can we just do a single assertEqual for the whole `schema` variable?

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+
+    # Test the POST route. The repeated field should be in the request body.
+    req_body = (self.openapi_desc_dict.get("paths")
+                .get("/metadata_test/method4").get("post").get("requestBody"))
+    req_schema = req_body["content"]["application/json"]["schema"]
+    self.assertEqual(req_schema["type"], "object")
+    schema = req_schema["properties"]["field_repeated"]
+    self.assertEqual(schema["type"], "array")
+    self.assertEqual(schema["items"]["type"], "integer")
+    self.assertEqual(schema["items"]["format"], "int64")
```
mbushkov, line 316: Same question here: can we compare the whole `schema` object?

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+
+  def testEnumField(self):
```
mbushkov, line 318: Same comment about the test method name.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    self.assertEqual(len(params), 1)
+
+    param = params[0]
+    self.assertEqual(param["in"], "query")
+    ref_get = param["schema"]["$ref"]
+    ref_nodes = ref_get.split("/")
+    ref_nodes = ref_nodes[1:]
+    schema = self.openapi_desc_dict
+    for node in ref_nodes:
+      schema = schema[node]
```
mbushkov, line 331: Here the testing logic is pretty complicated. For example, as a reader, I'm not sure I can assess whether why `ref_nodes = ref_nodes[1:]` is done. Would it be possible to simplify this by using predefined literals? - since we know which values to expect.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    self.assertEqual(schema["format"], "int32")
+    enum_values_openapi = set(schema["enum"])
+
+    message_d = MetadataEnumFieldMessage.protobuf.DESCRIPTOR
+    self.assertEqual(len(message_d.enum_types), 1)
+    enum_d = message_d.enum_types[0]
+    enum_values_ds = enum_d.values
+    enum_values_protobuf = set(
+      [enum_value_d.number for enum_value_d in enum_values_ds])
+    self.assertSetEqual(enum_values_openapi, enum_values_protobuf)
```
mbushkov, line 344: I would use literals here instead of relying on reflection to fetch enum_values_protobuf. Given that we control the test data, we know what to expect and can explicitly list the enum values we expect.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,382 @@
...
+    # Just check that the structure is correct and that the same component
+    # is referenced as above.
+    req_body = (self.openapi_desc_dict.get("paths")
+                .get("/metadata_test/method5").get("post").get("requestBody"))
+    req_schema = req_body["content"]["application/json"]["schema"]
+    self.assertEqual(req_schema["type"], "object")
+    ref_post = req_schema["properties"]["field_enum"]["$ref"]
+    self.assertEqual(ref_post, ref_get)
+
+  def testOpenApiTypeReferences(self):
```
mbushkov, line 356: Same comment about the test name. Should we use something like `testReferencesInCompositeTypesAreCorrectlyProcessed` or something along these lines?

grr/proto/grr_response_proto/api/metadata.proto
```diff
@@ -14,5 +14,5 @@ message ApiGetGrrVersionResult {
 
 // A message type representing the OpenAPI description of the entire API.
 message ApiGetOpenApiDescriptionResult {
-  optional string placeholder = 1;
+  optional string openapi_description = 1;
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you! I updated the use of "openapi" to "open_api" in this PR and kept "description" as we've discussed.

api_client/python/grr_api_client/api.py
```diff
@@ -78,6 +79,9 @@ def UploadYaraSignature(self, signature: Text) -> bytes:
   def username(self):
     return self._context.username
 
+  def GetOpenApiDescription(self, as_string=False):
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you! (For future reference: We've talked about this and we decided to keep "description" as it seems to be more used for this [use-case](https://github.blog/2020-07-27-introducing-githubs-openapi-description/), while "specification" is used to describe the [OpenAPI](https://swagger.io/specification/) standard).

api_client/python/grr_api_client/metadata.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
+"""Metadata-related part of GRR API client library."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import json
+
+
+def GetOpenApiDescription(as_string=False, context=None):
```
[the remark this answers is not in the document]
alexandrucosminmihai: I removed the `as_string` argument. Thank you!

grr/proto/grr_response_proto/api/metadata.proto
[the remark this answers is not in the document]
panhania: I have to slightly disagree here. While it is not a strong preference, I would pick `openai`/`Openapi` variants over `open_api`/`OpenApi`. The reason is that it is kind of one word, not `Open` and `API` separated by space. While not directly related, because different languages have different rules, it also looks like a general convention:

  * [here](https://metacpan.org/pod/Mojolicious::Plugin::OpenAPI) they use `openapi` as the variable/key
  * [here](https://loopback.io/doc/en/lb4/apidocs.openapi-v3.html) they use `openapi-v3` as the package name

Some other precedents with the `Open*` naming:

  * [OpenML](https://www.openml.org/) uses [`openml`](https://github.com/openml) as the name of the GitHub organisation and `openml-*` for language-specific libraries (e.g. [`openml-python`](https://github.com/openml/openml-python))
  * [OpenAI](https://openai.com/) uses [`openai`](https://github.com/openai) as the name of the GitHub organisation and as a Python [PIP package](https://pypi.org/project/openai/)
  * Valve uses [`openvr`](https://github.com/ValveSoftware/openvr) for the name of OpenVR repository

Finally, `openapi` has much better googlability then `open_api` (and searching `open_api` auto-corrects to `openapi`).

api_client/python/grr_api_client/metadata.py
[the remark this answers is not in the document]
panhania: Also, having such arguments also introduces type incorrectness. It is better to have two separate methods, one that returns a string and one that returns a proper object than to have too much functionality into one.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
panhania: In general, in tests always prefer readability and obviousness over conciseness and being generic. Tests should read like a linear scenario rather then force the reader to do some mental execution of the code.

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
```diff
@@ -0,0 +1,39 @@
...
+  """Tests metadata-related part of GRR Python API client library."""
+
+  @skip.If(
+    "openapi-spec-validator" not in {p.key for p in pkg_resources.working_set},
+    "The openapi-spec-validator module used for validating the OpenAPI "
+    "specification is not installed."
+  )
+  def testOpenApiDescriptionValidity(self):
+    import openapi_spec_validator
+    open_api_dict = self.api.GetOpenApiDescription()
```
panhania, line 28: +1 point for using `openapi` over `open_api` (inconsistency with `openapi_spec_validator).

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Will raise Exceptions when the OpenAPI specification is invalid.
```
panhania, line 30: Why `Exceptions` is capitalized? It looks like some identifier. If it is I would prefer it to be backticked, if not, it should be lowercase.

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
panhania, line 1: I don't think this is needed or useful.

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
```diff
@@ -0,0 +1,39 @@
...
+# Lint as: python3
+# -*- encoding: utf-8 -*-
```
panhania, line 3: Since Python 3, UTF-8 is the default encoding, no need to specify that.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      protobuf2.TYPE_BYTES: "protobuf2.TYPE_BYTES",
+      protobuf2.TYPE_UINT32: "protobuf2.TYPE_UINT32",
+      protobuf2.TYPE_SFIXED32: "protobuf2.TYPE_SFIXED32",
+      protobuf2.TYPE_SFIXED64: "protobuf2.TYPE_SFIXED64",
+      protobuf2.TYPE_SINT32: "protobuf2.TYPE_SINT32",
+      protobuf2.TYPE_SINT64: "protobuf2.TYPE_SINT64",
+    }
+    self.primitive_types_names = \
+      list(self.proto_primitive_types_names.values()) \
+      + ["BinaryStream",]
```
panhania, line 87: Using parens is preferred over line-continuation with `\`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
   args_type = None
   result_type = ApiGetOpenApiDescriptionResult
 
+  def __init__(self, router):
```
panhania, line 63: Missing type annotation.

api_client/python/grr_api_client/metadata.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+"""Metadata-related part of GRR API client library."""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import unicode_literals
+
+import json
+
+
+def GetOpenApiDescription(context=None):
```
panhania, line 10: Missing type annotation.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    server_obj = dict()
+    server_obj["url"] = "/"
+    server_obj["description"] = "Root path of the GRR API"
+    self.open_api_obj["servers"].append(server_obj)
+
+  def _AddPrimitiveTypesSchemas(self):
+    """Creates OpenAPI schemas for Protobuf primitives and BinaryStream."""
+    int_to_name = self.proto_primitive_types_names
+
+    # protobuf2.TYPE_DOUBLE == 1
```
panhania, line 183: I am not sure what is the purpose of these comments.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+
+  def _SetMetadata(self):
+    oas_version = "3.0.3"
+    self.open_api_obj["openapi"] = oas_version
+
+    # The Info Object "info" field.
+    info_obj = dict()
+    info_obj["title"] = "GRR Rapid Response API"
+    info_obj["description"] = "GRR Rapid Response is an incident response " \
+                              "framework focused on remote live forensics."
```
panhania, line 155: Why not use dictionary literals (here and below)? Also, since this is mostly static representation, maybe using a global constant for this is a better idea (something to discuss).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    nodes = path.split("/")
+    for node in nodes:
+      if len(node) > 0 and node[0] == '<' and node[-1] == '>':
+        simple_node = self._SimplifyPathNode(node)
+        simple_node = simple_node[1:-1]
+        path_args.append(simple_node)
+
+    return path_args
+
+  def _GetTypeName(self, t):
```
panhania, line 124: Prefer not to use one-letter identifiers. Since here, `t` is a shortcut for `type` (I presume), the usual Python convention is to use `cls`.

panhania: Also, try to always use type annotations. If it is not possible to type annotate something, explicitly mark it as `Any` (but here, `Type` would be more strict).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
   args_type = None
   result_type = ApiGetOpenApiDescriptionResult
 
+  def __init__(self, router):
+    self.router = router
+    self.open_api_obj = None # The main OpenAPI description object.
+    self.schema_objs = None
+
+    self.proto_primitive_types_names = {
```
panhania, line 68: This is not instance-specific, I would move it to a global variable.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+
+    self.schema_objs = dict()  # Holds OpenAPI representations of types.
+    self._AddPrimitiveTypesSchemas()
+
+    visiting = set()  # Holds state of types extraction (white/gray nodes).
+    router_methods = self.router.__class__.GetAnnotatedMethods()
+    for method_metadata in router_methods.values():
+      args_type = method_metadata.args_type
+      if args_type:
+        if inspect.isclass(args_type) \
```
panhania, line 366: Some comment about line continuation.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+            and issubclass(args_type, rdf_structs.RDFProtoStruct):
+          self._ExtractSchema(args_type.protobuf.DESCRIPTOR, visiting)
+        else:
+          self._ExtractSchema(args_type, visiting)
+
+      result_type = method_metadata.result_type
+      if result_type:
+        if inspect.isclass(result_type) \
```
panhania, line 374: Ditto.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+            and issubclass(result_type, rdf_structs.RDFProtoStruct):
+          self._ExtractSchema(result_type.protobuf.DESCRIPTOR, visiting)
+        else:
+          self._ExtractSchema(result_type, visiting)
+
+  def _SetComponents(self):
+    if self.schema_objs is None:
+      raise ValueError("Called _SetComponents before extracting schemas.")
```
panhania, line 382: That is not really a `ValueError`. More like an `AssertionError` and it leaks implementation details `_SetComponents` name.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    primitive_types_names = set(self.primitive_types_names)
+    # Create components only for composite types.
+    for type_name in type_names - primitive_types_names:
+      schemas_obj[type_name] = self.schema_objs[type_name]
+
+    components_obj["schemas"] = schemas_obj
+
+    self.open_api_obj["components"] = components_obj
+
+  def _GetSchemaOrReferenceObject(self, type_name: str, is_array=False):
```
panhania, line 398: Missing type annotation (here and in the all methods below).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    """Returns a Schema Object if primitive type, else a Reference Object.
+
+    Primitive, not composite types don't have an actual schema, but rather an
+    equivalent OpenAPI representation that gets returned for them.
+    More complex types are expected to have been previously defined as OpenAPI
+    components and are used through OpenAPI references.
+    """
```
panhania, line 405: Missing `Args` and `Returns` sections of the docstring.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+
+    if is_array:
+      schema_or_ref_obj = {
+        "type": "array",
+        "items": schema_or_ref_obj
+      }
+
+    return schema_or_ref_obj
+
+  def _GetParameters(self, path_params, query_params):
```
panhania, line 427: Missing docstring (can be one line, but method is quite long so our internal linter will surely complain). Here and in all methods below.

api_client/python/grr_api_client/metadata.py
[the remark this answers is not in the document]
alexandrucosminmihai: This makes sense, thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: I've added a separate test that asserts this. Please let me know if you want me to change it. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Made the change, thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Made the change and used predefined literals everywhere. Please let me know if they are not ok. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Thank you both for explaining me this! I had limited experience writing tests, but I believe I've learned a lot by working on this and I'm sure there is more to learn about testing and software engineering&development in general. I've made the change to literals wherever I thought it possible, but I might have missed something. Please let me know if I should change something else.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: I totally agree and made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: I opened the issue and added the comment in this test file, as well as in the integration test. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Thank you for pointing this out! I made the change to explicitly written literals throughout the testing code. Please let me know if there's something that needs to be changed.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: That's a logical and very useful notation. Made the change to all testing methods. Thank you!

grr/proto/grr_response_proto/api/metadata.proto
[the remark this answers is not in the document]
alexandrucosminmihai: I believe this is a good point. I personally like `openapi/OpenApi` because "OpenAPI" is, as @panhania  mentioned, sort of a single word with 2 components merged together to form a name and the 2 capital letters point out the 2 components while not making the word awkward to place next to and delimitate from other words (which would be the case with `OpenAPI`, ex: `def OpenAPIImplementation:`). On the other hand, I understand the readability value of keeping a naming scheme convention across a large code base, so I believe the current version of `open_api/OpenApi` is also a good choice.

Please let me know if you (@mbushkov @panhania) would like me to change the current names.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      protobuf2.TYPE_SFIXED32: "protobuf2.TYPE_SFIXED32",
+      protobuf2.TYPE_SFIXED64: "protobuf2.TYPE_SFIXED64",
+      protobuf2.TYPE_SINT32: "protobuf2.TYPE_SINT32",
+      protobuf2.TYPE_SINT64: "protobuf2.TYPE_SINT64",
+    }
+    self.primitive_types_names = \
+      list(self.proto_primitive_types_names.values()) \
+      + ["BinaryStream",]
+
+  def _SimplifyPathNode(self, node: str) -> str:
```
mol123, line 89: SimplifyPathComponent

(In /a/b/c the "parts" are called components).

Please change all occurrences of "node" to "component"

Also, it doesn't really "simplify", but it normalizes/translates?

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +60,546 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      return self.proto_primitive_types_names[t]
+
+    return str(t) # Cover "BinaryStream" and None.
+
+  def _SetMetadata(self):
+    oas_version = "3.0.3"
+    self.open_api_obj["openapi"] = oas_version
+
+    # The Info Object "info" field.
+    info_obj = dict()
```
mol123, line 152: Could this be less verbose with a dict notation? (All occurrences in this file?)

```
info = {
 "contact": {
   "name": ...
 }
```
}

EDIT: The code is indeed to verbose, and would benefit from more ```{"foo":"bar"}```

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
alexandrucosminmihai: Changed it. Thank you!

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
alexandrucosminmihai: Removed it. Thank you!

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
alexandrucosminmihai: Removed it. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Made the change to parens throughout the PR's code. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added type annotations here and to all other methods created in this PR. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I used them while writing the code and when I was done I thought they might be helpful in the future as a reference, maybe when debugging, as they represent the values of the constants. I removed them. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for pointing this out! I made the move to dictionary literals where I thought it was possible. Please let me know if you find other instances that make sense to be converted to dictionary literals.

Regarding the `info_obj`, I think it might be useful to keep it not set in stone in order to allow customization for special cases, if such special cases can exist (?). For example, would it be possible to want different info objects for different `routers` ? In this case we could override the associated method in subclasses that get instantiated by different routers' methods. Or we could use a static variable per subclass (this might require metaclasses) to replace the overriding ooor, we could pass the `info_obj` to the constructor of the `Handler` from the router's method. What do you think?

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Replaced one-letter identifiers and added type annotations throughout the code. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Moved it, thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Modified it. Thanks!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Made the change. Thanks!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Changed the type of the exception and the message for this case and all other similar cases in the code. Please let me know if they are not correct. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added them. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added them. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you! I added the missing docstring parts and new docstrings to methods that didn't have them.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Made the modification. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: I changed the name of the test methods to follow the naming scheme described in a previous comment. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: I agree. I made the changes to all such cases. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Oups, sorry about the "repetitive". Made the change and improved test method name. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Made the change to `assertLen`. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Definitely. Made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: Updated the name. Thanks!

api_client/python/grr_api_client/metadata.py
alexandrucosminmihai: Thank you! Added them.

grr/server/grr_response_server/gui/api_plugins/metadata.py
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you for pointing this out!

I replaced `self.proto_primitive_types_names` with a dictionary holding all information about primitive types, including `BinaryStream`, not only protobuf primitives. The key of the dictionary is meant to be the "internal identification token" of the type: int constants for protobuf primitive types, and the name of the type for `BinaryStream`.

I needed a dictionary instead of an array because in `_GetTypeName` I need to lookup the name of the protobuf primitives somehow (I need a mapping between the int constants and their names) and a dictionary helps to do this in O(1).

I also still need a separate set `primitive_types_names` because in `_GetSchemaOrReferenceObject` I need to differentiate between primitive and composite types which I do based on their name and the set allows for O(1) search complexity.

Please let me know if you have any suggestions for doing this in a better way.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I made the change from `Node` -> `Component` and from `Simplify` -> `Normalize`. Thank you for suggesting this! The reason I used "simplify" was because the Werkzeug URL may contain type converters which are eliminated in the resulting nodes and path, but I believe that the term "normalize" should cover this semantic as well as the idea that we're replacing "<>"s with "{}"s.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for this suggestion! I made the conversion to the `{}` dictionary notation wherever I thought it possible in the current form of the code. I also believe that the approach for representing the structure of the root OpenAPI object while building it, which you suggested in the top-level comment, is very elegant and makes the structure much easier to understand. I am currently working on making the required refactoring for this. Thank you very much!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+      "POST /metadata_test/method4": method4_path_dict["post"]["responses"],
+    }
+
+    self.assertDictEqual(expected_responses, open_api_responses)
+
+  def testPrimitiveTypesAreCorrectlyDescribedInOpenApiDescription(self):
+    expected_primitives = {
+      "field_double": {"type": "number", "format": "double"},
+      "field_float": {"type": "number", "format": "float"},
+      "field_int64": {"type": "integer", "format": "int64"},
```
max-vogler, line 555: This came up in a [recent code review](https://github.com/google/grr/pull/814#discussion_r462392915): [Protobuf serializes large integers as string](https://developers.google.com/protocol-buffers/docs/proto3), because [JavaScript loses precision](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) around `2^53-1`. Is this relevant for the OpenAPI description? Should we use `"type": "string", "format": "int64"`?

alexandrucosminmihai: Thank you very much for pointing this out, Max! I believe this is relevant for the OpenAPI description because, for example, it should let clients know what types to expect in the result JSON.

grr/proto/grr_response_proto/api/metadata.proto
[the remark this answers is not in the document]
mbushkov: @panhania, @alexandrucosminmihai  - thanks for the detailed argument in favour of the `openapi/OpenApi` - agreed!

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
mbushkov: Agreed.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
mbushkov: Looks good, thanks!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
mbushkov: Looks good!

grr/server/grr_response_server/gui/api_integration_tests/metadata_test.py
mbushkov: My understanding is that we can treat this is as a default assumption: the function that we're testing shouldn't raise. If it should raise, though, then we mark the code with self.assertRaises and the expectation of the test becomes clear.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
mbushkov: Looks good!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
mbushkov: @alexandrucosminmihai - thanks for making the changes! I've made another review for the testing code, will resolve this comment so that we can continue with the latest review.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
mbushkov: @alexandrucosminmihai - thanks for the latest changes! I did another pass at reviewing the testing code. Will resolve this comment, so that we can continue the conversation in the latest round of comments.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
alexandrucosminmihai: We discussed about this and the answer was that the behavior is the same. I made the changes where applicable. Thank you @max-vogler and @mbushkov !

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+      .get("field_repeated")
+    )
+
+    # Put the extracted descriptions together in a dictionary for testing.
+    open_api_repeated_field_schemas = {
+      "GET /metadata_test/method4": get_method4_repeated_field_schema,
+      "POST /metadata_test/method4": post_method4_repeated_field_schema,
+    }
+
+    self.assertDictEqual(expected_repeated_field_schemas,
```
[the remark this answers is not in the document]
max-vogler: FYI, there is typically no need to use type-equality functions (`assertDictEqual`) directly. Almost always, `assertEqual` should be used, which delegates to the proper function. See https://docs.python.org/3/library/unittest.html#unittest.TestCase.addTypeEqualityFunc

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+      "/metadata_test/method2",
+      "/metadata_test/method3",
+      "/metadata_test/method4",
+      "/metadata_test/method5",
+      "/metadata_test/method6",
+    }
+    open_api_paths = set(self.open_api_desc_dict["paths"].keys())
+    self.assertSetEqual(expected_paths, open_api_paths)
+
+    # Check that there are no extra/missing HTTP methods for each path.
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you for the clear explanation! It does make sense. I made the changes.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+
+      # Method6TypeReferences routes.
+      "GET /metadata_test/method6": {
+        "path": get_method6_params_path,
+        "query": get_method6_params_query,
+        "body": get_method6_params_body
+      },
+    }
+
+    self.assertDictEqual(expected_separation, open_api_separation)
```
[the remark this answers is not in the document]
alexandrucosminmihai: Done. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+    }
+    self.assertIsNone(get_method6_dict.get("requestBody"))
+    get_method6_params_body = set()
+
+    # Put all the extracted parameters in a dictionary to compare with the
+    # expected_separation dictionary.
+    open_api_separation = {
+      # Method1WithArgsType routes.
+      "GET /metadata_test/method1/{metadata_id}": {
+        "path": get_method1_params_path,
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you very much for the insight! I made the change.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+
+      # Method3PrimitiveTypes route and its response.
+      "GET /metadata_test/method3": method3_path_dict["get"]["responses"],
+
+      # Method4RepeatedField routes and their responses.
+      "GET /metadata_test/method4": method4_path_dict["get"]["responses"],
+      "POST /metadata_test/method4": method4_path_dict["post"]["responses"],
+    }
+
+    self.assertDictEqual(expected_responses, open_api_responses)
```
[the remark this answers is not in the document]
alexandrucosminmihai: Made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+    open_api_primitives["BinaryStream"] = (
+      operation_obj
+        .get("responses")
+        .get("200")
+        .get("content")
+        .get("application/octet-stream")
+        .get("schema")
+    )
+
+    self.assertDictEqual(expected_primitives, open_api_primitives)
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you for the detailed comment! I made the changes.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you very much for your comments! I made the changes.

grr/server/grr_response_server/gui/api_plugins/metadata_test.py
```diff
@@ -0,0 +1,756 @@
...
+    }
+
+    open_api_enum_type_schema = (
+      self.open_api_desc_dict
+      .get("components")
+      .get("schemas")
+      .get("grr.MetadataEnumFieldMessage.metadata_enum")
+    )
+
+    self.assertDictEqual(expected_enum_type_schema, open_api_enum_type_schema)
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you! Changed it.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -1,14 +1,112 @@
...
+             Dict[str, Union[SchemaReference, PrimitiveSchema, ArraySchema]]]
+]
+Schema = Union[PrimitiveSchema, EnumSchema, MessageSchema, ArraySchema]
+PrimitiveDescription = Dict[str, Union[str, PrimitiveSchema]]
+
+# Follows https://developers.google.com/protocol-buffers/docs/proto3#json as a
+# base, but whenever https://swagger.io/specification/#data-types provides a
+# more specific description of the same type, the OpenAPI Specification
+# version is used (noted by "OAS (type, format)" comments), with the exception
+# of int64 which uses "string" as a type.
```
panhania, line 38: Instead of putting links inline, I recommend using reference style:

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit [1], sed do eiusmod 
tempor incididunt ut labore et dolore magna aliqua [2].

[1]: http://google.com/
[2]: http://github.com/
```

Also, `int64` seems to be unquoted, whereas `string` is quoted but both refer to type names. I would use Markdown backticks for them (and in general, I tend to use Markdown for all code comments).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    # argument's type.
+    self.router: Any = router
+    # The main OpenAPI description object.
+    self.openapi_obj: Optional[Dict[str, Union[str, List, Dict]]] = None
+    self.openapi_obj_json: Optional[str] = None
+    self.schema_objs: Optional[Dict[str, Schema]] = None
+
+  def _NormalizePathComponent(self, component: str) -> str:
+    """Normalize the given path component to be used in a valid OpenAPI path."""
+    if len(component) > 0 and component[0] == '<' and component[-1] == '>':
```
panhania, line 164: I think `component.startswith("<") and component.endswith(">")` reads more naturally. Also, use double quotes for string literals (I suspect that you used single ones because of single characters, but the unfortunate reality is that there is no `char` type in Python).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      if len(component) > 0 and component[0] == '<' and component[-1] == '>':
+        normalized_component = self._NormalizePathComponent(component)
+        normalized_component = normalized_component[1:-1]
+        path_args.append(normalized_component)
+
+    return path_args
+
+  def _GetTypeName(
+      self,
+      cls: Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
```
panhania, line 208: Missing trailing comma.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    if self.schema_objs is None:  # Check required by mypy.
+      raise AssertionError(
+        "The container of OpenAPI type schemas is not initialized."
+      )
+
+    primitive_types_schemas = {
+      cast(str, primitive_type["name"]):
+        cast(Dict[str, str], primitive_type["schema"])
+      for primitive_type in primitive_types.values()
+    }
```
panhania, line 244: I would move the `cast`-ed values to a separate variables. Right now this dict comprehension is fairly loaded and hard to read (and it's not even obvious that it is a dict comprehension).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
       args: None,
       token: Optional[access_control.ACLToken] = None,
   ) -> ApiGetOpenApiDescriptionResult:
-    """Handles requests for getting the OpenAPI description of the GRR API."""
-
+    """Handle requests for the OpenAPI description of the GRR API."""
     result = ApiGetOpenApiDescriptionResult()
-    result.placeholder = "Work in progress."
 
+    if self.openapi_obj_json:
```
panhania, line 748: Since the type of this is `Optional` prefer to explicitly use `is not None`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      result.openapi_description = self.openapi_obj_json
+      return result
+
+    self.openapi_obj = {
+      "openapi": self._GetOpenApiVersion(),
+      "info": self._GetInfo(),
+      "servers": self._GetServers(),
+      "components": self._GetComponents(),
+      "paths": self._GetPaths(),
+    }
```
panhania, line 758: Any reason to make this an instance variable? After you serialize it to JSON, you don't seem to use it anymore.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+      router_method = router_methods[router_method_name]
+      for http_method, path, strip_root_types in router_method.http_methods:
+        normalized_path = self._NormalizePath(path)
+        path_args = set(self._GetPathArgsFromPath(path))
+
+        if normalized_path not in paths_obj:
+          paths_obj[normalized_path] = dict()
+
+        # The Path Object associated with the current path.
+        path_obj = paths_obj[normalized_path]
```
panhania, line 733: `path_obj = paths_obj.setdefault(normalized_path, dict())`, or even better: just use `paths_obj = collections.defaultdict(dict)`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you, it is more readable now. I used Markdown backticks for all references to code / OpenAPI Specification symbols in comments.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -51,14 +149,615 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
...
+    if isinstance(cls, EnumDescriptor):
+      return cls.full_name
+
+    if isinstance(cls, type):
+      return cls.__name__
+
+    if isinstance(cls, int):  # It's a protobuf.Descriptor.type value.
+      return cast(str, primitive_types[cls]["name"])
+
+    return str(cls)  # Cover "BinaryStream" and None.
```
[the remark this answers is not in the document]
alexandrucosminmihai: Thank you! I made the change.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for pointing this out! I changed all `'`s to `"`s where possible. Probably my subconscious was fixed on C/C++ chars and on SQL syntax :satisfied:.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added it, thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you, it was hard to read indeed! I casted the entire `primitive_types_schemas` when using it to update the `self.schema_objs` dictionary. I am not sure if this is what you meant, but please let me know if you know a cleaner way to do the casts.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: You're right, I converted it to a local variable. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Ooh, that's awesome! Thank you for the suggestion, made the change!

PR #814, opened 2020-07-29
Add client details page basic implementation

danielberbece:
A first implementation of the client details page. Only the current values of the client are displayed in this version. There are no timeline features for now, as they will be addressed in a later pull request.

To get to the client details page, I added a button in the client overview which will reroute to `v2/clients/:id/details` link, which will show the client details component.


grr/server/grr_response_server/gui/ui/components/approval/approval_test.ts
```diff
@@ -19,6 +19,11 @@ function makeClient(args: Partial<Client> = {}): Client {
     clientId: 'C.1234',
     fleetspeakEnabled: true,
     knowledgeBase: {},
+    users: [],
```
max-vogler, line 22: (optional) It's not part of your logic, but you'll get bonus points for reusing `newClient()` from `model_test_util.ts` here.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
```diff
@@ -0,0 +1,241 @@
...
+      <tr *ngIf="client.osInfo.installDate">
+        <td class="description">Install date</td>
+        <td class="value"><timestamp [date]="client.osInfo.installDate"></timestamp></td>
+      </tr>
+    </table>
+
+    <br>
+    <div class="section-title mat-h3">Users</div>
+    <table class="section">
+      <ng-template ngFor let-user [ngForOf]="client.users" let-i="index">
```
max-vogler, line 79: What's the benefit of this syntax over `*ngFor="let user of client.users; index as i"`?

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
```diff
@@ -0,0 +1,241 @@
+<div *ngIf="client$ | async as client" class="info">
+  <div class="header mat-body-1">
+    <mat-icon>computer</mat-icon>
+    <div class="mat-h1">{{client.knowledgeBase.fqdn}}
+      <div class="client-id mat-h3">{{client.clientId}}</div>
+    </div>
+  </div>
+  <div>
+    <div class="section-title mat-h3">General</div>
+    <table class="section">
```
max-vogler, line 10: Using `<table>` is pretty infamous for non-tabular data nowadays. I haven't found a best practice on displaying lists of key-value data in Material Design though, so the table usage is good for now!

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
```diff
@@ -0,0 +1,241 @@
...
+        </tr>
+        <tr>
+          <td class="description">IP addresses</td>
+          <td class="value">
+            <div *ngFor="let address of interface.addresses">
+              {{address.ipAddress}}
+            </div>
+          </td>
+        </tr>
+        <br *ngIf="i !== client.networkInterfaces.length - 1">
```
max-vogler, line 146: Use `last` from https://angular.io/api/common/NgForOf#local-variables instead of doing maths yourself.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
```diff
@@ -0,0 +1,241 @@
...
+      </ng-template>
+      <div class="empty-section" *ngIf="client.networkInterfaces.length === 0">Oops, this client seems to have no network interfaces.</div>
```
max-vogler, line 148: Although unexpected, I don't think the empty list of network interfaces should be introduced with "Oops". Instead, use a short placeholder like "(None)".

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size.ts
```diff
@@ -0,0 +1,34 @@
...
+    while (size >= 1024 && i < HumanReadableSizeComponent.UNITS.length - 1) {
+      size /= 1024;
+      i++;
+    }
+
+    if (i == 0) {
+      decimals = 0;
+    }
+
+    return size.toFixed(decimals) + ' ' + HumanReadableSizeComponent.UNITS[i];
```
max-vogler, line 32: Use format strings over string concatenation.

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size.ng.html
```diff
@@ -0,0 +1 @@
+<span>{{getHumanSize()}}</span>
```
max-vogler, line 1: It would be cool to show the exact number of bytes (possibly on hover or behind the human readable size) in a future PR. Analysts might want to copy the exact number of bytes.

max-vogler, line 1: (optional) Existing code that renders byte values (check out the form when selecting the flow "Collect single file" which comes from grr/server/grr_response_server/gui/ui/components/flow_args_form/collect_single_file_form.ng.html) links the unit to Wikipedia. You don't need to follow the same pattern, but I thought it's cool to clarify what `MiB` and similar mean.

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Shows a size in a human readable format.
+ */
+@Component({
+  selector: 'human-readable-size',
+  templateUrl: './human_readable_size.ng.html',
+})
+export class HumanReadableSizeComponent {
+  @Input() size?: number;
+  private static readonly UNITS = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB'];
```
max-vogler, line 12: You get extra style points for using base-1024 units!

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size_test.ts
```diff
@@ -0,0 +1,78 @@
...
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText).toEqual('1.0 TiB');
+
+    component.size = Math.pow(1024, 5);
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText).toEqual('1.0 PiB');
+
+    component.size = Math.pow(1024, 6) * 324;
+    fixture.detectChanges();
+    expect(fixture.nativeElement.innerText).toEqual('331776.0 PiB');
```
max-vogler, line 76: Please test the conversion more thoroughly, e.g. look at server/grr_response_server/gui/ui/components/form/byte_input/byte_conversion_test.ts. Currently, upper bounds of units (e.g. 1023.9 GiB) are untested. Also, the truncation of decimals is untested (e.g. how is 1/3 displayed?).

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size.ng.html
max-vogler, line 1: Please indicate that the human readable size is not exact, e.g. by prefixing it with `~`. 1025 B would be displayed as `1.0 KiB` which might be unexpected.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -10,22 +10,115 @@
...
+
+/** Api RdfValueDescriptor proto mapping. */
+export declare interface RdfValueDescriptor {
+  readonly name?: string;
+  readonly doc?: string;
+  readonly fields?: ReadonlyArray<RdfValueFieldDescriptor>;
+}
+
+/** ApiAllRdfValues proto mapping. */
+export declare interface ApiAllRdfValues {
```
max-vogler, line 31: The reflection stuff (ApiAllRdfValues, RdfValueDescriptor, RdfValueFieldDescriptor) is no longer used by the client details page right? If so, please remove it.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -10,22 +10,115 @@
...
+ * ApiClientInformation proto mapping.
+ */
+export declare interface ApiClientInformation {
+  readonly clientName?: string;
+  readonly clientVersion?: number;
+  readonly revision?: number;
+  readonly buildTime?: string;
+  readonly clientBinaryName?: string;
+  readonly clientDescription?: string;
+  readonly labels?: string[];
```
max-vogler, line 94: Please use immutable arrays in the type definitions (also below): `ReadonlyArray<string>`.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -10,22 +10,115 @@
...
+  readonly mountPoint?: string;
+  readonly options?: string;
+}
+
+/** ApiVolume proto mapping. */
+export declare interface ApiVolume {
+  readonly name?: string;
+  readonly devicePath?: string;
+  readonly fileSystemType?: string;
+  readonly totalAllocationUnits?: number;
```
max-vogler, line 115: Surprise: Protobuf fields that are `uint64` are actually rendered as string in JSON. You can check in proto/grr_response_proto/sysinfo.proto and the other Protobuf definition files which fields have what datatype. In the type definition, use `DecimalString` as type for these fields. In the API->Model translation layer, please parse the numbers (preferrably as `BigInt` to maximize precision for numbers larger than 9P.

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
```diff
@@ -30,6 +32,124 @@ function createClientLabel(label: ApiClientLabel): ClientLabel {
...
+}
+
+function createNetworkInterface(apiInterface: ApiInterface): NetworkInterface {
+  if (!apiInterface.macAddress) throw new Error('macAddress attribute is missing.');
+  if (!apiInterface.ifname) throw new Error('ifname attribute is missing.');
+
+  return {
+    macAddress: createMacAddress(decodeBase64(apiInterface.macAddress)),
+    interfaceName: apiInterface.ifname,
+    addresses: (apiInterface.addresses || []).map(createNetworkAddress),
```
max-vogler, line 97: Use `??` instead of `||` when right-hand side should be a replacement for null-ish values (here and many more occurences).

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
```diff
@@ -30,6 +32,124 @@ function createClientLabel(label: ApiClientLabel): ClientLabel {
...
+          apiVolume.actualAvailableAllocationUnits;
+    }
+  }
+
+  return {
+    name: apiVolume.name,
+    devicePath: apiVolume.devicePath,
+    fileSystemType: apiVolume.fileSystemType,
+    bytesPerSector: apiVolume.bytesPerSector,
+    totalSize: totalSize,
```
max-vogler, line 145: `totalSize` and `freeSpace` are `0` if `bytesPerSector` or `sectorsPerAllocationUnit` are undefined. Is this intended?

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
```diff
@@ -71,3 +71,72 @@ export function createUnknownObject(anyObject?: AnyObject): unknown|undefined {
   delete result['@type'];
   return result;
 }
+
+/**
+ * Decodes a encoded base64 string into a byte array.
+ * On error, returns an empty array.
```
max-vogler, line 77: Why is an empty array returned on error? This sort of masking makes actual issues harder to track and should only be used if reasons are well-documented.

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
```diff
@@ -71,3 +71,72 @@ export function createUnknownObject(anyObject?: AnyObject): unknown|undefined {
...
+  for (let i = 0; i < decodedString.length; i++) {
+    byteArray[i] = decodedString.charCodeAt(i);
+  }
+
+  return byteArray;
+}
+
+/** Returns the uppercase hex representation of the first byte of the provided number */
+export function byteToHex(number: number): string {
+  number = number & 0x0FF;
```
max-vogler, line 100: Omit trailing zero and use `0xFF`.

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
```diff
@@ -71,3 +71,72 @@ export function createUnknownObject(anyObject?: AnyObject): unknown|undefined {
...
+  }
+  const byteArray = new Uint8Array(decodedString.length);
+  for (let i = 0; i < decodedString.length; i++) {
+    byteArray[i] = decodedString.charCodeAt(i);
+  }
+
+  return byteArray;
+}
+
+/** Returns the uppercase hex representation of the first byte of the provided number */
```
max-vogler, line 98: Here's a trap: `number & 0xFF` zeroes out everything except the least-significant byte. The functions documentation says the "first byte" is used and the function name does not specify which byte at all.

One solution could be to write a function that shows the hexadecimal version of the whole input and remove the padding logic from the function. This could simplify the code, because three logic pieces (truncating number, hex-conversion, and padding) are split.

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive_test.ts
```diff
@@ -70,6 +70,65 @@ describe('createOptionalDateSeconds', () => {
...
+
+describe('createMacAddress', () => {
+  it('creates an empty string on non 6 bytes array input', () => {
+    expect(createMacAddress(new Uint8Array())).toEqual('');
+    expect(createMacAddress(new Uint8Array([1, 2, 3, 4, 5,]))).toEqual('');
+    expect(createMacAddress(new Uint8Array([1, 2, 3, 4, 5, 6, 7]))).toEqual('');
+  });
+  it('creates the MAC address representation of a 6 bytes array', () => {
+    expect(createMacAddress(new Uint8Array([1, 2, 3, 4, 5, 6]))).toEqual('01:02:03:04:05:06');
+    expect(createMacAddress(new Uint8Array([10, 11, 12, 13, 14, 15]))).toEqual('0A:0B:0C:0D:0E:0F');
```
max-vogler, line 132: Please add more test cases, especially how the mad and ip code handle intermediate blocks of zeroes and minimum and maximum values.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
danielberbece: This is the syntax when using `ng-template`, according to the docs: https://angular.io/api/common/NgForOf#description

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
danielberbece: Yes, would be better to initialize them with `undefined` and not show them in the UI if they are not available?

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ng.html
danielberbece: I first looked into Material Design ways to do this, but there was nothing that would fit my needs, so I went old school with this one.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
danielberbece: Why are they rendered as string? Is this Protobuf specific or does this happen in our implementation only?

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
max-vogler: Yes.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
max-vogler: Both integers and floating point numbers are represented by the [Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) type in JavaScript. When integers get too big, JavaScript stops representing them with perfect precision and silently uses floating point numbers instead. Only integers up to Number.MAX_SAFE_INTEGER (and Number.MIN_SAFE_INTEGER) are guaranteed to be stored without loss in accuracy.

MAX_SAFE_INTEGER is `2^53 - 1` (~9P) which is less than the max uint64 (`2^64 - 1`). That's why Protobuf serializes certain numeric types as JSON string instead of number. See: https://developers.google.com/protocol-buffers/docs/proto3#json

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
danielberbece: I'll update the function name and the documentation to better represent the implementation

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
danielberbece: I don't see this always as an issue, because also throwing errors can result in unexpected behaviours (remember ClientLabel?). One other way is to return undefined on error, as this will let the user expect a different output on error.

max-vogler: The unexpected behavior with ClientLabels was that ClientLabels did not show up at all. This was caused by an inconsistency in our API: Some API endpoints populated the `owner` field and some did not. The code that converts from Protobuf-JSON (ApiClientLabel) to the Frontend's internal data structure (ClientLabel) rightfully threw an Exception, which led us to discover this inconsistency in the first place.

`decodeBase64` is a library function that could be re-used in many places. If you return a special value on error (empty array or undefined), all users of your library function need to check if the returned value is the special error value on every use. If they don't, their code will likely fail (e.g. the IP address parser expects the array to have a certain length and fails) with a hard-to-debug Exception ("undefined is not a function" anyone?). Then, as a developer, it can be really hard to trace the root cause.

Returning special values that need to be checked for has similarities to golang's error handling and Java's checked exceptions. It incurs a lot of boilerplate code. Errors during parsing of base64 are not expected, but should only occur in *exceptional* circumstances, thus throwing an Exception is the right thing to do.

grr/server/grr_response_server/gui/ui/components/human_readable_size/human_readable_size.ts
```diff
@@ -28,7 +32,9 @@ export class HumanReadableSizeComponent {
     if (i == 0) {
       decimals = 0;
     }
+    const fixed = Math.pow(10, decimals);
 
-    return size.toFixed(decimals) + ' ' + HumanReadableSizeComponent.UNITS[i];
+    return `${(Math.floor(size * fixed) / fixed).toFixed(decimals)} ${
+        HumanReadableSizeComponent.UNITS[i]}`;
```
max-vogler: Please add a comment why you choose to do this math (truncating instead of rounding) and what the math does.

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
```diff
@@ -122,26 +126,32 @@ function createOptionalUnixVolume(volume?: ApiUnixVolume): UnixVolume | undefine
...
-      freeSpace = apiVolume.bytesPerSector *
-          apiVolume.sectorsPerAllocationUnit *
-          apiVolume.actualAvailableAllocationUnits;
+      freeSpace = BigInt(apiVolume.bytesPerSector) *
+          BigInt(apiVolume.sectorsPerAllocationUnit) *
+          BigInt(apiVolume.actualAvailableAllocationUnits);
     }
   }
 
+  if (apiVolume.bytesPerSector) {
```
max-vogler, line 146: Implicit testing for truthyness with `if (variable)` can lead to unexpected results. As an example, the `if` branch is executed for `"0"`, but not for `""` or `0`. Please use `===` instead.

See: https://dorey.github.io/JavaScript-Equality-Table/

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive.ts
```diff
@@ -85,7 +84,7 @@ export function decodeBase64(encodedString?: string): Uint8Array {
   try {
     decodedString = atob(encodedString);
   } catch (err) {
-    return new Uint8Array();
+    throw new Error('Invalid base64 encoding.');
```
max-vogler, line 87: `err` might contain important information for developers. This code swallows `err` and raises a new Error with no information to debug this. Instead, I suggest removing the try-catch block to let the original error bubble up.

grr/server/grr_response_server/gui/ui/lib/api_translation/client.ts
danielberbece: I'll check explicitly for undefined values

grr/server/grr_response_server/gui/ui/lib/api_translation/primitive_test.ts
```diff
@@ -81,7 +81,7 @@ describe('decodeBase64', () => {
 
   it('throws an error on invalid input', () => {
     expect(() => decodeBase64('Inv@lid $tring'))
-        .toThrow(new Error('Invalid base64 encoding.'));
+        .toThrowError('Failed to execute \'atob\' on \'Window\': The string to be decoded is not correctly encoded.');
```
max-vogler, line 84: This is a browser implementation detail that we likely do not depend upon: There is no downside for us, if the error message is not exactly the same string. If we test in a different browser or the browser changes this error message with a version update, the test breaks. Since you do not alter the thrown Error, it is fine here to only expect any Error to be thrown.

PR #819, opened 2020-08-02
configparser import problem in 2.7

Fadavvi:
Dynamic load of configparser library for different version of Python


Fadavvi:
@googlebot I signed it!

mbushkov:
@Fadavvi - thanks for the PR, but unfortunately we don't support Python 2.7 anymore. GRR runs on Python 3.6 and higher.

PR #823, opened 2020-08-13
Client details timeline & lists accordion button

danielberbece:
This PR contains changes for the following features:

- Timeline feature, that allows the selection of a specific version of the client for which the details are showed. The snapshots of the client are fetched from the API endpoint `api/v2/clients/{id}/versions`, and then are processed in order to get only the client versions with real differences. The differences between the snapshots are a result of applying `deep-diff` algorithm only on the properties that are showed in the page. Then, using the `deep-diff` result, the timeline entry descriptions are made. The description aggregates changes of the same entry (e.g. 3 new network interfaces) into one line. The description may contain up to 3 different changes, and if there are more changes, a general description will be shown (e.g. "N new changes").
- Show more/less button for lists. In the case of users, network interfaces and volumes being lists, we decided to show by default only a few, with the ability to show more/less items on a click of a button. The button may have 3 states: `no-button` state, in which the button shouldn't show at all because the number of items in the list is less than or equal to the maximum initial number of items showed. `show-more` state, which happens when there are more items available to be shown than the number of items currently displayed, and `show-less`, when there are no more items to show, but the list is bigger than the initial number of items showed.


grr/server/grr_response_server/gui/ui/components/client_details/client_details.ts
```diff
@@ -33,6 +102,116 @@ export class ClientDetails implements OnInit, OnDestroy {
...
+  /**
+   * Returns the number of entries changed and a list of the aggregated changes
+   * descriptions (e.g. "2 users added")
+   */
+  static getDiffDescriptions(diff?: Diff<Client>[]): [number, string[]] {
+    if (diff === undefined) {
+      return [0, []];
+    }
+
+    let added: Map<string, number> = new Map();
```
max-vogler, line 125: Use const instead of let if possible.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ts
```diff
@@ -33,6 +102,116 @@ export class ClientDetails implements OnInit, OnDestroy {
...
+    let deleted: Map<string, number> = new Map();
+    let updated: Map<string, number> = new Map();
+    let diffDescriptions = [] as string[];
+    let numChanges = 0;
+
+    const classifyDiffItem = function(element: Diff<Client>) {
```
max-vogler, line 131: use named function `function classifyDiffItem(...)` instead of assigning an unnamed function to a `const`. Also, do not use unnamed functions, but lambdas instead, because lambdas have cleaner scoping of `this`.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ts
```diff
@@ -33,6 +102,116 @@ export class ClientDetails implements OnInit, OnDestroy {
...
+      return 'no-button';
+    }
+    return 'show-less';
+  }
+
+  /**
+   * Returns the number of entries changed and a list of the aggregated changes
+   * descriptions (e.g. "2 users added")
+   */
+  static getDiffDescriptions(diff?: Diff<Client>[]): [number, string[]] {
```
max-vogler, line 120: Please extract the diffing code out of the client details component into a standalone file.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.ts
```diff
@@ -33,6 +102,116 @@ export class ClientDetails implements OnInit, OnDestroy {
...
+        default:
+          updated.set(label, (updated.get(label) ?? 0) + 1);
+      }
+    };
+
+    diff.forEach(classifyDiffItem);
+
+    const addToChanges = function(
+        classMap: Map<string, number>, changeKeyword: string) {
+      classMap.forEach((occurances, label) => {
```
max-vogler, line 159: Typo: occurences.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -163,6 +163,26 @@ export declare interface ApiClient {
...
+ */
+export declare interface ApiGetClientVersionsResult {
+  readonly items: ReadonlyArray<ApiClient>;
+}
+
+/**
+ * ApiGetClientVersionsArgs proto mapping.
+ */
+export declare interface ApiGetClientVersionsArgs {
+  readonly clientId: string;
```
max-vogler, line 181: Missing `?`.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -163,6 +163,26 @@ export declare interface ApiClient {
...
   readonly labels?: ReadonlyArray<ApiClientLabel>;
+
+  readonly age?: string;
+}
+
+/**
+ * ApiGetClientVersionsResult proto mapping.
+ */
+export declare interface ApiGetClientVersionsResult {
+  readonly items: ReadonlyArray<ApiClient>;
```
max-vogler, line 174: Missing `?`.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -194,6 +204,14 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
       skip(1),
   );
 
+  readonly selectedClientVersions$: Observable<Client[]> = of(undefined).pipe(
+      tap(() => this.fetchSelectedClientVersions()),
```
max-vogler, line 208: Wrap lambda body in curly braces to show that nothing is returned.

package-lock.json
```diff
@@ -0,0 +1,11 @@
+{
```
max-vogler, line 1: Please delete this file.

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
danielberbece: I decided to not use `?` as the `clientId` is a must have argument in order to get the client versions. Without it, we wouldn't be able to fetch the client versions, so it would be an incorrect use.

grr/server/grr_response_server/gui/ui/components/client_details/client_details.scss
```diff
@@ -1,56 +1,114 @@
...
+    .heading {
+      font-family: c.$google-sans-family;
+      font-size: 18px;
+      line-height: 24px;
+      padding-left: 16px;
     }
 
-    .fqdn {
+    .since {
+      font-family: Roboto;
```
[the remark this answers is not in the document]
danielberbece: I don't see any equivalent typography class to match the design specifications. Do you have any suggestions on how to achieve the same result?

max-vogler:
In future PRs, please:
* Introduce `client_details_page_facade` similar to `client_page_facade`, that maintains the state of everything client details related.
* Introduce an expandable accordion component, or [structural directive](https://angular.io/guide/structural-directives) to remove this state from the client details and make the functionality reusable.

PR #824, opened 2020-08-14
Use references to primitive types in OpenAPI description of the GRR API

alexandrucosminmihai:
In this PR, logic is added to define the primitive types only once, in the `components` field of the root [`OpenAPI Object`](https://swagger.io/specification/#openapi-object), then use `Reference Object`s to refer to them, just like we do with composite types. Previously, primitive types would have their schema added whenever they were used (for example for describing route parameters).

This change reduces the complexity of the code, as the need to differentiate between composite types and primitive types throughout the description generation logic is reduced and it also can reduce the size of the generated JSON (depending on how many times primitive types are used).

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -339,25 +339,25 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       type_name: str,
       is_array: bool = False
   ) -> Union[SchemaReference, ArraySchema]:
-    """Get a Reference Object that points to a schema definition.
+    """Get a `Reference Object` that points to a schema definition.
 
     All types (including protobuf primitives) are expected to have been
-    previously defined in the `components` field of the root OpenAPI object
+    previously defined in the `components` field of the root `OpenAPI Object`
     and are used via OpenAPI references.
 
     Args:
-      type_name: The name of the type for which an OpenAPI reference object will
-        be created and returned.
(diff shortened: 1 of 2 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -340,63 +334,46 @@ def _CreateSchemas(self) -> None:
...
-        OpenAPI schema definition of the selected type.
+      If the `is_array` argument is set to `False`, then an OpenAPI `Reference
+      Object` representing the path to the actual OpenAPI schema definition of
+      the selected type.
       If the `is_array` argument is set to `True`, then an OpenAPI array schema
-      is constructed that uses for its `items` field:
-      - if the type is primitive, the actual OpenAPI schema of the type;
-      - else, an OpenAPI `Reference Object` associated with the type's schema.
+      is constructed that uses for the `items` field an OpenAPI `Reference
+      Object` associated with the type's schema.
```
panhania, line 360: Something for the future PR: this should be probably split into two method.

> Boolean arguments loudly declare that the function does more than one thing. They are confusing and should be eliminated.
>
> — "Clean Code"

This will also allow you to avoid this nasty `Union` in the return type.

panhania (approved):

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for pointing this out! I'll make the change to 2 methods.

PR #826, opened 2020-08-17
Add removal of client label

danielberbece:
Added a feature to let an user remove an already assigned label of a client. The feature consists of an `X` button inside the client label chip. Upon click, a POST request is sent to the endpoint `/api/v2/clients/labels/remove` with the label name and the clientId of the selected client. After the POST request, the client is re-fetched and the UI is automatically updated in order to reflect the change (be it successful or not).


grr/server/grr_response_server/gui/ui/components/client/client.ts
```diff
@@ -49,6 +49,10 @@ export class Client implements OnInit, OnDestroy {
     });
   }
 
+  removeLabel(labelName: string) {
+    this.clientPageFacade.removeClientLabel(labelName);
+  }
```
panhania, line 54: Nit: the argument name in `addLabel` (and pretty much everything else) is `label`, but here it is `labelName`—looks a bit inconsistent.

max-vogler:
This looks like a good PR for @panhania to review! I suggest displaying a snackbar with an undo button after a label has been removed, see [Material Design: Confirmation & acknowledgement](https://material.io/design/communication/confirmation-acknowledgement.html#acknowledgement).

grr/server/grr_response_server/gui/ui/components/client/client.ts
danielberbece: True, will change to `label`

danielberbece:
I addressed the issue and added the suggested feature: a snackbar with the message `Label <name> removed` and action `UNDO` when the request to remove the label succeeds.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -491,7 +503,13 @@ export class ClientPageFacade {
...
 
+  /** Adds a label to the selected client */
   addClientLabel(label: string) {
     this.store.addClientLabel(label);
   }
+
+  /** Removes a label from the selected client */
+  removeClientLabelReq(label: string): Observable<HttpResponse<{}>> {
+    return this.store.removeClientLabel(label);
+  }
```
panhania, line 514: I find it a bit awkward that we have a `addClientLabel` method (that returns nothing) and a `removeClientLabelReq` method (that returns an observable). Is this inconsistency necessary?

danielberbece: Totally agree here, but in order to show the `UNDO` snackbar we needed to know if the remove request is successful or not. If there is a better way to achieve this behaviour, I am open to suggestions.

panhania: @max-vogler should have some suggestions.

grr/server/grr_response_server/gui/ui/components/client/client.ng.html
```diff
@@ -7,8 +7,8 @@ <h1 class="mat-h1">{{client.knowledgeBase.fqdn}}</h1>
       <div class="chips">
         <mat-chip-list>
           <online-chip [lastSeen]="client.lastSeenAt"></online-chip>
-          <mat-chip *ngFor="let label of client.labels">
-            {{label.name}}
+          <mat-chip *ngFor="let label of client.labels" (removed)="removeLabel(label.name)">
```
max-vogler, line 10: Please add `trackBy` and track elements by label name. Two objects are never `===` unless they are the same instance, so Angular will recreate the DOM upon every change even though labels technically stay the same.

grr/server/grr_response_server/gui/ui/lib/api/http_api_service.ts
```diff
@@ -223,6 +223,14 @@ export class HttpApiService {
         url, {client_ids: [clientId], labels: [label]});
   }
 
+  removeClientLabel(clientId: string, label: string): Observable<HttpResponse<{}>> {
```
max-vogler, line 226: HttpApiResponse is an implementation detail and should not be returned. Instead, return {} or void.

grr/server/grr_response_server/gui/ui/lib/api/http_api_service.ts
```diff
@@ -223,6 +223,14 @@ export class HttpApiService {
...
+    const url = `${URL_PREFIX}/clients/labels/remove`;
+    return this.http.post<{}>(
+        url, {client_ids: [clientId], labels: [label]}, {
+          observe: 'response',
```
max-vogler, line 230: Im on the run right now, cannot look it up: What does observe response do?

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -428,6 +429,17 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
                   this.httpApiService.addClientLabel(clientId, label)),
           tap(() => this.fetchClient()),
           ));
+
+  // An effect to remove a label from the selected client
+  removeClientLabel(label: string): Observable<HttpResponse<{}>> {
```
max-vogler, line 434: Functions that trigger change should only return void in NgRx (our store implementation). Instead of returning an observable here, create an observable as member of the Store. Then, when deletion of the label is completed, update the state which should then trigger emission of the dedicated observable. E.g.: the observable that tracks flow creation state.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -491,7 +503,13 @@ export class ClientPageFacade {
     this.store.stopFlowConfiguration();
   }
 
+  /** Adds a label to the selected client */
   addClientLabel(label: string) {
     this.store.addClientLabel(label);
   }
+
+  /** Removes a label from the selected client */
+  removeClientLabelReq(label: string): Observable<HttpResponse<{}>> {
```
max-vogler, line 512: `Req` sounds like an implementation detail. Please name it `removeClientLabel`.

max-vogler (changes requested):
Please add tests for the undo function.

grr/server/grr_response_server/gui/ui/lib/api/http_api_service.ts
danielberbece: By default, the returned observable would be the data we are requesting for. In these case, I wanted to get the status, and in order to do that I had to specify `response` to be returned.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -32,6 +33,17 @@ export type StartFlowState = {
   readonly error: string,
 };
 
+export type ChangeRequestState = {
```
max-vogler, line 36: This can be simplified and `clearRemoveClientLabelState` removed:

We are only interested in the fact that a label was deleted successfully (to show the snackbar and remove it from the label list). You can introduce an observable that only emits the label name whenever the deletion was successful.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
danielberbece: This is a very good idea, thanks for the suggestion!

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -457,6 +458,27 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
...
           ));
+
+  /** An effect to remove a label from the selected client */
+  readonly removeClientLabel = this.effect<string>(
+      obs$ => obs$.pipe(
+          withLatestFrom(this.selectedClientId$),
+          switchMap(
+              ([label, clientId]) =>
+                  this.httpApiService.removeClientLabel(clientId, label)),
+          catchError(() => {
```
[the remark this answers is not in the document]
danielberbece: Removing the `catchError()` pipe step causes the store `effect` to throw the error. I'd suggest to catch the error here in order to throw it further in the `removedClientLabels$` observable

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -457,6 +458,26 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
...
+  /** An effect to remove a label from the selected client */
+  readonly removeClientLabel = this.effect<string>(
+      obs$ => obs$.pipe(
+          withLatestFrom(this.selectedClientId$),
+          switchMap(
+              ([label, clientId]) =>
+                  this.httpApiService.removeClientLabel(clientId, label)),
+          tap((label) => {
+            this.fetchClient();
+            this.removedClientLabels$.next(label);
```
max-vogler, line 471: Please loop this through the store's state: Extend `ClientPageState` with a field (e.g. lastRemovedClientLabel`), which is then updated on successful removal (in the `tap`). You can ignore the errors for now - remove `catchError`.

removedClientLabels$ should not be a `new Subject`, instead, use `this.selectt()` to read the property of the state.

Looping all state changes through one state is a typical pattern for NgRx.

max-vogler (changes requested):

max-vogler (approved):

Issue #822, opened 2020-08-12
OpenAPI doesn't support the semantics of protobuf.oneof so we can't implement it at the moment


alexandrucosminmihai:
(Note: I use `oneOf`, `anyOf`, `allOf` (camel-case) to refer to OpenAPI's keywords and `protobuf.oneof` to refer to protobuf's.)

After more thorough investigation and experimentation I figured out that at the moment there is no way to correctly describe `protobuf.oneof`’s behaviour using the OpenAPI Specification. 

The first big problem is related to route parameters descriptions in OpenAPI. OpenAPI [does not allow interdependence between parameters](https://github.com/OAI/OpenAPI-Specification/issues/256) (including mutual exclusion which is the core semantic of `protobuf.oneof`). Therefore, there is no way to describe the behavior of `protobuf.oneof` when used to describe a route parameter, because the best we could do with OpenAPI is describe a single parameter with a *single name* (`field_oneof1`, for example) that accepts `anyOf` the schemas of the fields that are part of the `protobuf.oneof`. This is not correct because the `protobuf.oneof` does not define a field with multiple possible types, but rather it defines a mutual exclusion between multiple fields, each with *their own name and type*, so this OpenAPI solution would fail to describe the API correctly (i.e. instead of specifying that a method expects either a `field_one` argument of type `int32`, or a `field_two` argument of type `string`, the OpenAPI description would specify that the API method expects an argument `field_oneof` which can be either of type `int32` or `string`). Therefore, defining the `protobuf.oneof` semantic for method arguments with OpenAPI is impossible at the moment, but what about response types?

The OpenAPI Specification offers the special operators `oneOf`, `anyOf`, `allOf` which can be used to describe a schema by combining other schemas in the following ways: 

- “this schema should respect exactly `oneOf` the following schemas”, 
- “this schema should respect at least a schema of `anyOf` of the following schemas”, 
- “this schema should respect `allOf` the the following schemas (i.e. this schema is a merge of all the following schemas)”.

[...]

PR #827, opened 2020-08-19
Generate OpenAPI descriptions for protobuf.oneof fields + other changes

alexandrucosminmihai:
In this PR I added support for generating an OpenAPI description of `protobuf.oneof` fields.

At the moment, the OpenAPI Specification does not support mutual exclusions between fields/parameters, so it does not support the semantics of `protobuf.oneof`. For a thorough examination of this situation, with references, please see the content of [issue 822](https://github.com/google/grr/issues/822).

The current approach introduced by this PR is to mark fields of the same `protobuf.oneof` with a `description` field mentioning that only one field per `oneof` should be present (either received as part of a response or sent in a request as a parameter).

In this PR I've also:
- added a new test for the `protobuf.oneof` description generation, updated the other tests that needed to consider this case and refactored the tests by introducing the `_GetParamSchema` top-level function;
- converted local variables type annotations to `var: Type` notation, from `var = expr # type: Type` notation;
- split the `_GetReferenceObject` function by moving array logic to a new `_GetArraySchema` function.


panhania:
> In this PR I've also: (...)

If you need to provide a list of things that the PR also does, it is likely that the PR does too much. By splitting the PR into small logical units, the repository history becomes more clear and the reviewers have easier job. Ideally, PR should do only what the title says, not more.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -735,3 +686,88 @@ def _GetTypeName(
...
+  """Get the schema of an array with items of the given type."""
+
+  return {
+    "type": "array",
+    "items": _GetReferenceObject(items_type_name)
+  }
+
+
+def _GetFieldSchema(
+    field_descriptor: FieldDescriptor
```
panhania, line 721: Missing trailing comma.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -735,3 +686,88 @@ def _GetTypeName(
...
+    "$ref": f"#/components/schemas/{type_name}"
+  }
+
+
+def _GetArraySchema(items_type_name: str) -> ArraySchema:
+  """Get the schema of an array with items of the given type."""
+
+  return {
+    "type": "array",
+    "items": _GetReferenceObject(items_type_name)
```
panhania, line 716: Missing trailing comma.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -735,3 +686,88 @@ def _GetTypeName(
...
+  Args:
+    type_name: The name of the type for which an OpenAPI `Reference Object`
+      will be created and returned.
+
+  Returns:
+    An OpenAPI `Reference Object` representing the path to the actual OpenAPI
+    schema definition of the selected type.
+  """
+  return {
+    "$ref": f"#/components/schemas/{type_name}"
```
panhania, line 707: Missing trailing comma (autoformatter would re-format it to ` return {"$ref": f"..."}`, trailing comma ensure that this does not happen and make diffs nicer).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -735,3 +686,88 @@ def _GetTypeName(
...
+  # OpenAPI Specification. See this GitHub issue [1] for more details.
+  #
+  # [1]: github.com/google/grr/issues/822
+  containing_oneof: OneofDescriptor = field_descriptor.containing_oneof
+  if containing_oneof is None:
+    if field_descriptor.label == protobuf2.LABEL_REPEATED:
+      return _GetArraySchema(_GetTypeName(field_descriptor))
+
+    return _GetReferenceObject(_GetTypeName(field_descriptor))
+  else:
```
panhania, line 749: I think you can remove this `else` to reduce nesting.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for your review! I have a question: should there be a trailing comma to the list of arguments for all function/method signatures or maybe only for multiline arguments lists?

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you! Removed it.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added them. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Added them. Thank you!

PR #831, opened 2020-08-25
Implement OpenAPI description generation for Protobuf maps

alexandrucosminmihai:
In this PR I introduce OpenAPI description generation for `protobuf.map` types, as well as a test that verifies the generated output. 

There is a difference between `protobuf.map` and the OpenAPI Specification descriptions of maps: `protobuf.map`s support ["any integral or string type" for the keys' type](https://developers.google.com/protocol-buffers/docs/proto#maps), whereas the OpenAPI Specification allows only [`string` for the keys' type](https://swagger.io/docs/specification/data-models/dictionaries/).

In order to let users know the actual type of the keys, I use the same approach as in the case of [`protobuf.oneof` semantics](https://github.com/google/grr/pull/827), which is to specify the actual types in a comment stored in the `description` field of the `schema` dictionary used to describe route parameters and composite types' properties, as well as in the `description` field of the actual schema of the map type. 
The parameter/properties `description`s are rendered by documentation generation tools such as [ReDoc](https://github.com/Redocly/redoc) , whereas the `description` field of the map type schema (from the `components` field of the root `OpenAPI Object`) can be helpful when reading the generated JSON as is, for debugging purposes.

Please note that I use the ":thumbsup:" emoji to keep track of comments that I've addressed in a yet to be committed/pushed commit.

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -34,6 +34,7 @@ DescribedSchema = Dict[str,
                        Union[str, List[SchemaReference], List[ArraySchema]]]
 Schema = Union[PrimitiveSchema, EnumSchema, MessageSchema, ArraySchema]
 PrimitiveDescription = Dict[str, Union[str, PrimitiveSchema]]
+TypeHinter = Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
 
 # Follows the proto3 JSON encoding [1] as a base, but whenever
 # the OpenAPI Specification [2] provides a more specific description of the
@@ -334,9 +335,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
 
   def _CreateSchema(
       self,
-      cls: Optional[
-        Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
-      ],
+      cls: Optional[TypeHinter],
       visiting: Set[str],
   ) -> None:
     """Create OpenAPI schema from any valid type descriptor or identifier."""
@@ -828,11 +827,7 @@ def _GetPathArgsFromPath(path: str) -> List[str]:
   return path_args
 
 
-def _GetTypeName(
-    cls: Optional[
-      Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
-    ],
-) -> str:
+def _GetTypeName(cls: Optional[TypeHinter]) -> str:
   """Extract type name from protobuf `Descriptor`/`type`/`int`/`str`."""
   if isinstance(cls, FieldDescriptor):
     if _IsMapField(cls):
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -301,9 +301,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     type_name: str = _GetTypeName(field_descriptor)
     visiting.add(type_name)
 
-    key_field_d, value_field_d = (
-      _GetMapFieldKeyValueTypes(field_descriptor)
-    )
+    key_field_d, value_field_d = _GetMapFieldKeyValueTypes(field_descriptor)
 
     # pylint: disable=line-too-long
     # `protobuf.map` key types can be only a subset of the primitive types [1],
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -439,7 +439,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
 
     type_name = _GetTypeName(field_descriptor)
     containing_oneof: OneofDescriptor = field_descriptor.containing_oneof
-    description: str = ""
+    description = ""
     array_schema = None
     reference_obj = None
 
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -30,8 +30,8 @@ MessageSchema = Dict[
   str, Union[str,
              Dict[str, Union[SchemaReference, ArraySchema]]]
 ]
-OneofFieldSchema = Dict[str,
-                        Union[str, List[SchemaReference], List[ArraySchema]]]
+DescribedSchema = Dict[str,
+                       Union[str, List[SchemaReference], List[ArraySchema]]]
 Schema = Union[PrimitiveSchema, EnumSchema, MessageSchema, ArraySchema]
 PrimitiveDescription = Dict[str, Union[str, PrimitiveSchema]]
 
@@ -290,7 +290,15 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     Returns:
       Nothing, the schema is stored in `self.schema_objs`.
     """
-    type_name = _GetTypeName(field_descriptor)
+    if self.schema_objs is None:  # Check required by mypy.
+      raise AssertionError(
+        "The container of OpenAPI type schemas is not initialized."
+      )
+
+    if field_descriptor is None:  # Check required by mypy.
+      raise AssertionError(f"`field_descriptor` is None.")
+
+    type_name: str = _GetTypeName(field_descriptor)
     visiting.add(type_name)
 
     key_field_d, value_field_d = (
@@ -307,19 +315,30 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     # pylint: enable=line-too-long
     self._CreateSchema(value_field_d, visiting)
 
+    if key_field_d is None:  # Check required by mypy.
+      raise AssertionError(
+        f"The map entry message type has no `key` FieldDescriptor."
+      )
+
     visiting.remove(type_name)
 
-    self.schema_objs[type_name] = {
-      "description": f"This is a map with real key type="
-                     f"\"{_GetTypeName(key_field_d)}\" and value type="
-                     f"\"{_GetTypeName(value_field_d)}\"",
-      "type": "object",
-      "additionalProperties": _GetReferenceObject(_GetTypeName(value_field_d)),
-    }
+    self.schema_objs[type_name] = cast(
+      Dict[str, Union[str, SchemaReference]],
+      {
+        "description": f"This is a map with real key type="
+                       f"\"{_GetTypeName(key_field_d)}\" and value type="
+                       f"\"{_GetTypeName(value_field_d)}\"",
+        "type": "object",
+        "additionalProperties": _GetReferenceObject(
+          _GetTypeName(value_field_d)),
+      }
+    )
 
   def _CreateSchema(
       self,
-      cls: Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str],
+      cls: Optional[
+        Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
+      ],
       visiting: Set[str],
   ) -> None:
     """Create OpenAPI schema from any valid type descriptor or identifier."""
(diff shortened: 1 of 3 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -333,6 +413,96 @@ def _CreateSchemas(self) -> None:
...
+      includes a `description` entry along the `Reference Object` or schema.
+    """
+    if self.schema_objs is None:  # Check required by mypy.
+      raise AssertionError(
+        "The container of OpenAPI type schemas is not initialized."
+      )
+
+    type_name = _GetTypeName(field_descriptor)
+    containing_oneof: OneofDescriptor = field_descriptor.containing_oneof
+    description: str = ""
```
mbushkov, line 444: Minor nit: I'd expect Mypy static type analyzer to be smart enough (hopefully) to infer that `description` is a string if it gets initialized by assigning a string literal. So, type annotation is likely not needed here.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -662,12 +831,27 @@ def _GetPathArgsFromPath(path: str) -> List[str]:
 
 
 def _GetTypeName(
-    cls: Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str],
+    cls: Optional[
+      Union[Descriptor, FieldDescriptor, EnumDescriptor, Type, int, str]
```
mbushkov, line 835: This type is used twice (here and in the `CreateSchema` method). Would it make sense to define and use an alias for it?

mbushkov (commented):
Looks good to me! (only 2 minor comments)
Will leave it to @panhania to formally approve.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -268,9 +263,82 @@ def _CreateMessageSchema(
...
+
+    self.schema_objs[type_name] = cast(
+      Dict[str, Union[str, SchemaReference]],
+      {
+        "description": f"This is a map with real key type="
+                       f"\"{_GetTypeName(key_field_d)}\" and value type="
+                       f"\"{_GetTypeName(value_field_d)}\"",
+        "type": "object",
+        "additionalProperties": _GetReferenceObject(
+          _GetTypeName(value_field_d)),
```
panhania, line 333: I would move this to a separate variable, it's hard to parse nested method calls with a line break (for a moment I thought that this is a pair).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -268,9 +263,82 @@ def _CreateMessageSchema(
...
+
+    if field_descriptor is None:  # Check required by mypy.
+      raise AssertionError(f"`field_descriptor` is None.")
+
+    type_name: str = _GetTypeName(field_descriptor)
+    visiting.add(type_name)
+
+    key_field_d, value_field_d = (
+      _GetMapFieldKeyValueTypes(field_descriptor)
+    )
```
panhania, line 306: Does it really not fit the 80-character limit?

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -268,9 +263,82 @@ def _CreateMessageSchema(
...
+        OpenAPI schemas constructed and have their associated `_Create*Schema`
+        call in the current call stack.
+
+    Returns:
+      Nothing, the schema is stored in `self.schema_objs`.
+    """
+    if self.schema_objs is None:  # Check required by mypy.
+      raise AssertionError(
+        "The container of OpenAPI type schemas is not initialized."
+      )
```
panhania, line 296: I would make it shorter to fit it the 80-character limit, e.g.: _"OpenAPI type schemas not initialized"_. In general, I believe that error messages should be rather concise (as long as we don't make them less informative), because then it is easier to follow the stacktrace (which usually is also line-wrapped at some point).

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -662,12 +831,27 @@ def _GetPathArgsFromPath(path: str) -> List[str]:
...
+      map_type_name = _GetTypeName(cls.message_type)
+      if map_type_name.endswith("Entry"):
+        map_type_name = map_type_name[:-5]
+
+      key_type_d, value_type_d = _GetMapFieldKeyValueTypes(cls)
+
+      return (
+        f"{map_type_name}Map_"
+        f"{_GetTypeName(key_type_d)}:{_GetTypeName(value_type_d)}"
+      )
```
panhania, line 850: I would put results of `_GetTypeName` and make the string literal fit one line, right now it is hard to parse.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -716,58 +900,66 @@ def _GetArraySchema(items_type_name: str) -> ArraySchema:
   }
 
 
-def _GetFieldSchema(
+def _GetMapEntryTypeName(field_name: str) -> str:
+  """Extract the name of the associated map type from a field's name."""
+  capitalized_name_components = [
+    comp.capitalize() for comp in field_name.split("_")
+  ]
```
panhania, line 907: I find `map(str.capitalize, field_name.split("_"))` to be more elegant than list comprehensions.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -716,58 +900,66 @@ def _GetArraySchema(items_type_name: str) -> ArraySchema:
...
+
+  return f"{''.join(capitalized_name_components)}Entry"
+
+
+def _GetMapFieldKeyValueTypes(
     field_descriptor: FieldDescriptor,
-) -> Union[SchemaReference, ArraySchema, OneofFieldSchema]:
-  """Create a dictionary to be used as the schema of a message field.
+) -> Tuple[Optional[FieldDescriptor], Optional[FieldDescriptor]]:
```
panhania, line 914: The type of your function is `Tuple[Optional[X], Optional[Y]]` but in reality, you only return the key if and only if you return  the value. You could make this more explicit by having the following type: `Optional[Tuple[X, Y]]`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -716,58 +900,66 @@ def _GetArraySchema(items_type_name: str) -> ArraySchema:
...
+  ):
+    return entry_descriptor.fields[1], entry_descriptor.fields[0]
+
+  return None, None
+
 
-  return oneof_member
+def _IsMapField(field_descriptor: FieldDescriptor) -> bool:
+  """Checks that a `FieldDescriptor` is of a map type."""
+  return _GetMapFieldKeyValueTypes(field_descriptor) != (None, None)
```
panhania, line 965: If you apply the previous suggestion, you can simplify this with `is not None`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for pointing this out! I believe I had a cast there at some point, then removed it and overlooked the fact that the remaining expression would fit on one line.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: You're right, thank you! Removed the unnecessary type annotation.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: It makes sense. Thank you for pointing this out! I introduced the alias named `TypeHinter` because the types in the union are either representing the type directly (such as `Descriptor`, `EnumDescriptor` and `Type)`, or identify a type through a label (such as an `int` `protobuf.TYPE_*` constant or a `str` represeting the name of the type for `BinaryStream`), or hold a type descriptor in their structure (`FieldDescriptor`) and "Hinter" is the "least generic" name I came up with. If you have a better suggestion for the alias name, please let me know.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for your suggestion! I created variables for the key and value type names, this removes the nested method calls and also avoids duplication with the code above. Please let me know if you would like me to treat it differently (such as creating a separate variable only/also for the `Reference Object`).

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for the detailed explanation! I made the change to the shorter message.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I agree, it's not very readable. I made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I agree. Made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: That is correct, but using `Optional[Tuple[X, Y]]` would require that the returned value of this function to be stored in a variable to check whether the value `is None` before accessing the tuple elements as required by static type checking. This would imply that either:
- we don't use sequence unpacking anymore and rather index the new variable used to store the returned tuple (or `None`);
- we continue to use sequence unpacking, but use a temporary variable only to check whether the function returned `None` or not.

Which of the 3 approaches (3rd being the one used right now) do you consider to be the best?

panhania: How about defining a `KeyValueDescriptor` which is a named tuple with `key` and `value` properties? Then you can use the first approach, where you don't unpack, but you don't have to use ugly and confusing indexing.

alexandrucosminmihai: Awesome idea! Thank you for your suggestion, I implemented it.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Made the change. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -716,58 +887,68 @@ def _GetArraySchema(items_type_name: str) -> ArraySchema:
...
+  if len(entry_descriptor.fields) != 2:
+    return None
+
+  if (
+      entry_descriptor.fields[0].name == "key" and
+      entry_descriptor.fields[1].name == "value"
+  ):
+    return KeyValueDescriptor(
+      entry_descriptor.fields[0], entry_descriptor.fields[1]
+    )
```
panhania, line 939: Prefer to use keyword arguments when initializing struct-like things. This makes the code more resilient to changes like re-ordering the fields. Here we use `NamedTuple` because data classes were introduced in Python 3.7, so we cannot use them yet.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -716,58 +887,68 @@ def _GetArraySchema(items_type_name: str) -> ArraySchema:
...
+
+  if (
+      entry_descriptor.fields[0].name == "value" and
+      entry_descriptor.fields[1].name == "key"
+  ):
+    return KeyValueDescriptor(
+      entry_descriptor.fields[1], entry_descriptor.fields[0]
+    )
```
panhania, line 947: Ditto.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for this suggestion! I made the change.

panhania (approved):

PR #832, opened 2020-08-26
Integrating GRR & Fleetspeak with Grafana - GRRafana HTTP server

tsehori:
We introduce a new component of GRR response server nicknamed **GRRafana**, that will be in charge of serving requests from [Grafana](https://grafana.com/) using the [data source plugin](https://grafana.com/docs/grafana/latest/features/datasources/#data-source-plugins) [JSON Datasource](https://github.com/simPod/grafana-json-datasource). This component will serve metrics for a Fleetspeak-enabled setup and depend on a connection to Fleetspeak over gRPC. Therefore, it will not be supported by legacy GRR deployments.

Here's a short diagram that can demonstrate the relationship of Grafana, GRR and Fleetspeak with the new component:
GRRafana
Given that this PR involves communication with Fleetspeak's database, there is a pending corresponding PR on Fleetspeak repository as well, google/fleetspeak#281. Once both PRs are merged, the HTTP server will work as expected.

With this component, we are able to achieve multiple goals:
- Grafana is completely neutral towards the underlying datastore implementation of GRR and Fleetspeak, and does not depend on it in any way. That was our main issue when we created Grafana dashboards that depend on a specific datastore implementation and used a plugin such as the built in [MySQL plugin](https://grafana.com/docs/grafana/latest/features/datasources/mysql/#using-mysql-in-grafana) for a specific datastore implementation. Furthermore, individuals and companies with proprietary database technologies can utilize this with no significant overhead other than implementing the corresponding specific methods (and not create or find a new Grafana data source plugin).
- This approach makes the querying communication testable, so we can implement several tests to make continuous integration smooth. ~~At the moment, we have unit tests for the new methods in `fleetspeak_utils.py`, but more will come for `grrafana.py`~~ All new code in both `fleetspeak_utils.py` and `grrafana.py` is added with tests in this PR.

I will briefly go over the structure of the new GRRafana HTTP server. In a similar manner to other GRR response server components, GRRafana HTTP server runs on top of [Werkzeug](https://werkzeug.palletsprojects.com/en/1.0.x/), and the main file `grrafana.py` is located in `grr/server/grr_response_server/bin` along other GRR response server components.
At the moment, GRRafana serves 3 of the 4 major [API endpoints](https://github.com/simPod/grafana-json-datasource#api) of JSON Datasource plugin (all but annotations). Here are some details about the structure and implementation of GRRafana on GRR's side (Fleetspeak's side is described in the other PR):
- The endpoints and their corresponding methods on the server are explicitly stated in the constructor.
- `GET /`: It only returns an empty message which will be read by Grafana as an OK message. The logic is, as GRRafana is only part of Fleetspeak-enabled setups, then if GRRafana is up, so should be Fleetspeak and its database.
- `POST /search`: It gets a bit more confusing. This endpoint essentially serves two different possible results to Grafana based on queries from different parts of Grafana; it is documented more explicitly within the method.
- `POST /query`: For this endpoint, we utilized an [existing database method](https://github.com/google/fleetspeak/blob/690991be00993813230a8f6c3aad703b21dfb0c5/fleetspeak/src/server/db/store.go#L233) on Fleetspeak's side which did not have a corresponding RPC method, so we've created a new RPC method called `FetchClientResourceUsageRecords` to not complicate the existing implementation. After retrieving of the records, we manipulate them on GRRafana in order to serve it to Grafana in the format it expects, as can be seen in the function `fetchDatapointsForTargets`.

Note that this PR lacks configurations to the initial setup of GRR and Fleetspeak so that they support GRRafana; these will be added in a separate PR later on.


grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+from grr_response_server import fleetspeak_utils
+from grr_response_server import server_startup
+from grr_response_core import config
+from grr_response_core.config import contexts
+
+from fleetspeak.src.server.proto.fleetspeak_server.resource_pb2 import ClientResourceUsageRecord
+
+
+AVAILABLE_METRICS = ['mean_user_cpu_rate', 'max_user_cpu_rate', 'mean_system_cpu_rate', 
+                    'max_system_cpu_rate', 'mean_resident_memory_mib', 'max_resident_memory_mib']
```
panhania, line 30: I would keep this list as a "listing", i.e.:
```
FOO = [
    "foo",
    "bar",
    "baz",
]
```
(note that the trailing comma, it is needed to prevent the auto-formatter from folding it)

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+  A full description of all endpoints implemented within this HTTP
+  server can be found in https://github.com/simPod/grafana-json-datasource#api."""
+
+  def __init__(self, config: dict) -> None:
+    """Constructor."""
+    self.url_map = Map([
+    Rule('/', endpoint='Root', methods=["GET"]),
+    Rule('/search', endpoint='Search', methods=["POST"]),
+    Rule('/query', endpoint='Query', methods=["POST"]),
+    Rule('/annotations', endpoint='Annotations', methods=["POST"])
```
panhania, line 59: Same comment about trailing comma. This block should also be indented.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    except HTTPException as e:
+      return e
+
+  def wsgiApp(self, environ, start_response) -> Response:
+    request = JSONRequest(environ)
+    response = self.dispatchRequest(request)
+    return response(environ, start_response)
+
+  def __call__(self, environ, start_response):
+    return self.wsgiApp(environ, start_response)
```
panhania, line 77: Is this needed? Seems like a hack.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    requested_client_id = extractClientIdFromVariable(request_json_format)  # There must be a ClientID variable declated in Grafana.
+    requested_targets = [entry["target"] for entry in request_json_format["targets"]]
+    response = fetchDatapointsForTargets(requested_client_id, request_json_format["maxDataPoints"], requested_targets)
+    return JSONResponse(response=json.dumps(response), mimetype="application/json")
+
+  def onAnnotations(self, request: JSONRequest) -> JSONResponse:
+    pass
+
+
+def fetchClientIds() -> List[Text]:
```
panhania, line 109: Any reason for using camel case? Seems unpythonic.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+  # Based on an assumption that there is only one Grafana variable.
+  return next(iter(req["scopedVars"].values()))["value"]
+
+
+def main(argv: Any) -> None:
+  """Main."""
+  del argv  # Unused.
+
+  if flags.FLAGS.version:
+    print("GRRafana server {}".format(config_server.VERSION["packageversion"]))
```
panhania, line 155: Prefer to use f-strings (`f"GRRafana server {config_server.VERSION["packageversion"]}"`).

grr/server/grr_response_server/bin/grrafana_test.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+# Lint as: python3
+"""Unittest for GRRafana HTTP server."""
+from grr.test_lib import test_lib
+from grr_response_server.bin import grrafana
+
+# TODO(tsehori): We should implement tests for grrafana.py.
+class GRRafanaTest(test_lib.GRRBaseTest):
```
panhania, line 8: Please, avoid using `GRRBaseTest` unless there is a compelling reason to do so. `absltest.TestCase` is enough, especially that this is empty right now.

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,34 @@ def GetLabelsFromFleetspeak(client_id):
...
+
+  Returns:
+    A list of Client IDs.
+  """
+  res = fleetspeak_connector.CONN.outgoing.ListClients(admin_pb2.ListClientsRequest())
+  if not res.clients or not res.clients[0].client_id:
+    return []
+  clients_list = list(res.clients)
+  client_ids_list = [FleetspeakIDToGRRID(client.client_id) for client in clients_list]
+  return list(client_ids_list)
```
panhania, line 129: `client_ids_list` is already a list.

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,34 @@ def GetLabelsFromFleetspeak(client_id):
...
+def GetClientIdsFromFleetspeak() -> List[Text]:
+  """Returns client IDs of Fleetspeak-enabled database.
+
+  Returns:
+    A list of Client IDs.
+  """
+  res = fleetspeak_connector.CONN.outgoing.ListClients(admin_pb2.ListClientsRequest())
+  if not res.clients or not res.clients[0].client_id:
+    return []
+  clients_list = list(res.clients)
```
panhania, line 127: Do you need to convert it to a list?

panhania (commented):
I love the name!

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
+# Lint as: python3
+import os
+import json
+import typing
+from typing import Any, List, Tuple, Dict, Text
+
+from werkzeug.wrappers import Request, Response
```
mbushkov, line 8: We generally do not import individual classes, modules are imported instead. The only exception are imports from "typing". In case when an imported module name is too generic/ambiguous, an alias might be used. Please see https://github.com/google/grr/blob/e5549485f98df0b1b155514b3d42584f7275cd12/grr/server/grr_response_server/gui/wsgiapp.py#L26 as an example.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
+# Lint as: python3
+import os
+import json
+import typing
```
mbushkov, line 5: Instead of importing `typing` and using `typing.cast`, please add `cast` to the line below. I.e.:
`from typing import cast, Any, Dict, List, Text, Tuple`

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+from typing import Any, List, Tuple, Dict, Text
```
mbushkov, line 6: Please sort imported names alphabetically. I.e. `Any, Dict, List, Text, Tuple`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+
+from werkzeug.wrappers import Request, Response
+from werkzeug.wrappers.json import JSONMixin
+from werkzeug.routing import Map, Rule
+from werkzeug.exceptions import HTTPException, NotFound
+from werkzeug.middleware.shared_data import SharedDataMiddleware
+from werkzeug.utils import redirect
+from werkzeug.serving import run_simple
```
mbushkov, line 14: Please sort the imports using the following principle:

1. First, all Python standard library imports in alphabetical order.
2. Second, all other imports, in alphabetical order. It's ok to group the import and separate groups by newlines, but the overall order has to be alphabetical.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+from absl import flags
+
+from grr_response_core.config import server as config_server
+from grr_response_server import fleetspeak_connector
+from grr_response_server import fleetspeak_utils
+from grr_response_server import server_startup
+from grr_response_core import config
+from grr_response_core.config import contexts
+
+from fleetspeak.src.server.proto.fleetspeak_server.resource_pb2 import ClientResourceUsageRecord
```
mbushkov, line 26: Same as above: please import `resource_pb2` and use `resource_pb2.ClientResourceUsageRecord` - then it's immediately clear where the class name is coming from.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+  def __init__(self, config: dict) -> None:
+    """Constructor."""
+    self.url_map = Map([
+    Rule('/', endpoint='Root', methods=["GET"]),
+    Rule('/search', endpoint='Search', methods=["POST"]),
+    Rule('/query', endpoint='Query', methods=["POST"]),
+    Rule('/annotations', endpoint='Annotations', methods=["POST"])
+    ])
+
+  def dispatchRequest(self, request: JSONRequest) -> Any:
```
mbushkov, line 62: We use the PascalCase (a variation of camelCase with first character capitalized) for class, methods and function names throughout the code. This should be `DispatchRequest`.

mbushkov, line 62: Isn't this method supposed to return `Response`, not Any?

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    """Maps requests to different methods."""
+    adapter = self.url_map.bind_to_environ(request.environ)
+    try:
+      endpoint, values = adapter.match()
+      return getattr(self, 'on' + endpoint)(request, **values)
+    except HTTPException as e:
+      return e
+
+  def wsgiApp(self, environ, start_response) -> Response:
```
mbushkov, line 71: This should be `WsgiApp`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    request = JSONRequest(environ)
+    response = self.dispatchRequest(request)
+    return response(environ, start_response)
+
+  def __call__(self, environ, start_response):
+    return self.wsgiApp(environ, start_response)
+
+  def onRoot(self, request: JSONRequest) -> JSONResponse:
```
mbushkov, line 79: This should be called `OnRoot`. Same for other methods below.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    resource usage metrics from Fleetspeak database, or possible values
+    for a defined Grafana variable (currently supports only variables based
+    on client IDs)."""
+    if "type" in request.json:
+      # Request issued on Panel > Queries page.
+      response = AVAILABLE_METRICS
+    else:
+      # Grafana issued request on Variables > New/Edit page.
+      response = fetchClientIds()  # todo: support Grafana variables other than ClientID.
+    return JSONResponse(response=json.dumps(response), mimetype="application/json")
```
mbushkov, line 94: Please put the mime type into a constant, as it's used multiple times.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+  def onRoot(self, request: JSONRequest) -> JSONResponse:
+    """Returns OK message to database connection check."""
+    return JSONResponse()
+
+  def onSearch(self, request: JSONRequest) -> JSONResponse:
+    """Depending on the request type, returns either available client
+    resource usage metrics from Fleetspeak database, or possible values
+    for a defined Grafana variable (currently supports only variables based
+    on client IDs)."""
+    if "type" in request.json:
```
mbushkov, line 88: As far as I understand presense of "type" in request.json means different Grafana request types. Please add a comment detailing this.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+      response = AVAILABLE_METRICS
+    else:
+      # Grafana issued request on Variables > New/Edit page.
+      response = fetchClientIds()  # todo: support Grafana variables other than ClientID.
+    return JSONResponse(response=json.dumps(response), mimetype="application/json")
+
+  def onQuery(self, request: JSONRequest) -> JSONResponse:
+    """Given a client ID as a Grafana variable and targets (resource usages),
+    returns datapoints in a format Grafana can interpret."""
+    request_json_format = request.json
```
mbushkov, line 99: `request_json_format` feels like a slightly misleading name here. The variable contains not the format name, but the data in this format. Maybe something along the lines of `json_data`?

grr/server/grr_response_server/bin/grrafana.py
mbushkov, line 109: Any reason to have this function instead of simply inlining the call to `fleetspeak_utils.GetClientIdsFromFleetspeak()`?

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    requested_targets = [entry["target"] for entry in request_json_format["targets"]]
+    response = fetchDatapointsForTargets(requested_client_id, request_json_format["maxDataPoints"], requested_targets)
+    return JSONResponse(response=json.dumps(response), mimetype="application/json")
+
+  def onAnnotations(self, request: JSONRequest) -> JSONResponse:
+    pass
+
+
+def fetchClientIds() -> List[Text]:
+  """Fetches GRR client IDs that have resource usage records in Fleetspeak database."""
```
mbushkov, line 110: I'd suggest not fetching the client ids. This will work on trivial deployments only. Otherwise you might get hundreds of thousands of values and that would make the autocompletion meaningless.
Will Grafana UI allow you to insert an abritrary client id if you simply return an empty list here?

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    """Depending on the request type, returns either available client
+    resource usage metrics from Fleetspeak database, or possible values
+    for a defined Grafana variable (currently supports only variables based
+    on client IDs)."""
+    if "type" in request.json:
+      # Request issued on Panel > Queries page.
+      response = AVAILABLE_METRICS
+    else:
+      # Grafana issued request on Variables > New/Edit page.
+      response = fetchClientIds()  # todo: support Grafana variables other than ClientID.
```
mbushkov, line 93: Fetching the clients might not be feasible - please see my comment below detailing this issue.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+
+def fetchClientIds() -> List[Text]:
+  """Fetches GRR client IDs that have resource usage records in Fleetspeak database."""
+  return fleetspeak_utils.GetClientIdsFromFleetspeak()
+
+
+Datapoint = Tuple[float, int]
+Datapoints = List[Datapoint]
+TargetWithDatapoints = Dict[Text, Datapoints]
+def fetchDatapointsForTargets(client_id: Text, limit: int, targets: List[Text]) -> List[TargetWithDatapoints]:
```
mbushkov, line 117: Top-level functions and class definitions should be separated by 2 blank lines from other definitions. Note that type aliases can be grouped together. I.e. this code should look like:
```
Datapoints = List[Datapoint]
TargetWithDatapoints = Dict[Text, Datapoints]


def _FetchDatapointsForTargets(client_id: Text, limit: int, targets: List[Text]) -> List[TargetWithDatapoints]:
```

mbushkov, line 117: If this function is not used outside of this module, please declare it as a module-private function: `_FetchDatapointsForTargets`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+  """Fetches a list of <datapoint, timestamp> tuples for each target metric from Fleetspeak database."""
+  records_list = fleetspeak_utils.FetchClientResourceUsageRecordsFromFleetspeak(client_id, limit)
+  response = list()
```
mbushkov, line 120: Better: `response = []`

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,165 @@
...
+    record_values = [record.mean_resident_memory_mib for record in records_list]
+  elif target == "max_resident_memory_mib":
+    record_values = [record.max_resident_memory_mib for record in records_list]
+  return [(v, r.server_timestamp.seconds * 1000) for (v, r) in zip(record_values, records_list)]
+
+
+def extractClientIdFromVariable(req: JSONRequest) -> Text:
+  """Extracts the client ID from a Grafana JSON request."""
+  # Based on an assumption that there is only one Grafana variable.
+  return next(iter(req["scopedVars"].values()))["value"]
```
mbushkov, line 147: This is very complicated. Please rewrite as:
```
scoped_vars_values = list(req["scopedVars"].values())
return scoped_vars_values[0]["value"]
```

Quick question - I'd expect the client id variable to have a name. Instead of using `.values` can we just access the dictionary by key? I.e. `req["scopedVars"]["clientId"]["value"]`?

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,34 @@ def GetLabelsFromFleetspeak(client_id):
       grr_labels.append(fs_label.label)
 
   return grr_labels
+
+
+def GetClientIdsFromFleetspeak() -> List[Text]:
```
mbushkov, line 118: Please see my comment above. Listing all the clients is guaranteed to fail on large deployments, since the number of results will be too great.

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,34 @@ def GetLabelsFromFleetspeak(client_id):
...
+  """Returns aggregated resource usage metrics of a client in Fleetspeak-enabled database.
+
+  Args:
+    client_id: Id of the client to fetch Fleetspeak resource usage records for.
+    limit: Max number of resource usage records to retrieve.
+
+  Returns:
+    A list of client resource usage records retrieved from Fleetspeak.
+  """
+  res = fleetspeak_connector.CONN.outgoing._stub.FetchClientResourceUsageRecords(
```
mbushkov, line 142: Why do you need to use `CONN.outgoing._stub` and not simply `CONN.outgoing`. `_stub`, given that its name starts with an underscore, is a semi-private attribute that is likely an implementation detail of the `CONN.outgoing` object.

grr/server/setup.py
```diff
@@ -185,6 +185,8 @@ def make_release_tree(self, base_dir, files):
             "grr_response_server.distro_entry:Worker",
             "grr_admin_ui = "
             "grr_response_server.distro_entry:AdminUI",
+            "grr_afana = "
```
mbushkov, line 188: I like the `grrafana` name, but naming the binary as `grr_afana` feels somewhat confusing. `grrafana` would be much easier, but then we'd break the convention of prefixing all our binary names with `grr_`. @panhania - what's your take on this?

grr/server/grr_response_server/bin/grrafana.py
mbushkov: I think this is a requirement of the WSGI standard. It treats a web server as a function that gets a request and a callback function to build a response.

mbushkov (commented):
Looks good for the first iteration. Please see my comments below.

grr/server/grr_response_server/bin/grrafana.py
tsehori: This is actually a part of boilerplate code similar to this:
https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/grr/server/grr_response_server/bin/frontend.py#L325-L327
What do you think? Should we change both or stay consistent with the existing codebase?

grr/server/grr_response_server/fleetspeak_utils.py
tsehori: You're right, we don't :)

grr/server/grr_response_server/bin/grrafana.py
tsehori: `JSONResponse` to be exact, but you're definitely right

grr/server/grr_response_server/bin/grrafana.py
tsehori: No particular reason, I'll redirect it.

grr/server/grr_response_server/bin/grrafana.py
tsehori: That's a great question, and I believe we also briefly discussed this with @max-vogler. I'll research it.

grr/server/grr_response_server/bin/grrafana.py
tsehori: This is @mbushkov's suggestion, see his comment here https://github.com/google/grr/pull/832#discussion_r477530893

grr/server/grr_response_server/bin/grrafana.py
tsehori: I made note and changed everything to do with the imports; please let me know if it's OK!

grr/server/setup.py
tsehori: Another possibility can be `grr_grrafana` I guess.

grr/server/grr_response_server/fleetspeak_utils.py
tsehori: You're right, that's definitely a better way of achieving this goal. I implemented a new method in class `OutgoingConnection` in Fleetspeak to support it, you can check it out here google/fleetspeak@1947d45.

grr/server/grr_response_server/bin/grrafana.py
tsehori: That's a small issue. You're right that the Client ID variable has a name, the problem is it is up to the user to choose its name on creation within the Grafana UI. You can check out the contents of a sample `req["scopedVars"]` dictionary here:
```
{
'myname': {'text': 'C.<some_client_id>', 'value': 'C.<some_client_id>'}, 
'__interval': {'text': '30m', 'value': '30m'}, 
'__interval_ms': {'text': '1800000', 'value': 1800000}
}
```
We can ask users to name the variable with a specific name with documentation, but I did not think it was optimal. Do you have any ideas?

grr/server/grr_response_server/fleetspeak_utils.py
tsehori: I'll reply to all concerns related to listing all client IDs in this thread https://github.com/google/grr/pull/832#discussion_r477534903.

grr/server/grr_response_server/bin/grrafana.py
tsehori: I'll reply to all concerns related to listing all client IDs in this thread https://github.com/google/grr/pull/832#discussion_r477534903.

grr/server/grr_response_server/bin/grrafana.py
tsehori: Ok so I have some findings.

There are [different types of Grafana variables](https://grafana.com/docs/grafana/latest/variables/templates-and-variables/#variable-types). A user may add a Grafana variable of type [text box](https://grafana.com/docs/grafana/latest/variables/variable-types/add-text-box-variable/#add-a-text-box-variable) (instead of the default [query type](https://grafana.com/docs/grafana/latest/variables/variable-types/add-query-variable/#add-a-query-variable) variable), which will *not* issue a request by Grafana for a list of possible values. Within this text box variable, a user can type in the relevant Client ID, all of that without support for auto-completion of course.

In this scenario, we might decide to not support the second type of response of the `/search`, endpoint which is issued when the default query type is selected for the Grafana variable. This is also discussed in this thread https://github.com/google/grr/pull/832#discussion_r477532744.

What do you think @mbushkov?

grr/server/grr_response_server/bin/grrafana.py
panhania: This linked code fragment was written way before we migrated to Python 3, so it still uses "the old way". In my opinion, new code should not repeat old mistakes for the sake of consistency. The linked fragment should be changed, I agree, but, since it is not related, not as part of this pull request.

grr/server/setup.py
panhania: No clear opinion here, I am fine with all of them. For the sake of consistency, I would probably pick `grr_grrafana`.

grr/server/grr_response_server/bin/grrafana_test.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+# Lint as: python3
+"""Unittest for GRRafana HTTP server."""
+from absl.testing import absltest
+
+from grr_response_server.bin import grrafana
+
+# TODO(tsehori): We should implement tests for grrafana.py.
+class GRRafanaTest(absltest.TestCase):
```
panhania, line 9: Nit: The test case should correspond to the class or function name it is testing. Since the tested class is named `Grrafana`, the test case name should be `GrrafanaTest`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,188 @@
...
+from werkzeug.wrappers import json as werkzeug_wrappers_json
+
+
+AVAILABLE_METRICS = [
+    'mean_user_cpu_rate',
+    'max_user_cpu_rate',
+    'mean_system_cpu_rate',
+    'max_system_cpu_rate',
+    'mean_resident_memory_mib',
+    'max_resident_memory_mib',
```
panhania, line 30: Almost all strings in this file use double quotes (as is the rest of the GRR codebase) and I don't think there is a compelling reason to do it differently here.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,188 @@
...
+  server_startup.Init()
+  fleetspeak_connector.Init()
+  werkzeug_serving.run_simple('127.0.0.1',
+             5000,
+             Grrafana({}),
+             use_debugger=True,
+             use_reloader=True)
+
+
+if __name__ == '__main__':
```
panhania, line 187: Same comment about the string quotes.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,188 @@
...
+
+  if flags.FLAGS.version:
+    print("GRRafana server {}".format(config_server.VERSION["packageversion"]))
+    return
+
+  config.CONFIG.AddContext(contexts.GRRAFANA_CONTEXT,
+                           "Context applied when running GRRafana server.")
+  server_startup.Init()
+  fleetspeak_connector.Init()
+  werkzeug_serving.run_simple('127.0.0.1',
```
panhania, line 180: Same comment about the string quotes.

tsehori:
Thanks for for your comments and clarifications @panhania and @mbushkov. I appreciate it!

In addition to the tests for `fleetspeak_utils.py`, I [added tests](https://github.com/google/grr/pull/832/files#diff-5c3c390f16ef2377252f0d7523948629R99) in `grrafana_test.py` for the `Grrafana` object. It seemed a bit odd to have tests for one set of functions but not the other on the same PR.
I generally took a similar approach to [`ApiIntegrationTest`](https://github.com/google/grr/blob/0e03976c5c5b694c210c0f392581364ddefa1f4b/grr/server/grr_response_server/gui/api_integration_test_lib.py#L31), but with one major difference. Werkzeug has its own set of [test utilities](https://werkzeug.palletsprojects.com/en/1.0.x/test/#module-werkzeug.test) which is quite straightforward, so I utilized it to make requests to the GRRafana application without using the [requests](https://requests.readthedocs.io/en/master/) library like `ApiIntegrationTest` does.

@panhania - I think I addressed most of your concerns. Please let me know if there's anything else you think worth iterating over.
@mbushkov - There are some things left to decide, I left the conversations unresolved so we can take an extra look at them and discuss them.

grr/server/grr_response_server/bin/grrafana.py
mbushkov: Following an offline discussion: we will be the ones providing the Grafana dashboard to display clients load, so I'd expect that we would be able to pre-configure the variable name in the dashboard configuration. If this is correct, then there's no issue and we can rely on the variable name being predefined.

But in any case, even if the user has to define a variable by hand, it's much cleaner to document the intended name and ask the user to use it. The current code would effectively break if there's more than one variable defined, which feels very fragile.

grr/server/grr_response_server/bin/grrafana.py
mbushkov: Resolving, as the issue seems to be fixed. For the record: I meant PascalCase, which is a camelCase starting with a capitalized character.

grr/server/setup.py
mbushkov: +1 for grr_grrafana.

grr/server/grr_response_server/bin/grrafana.py
tsehori: Following an offline discussion with @mbushkov, I'll remove the method `fetchClientIds` from the Grrafana object and the function `GetClientIdsFromFleetspeak` from `fleetspeak_utils.py` that support this functionality.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,188 @@
...
+
+AVAILABLE_METRICS = [
+    "mean_user_cpu_rate",
+    "max_user_cpu_rate",
+    "mean_system_cpu_rate",
+    "max_system_cpu_rate",
+    "mean_resident_memory_mib",
+    "max_resident_memory_mib",
+]
+RESPONSE_MIME_TYPE = "application/json"
```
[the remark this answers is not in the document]
panhania: I would not even put it in the constant. The fact that it is `application/json` is an inherent consequence of the implementation, you cannot change it. Having this constant named as `JSON_MIME_TYPE` would be fine, but `RESPONSE_MIME_TYPE` suggests you can change the value and still have coherent code (e.g. like you can add and remove items from `AVAILABLE_METRICS`), but you cannot—it is deeply coupled with the implementation.

grr/server/grr_response_server/bin/grrafana.py
tsehori: As you suspected, we will indeed be able to pre-configure the variable name in the Grafana dashboard when we supply it, within the [textbox variable](https://grafana.com/docs/grafana/latest/variables/variable-types/add-text-box-variable/#add-a-text-box-variable) as discussed in the [other conversation](https://github.com/google/grr/pull/832#discussion_r478403721). Therefore, the issue with the code here is resolved.

However, there is one minor thing left to discuss. The library [grafanalib](https://github.com/weaveworks/grafanalib), which we also use within GRR in the [`grr_grafanalib_dashboards`](https://github.com/google/grr/tree/master/monitoring/grafana) package, [does support adding variables](https://grafanalib.readthedocs.io/en/latest/api/grafanalib.html#grafanalib.core.Template) to scripted dashboards, but does not support the specific type of variable we need (textbox).
I don't suspect it to be a problem when we get to the dashboards in a later PR, but it's something to be aware off, since we will essentially be forced to work around the API of grafanalib (we did in it [other cases](https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/monitoring/grafana/grr_grafanalib_dashboards/dashboards_to_generate/fleetspeak.dashboard.py#L52) before, though not as significant as this one).

grr/server/grr_response_server/bin/grrafana.py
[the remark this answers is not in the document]
mbushkov: @panhania , I think it still makes sense to make it a constant to avoid duplicating a non-trivial string literal. +1 re the constant name. JSON_MIME_TYPE is a better one.

grr/server/grr_response_server/fleetspeak_utils_test.py
```diff
@@ -133,6 +133,35 @@ def testRestartFleetspeakGrrService(self, mock_conn):
...
+          resource_pb2.ClientResourceUsageRecord(mean_user_cpu_rate=1,
+                                                 max_system_cpu_rate=2),
+          resource_pb2.ClientResourceUsageRecord(mean_user_cpu_rate=4,
+                                                 max_system_cpu_rate=8)
+      ]
+      self.assertListEqual(
+          fleetspeak_utils.FetchClientResourceUsageRecordsFromFleetspeak(
+              _TEST_CLIENT_ID, 10), expected_records_list)
+      conn.outgoing.FetchClientResourceUsageRecords.assert_called_once()
+      insert_args, _ = conn.outgoing.FetchClientResourceUsageRecords.call_args
```
mbushkov, line 157: "insert_args" sounds as an odd name. Is it a copy-paste artifact? Also, would it make sense to use https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with instead of getting the call_args and then doing assertEqual?

mbushkov (changes requested):
Looks very good, just 2 small points

grr/server/grr_response_server/fleetspeak_utils_test.py
tsehori: It isn't an artifact, but it is implemented in a similar way in existing code in tests for other `fleetspeak_utils.py` functions. For instance:
https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/grr/server/grr_response_server/fleetspeak_utils_test.py#L112-L117
However I do agree with you that using `assert_called_with` is far more readable. In this case, you actually do not even need to use `assertEqual`.

grr/server/grr_response_server/bin/grrafana.py
mbushkov: Thanks for researching this. I'd look into adding this functionality to grafanalib and submitting it upstream. If adding such support is not too much work this would be the best option. Otherwise, yes, we can work around the API of grafanlib if needed.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+  A full description of all endpoints implemented within this HTTP
+  server can be found in https://github.com/simPod/grafana-json-datasource#api."""
+
+  def __init__(self) -> None:
+    """Constructor."""
+    self.url_map = werkzeug_routing.Map([
+        werkzeug_routing.Rule("/", endpoint="Root", methods=["GET"]),
+        werkzeug_routing.Rule("/search", endpoint="Search", methods=["POST"]),
+        werkzeug_routing.Rule("/query", endpoint="Query", methods=["POST"]),
+        werkzeug_routing.Rule("/annotations", endpoint="Annotations", methods=["POST"]),
```
max-vogler, line 62: Please enforce an 80-character maximum line width for all Python files.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+class Grrafana(object):
+  """GRRafana HTTP server instance.
+  
+  A full description of all endpoints implemented within this HTTP
+  server can be found in https://github.com/simPod/grafana-json-datasource#api."""
+
+  def __init__(self) -> None:
+    """Constructor."""
```
max-vogler, line 57: "Constructor." violates [pystyle](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) and contains no information. Please avoid non-descriptive docstrings. I think our linter would complain about missing docstrings, so use something like `"""Initializes a new GRRafana HTTP server instance."""`

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+    self.url_map = werkzeug_routing.Map([
```
max-vogler, line 58: `url_map` is not part of a public API of the `Grrafana` class (e.g. code outside of the class doesn't need to read or manipulate it). Please mark it as private by prepending an `_`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+        werkzeug_routing.Rule("/", endpoint="Root", methods=["GET"]),
+        werkzeug_routing.Rule("/search", endpoint="Search", methods=["POST"]),
+        werkzeug_routing.Rule("/query", endpoint="Query", methods=["POST"]),
+        werkzeug_routing.Rule("/annotations", endpoint="Annotations", methods=["POST"]),
+    ])
+
+  def DispatchRequest(self, request: JSONRequest) -> JSONResponse:
```
max-vogler, line 65: `DispatchRequest` and `On***` are not directly referenced by code outside of the `Grrafana` class. It is best practice to mark them as private to allow flexible development in the future.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+    """Maps requests to different methods."""
+    adapter = self.url_map.bind_to_environ(request.environ)
+    try:
+      endpoint, values = adapter.match()
+      return getattr(self, "On" + endpoint)(request, **values)
```
max-vogler, line 70: Metaprogramming like `getattr` prevents static code introspection: Linters and IDEs cannot know that this line actually calls the methods defined below. Please reference methods statically, e.g. by using a big if-else block or defining the lookup table in a dictionary.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+    except werkzeug_exceptions.HTTPException as e:
+      return e
+
+  def WsgiApp(self, environ, start_response) -> werkzeug_wrappers.Response:
+    request = JSONRequest(environ)
+    response = self.DispatchRequest(request)
+    return response(environ, start_response)
+
+  def __call__(self, environ, start_response):
+    return self.WsgiApp(environ, start_response)
```
max-vogler, line 80: Since `__call__` only calls `WsgiApp` and `WsgiApp` does not seem to be called elsewhere, can you inline it into `__call__`?

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+
+  def OnRoot(self, request: JSONRequest) -> JSONResponse:
+    """Returns OK message to database connection check."""
+    return JSONResponse()
+
+  def OnSearch(self, request: JSONRequest) -> JSONResponse:
+    """Depending on the type of request Grafana is issuing, this method returns either
```
max-vogler, line 87: Per [pystyle](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods), docstrings should have a descriptive one-liner, and then after a blank line can contain more information. The idea is that you can roughly understand what a function does by reading one line only.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+    default=False,
+    allow_override=True,
+    help="Print the GRR console version number and exit immediately.")
+
+
+class JSONRequest(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Request):
+  pass
+
+
+class JSONResponse(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Response):
```
max-vogler, line 46: `JSONMixin` actually does not affect constructing a Response, see [json.py](https://github.com/pallets/werkzeug/blob/master/src/werkzeug/wrappers/json.py). The JSONMixin is only used for parsing JSON. This is reflected in your following code, where you need to call `json.dumps` and setting the mimetype manually every time.

I suggest removing `JSONMixin` from `JSONResponse` and overriding `__init__` to call `json.dumps` on whatever data is passed as `response` parameter. Also, the new constructor should always set the mimetype.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+Datapoint = Tuple[float, int]
+Datapoints = List[Datapoint]
+TargetWithDatapoints = Dict[Text, Datapoints]
+
+
+def _FetchDatapointsForTargets(
+    client_id: Text, limit: int,
+    targets: List[Text]) -> List[TargetWithDatapoints]:
```
max-vogler, line 125: Functions should be lax in what type they receive. Since you only iterate over targets, you don't need it to be a `List` - you don't require constant time access to random fields. Instead, use a laxer type like `targets: Iterable[Text]`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+  """Fetches a list of <datapoint, timestamp> tuples for each target metric from Fleetspeak database."""
+  records_list = fleetspeak_utils.FetchClientResourceUsageRecordsFromFleetspeak(
+      client_id, limit)
+  response = []
+  for target in targets:
+    datapoints_for_single_target = _CreateDatapointsForTarget(
+        target, records_list)
+    target_datapoints_dict = cast(TargetWithDatapoints, {
```
max-vogler, line 133: The type definition is unfitting: `TargetWithDatapoints` is defined as Dict with string keys and Datapoints values, yet you assign "target" to a string. Dict is not the type you want here, define a class or a namedtuple instead (our Python version does not support dataclasses yet).

max-vogler: Then, this cast is no longer needed.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+    record_values = [record.max_user_cpu_rate for record in records_list]
+  elif target == "mean_system_cpu_rate":
+    record_values = [record.mean_system_cpu_rate for record in records_list]
+  elif target == "max_system_cpu_rate":
+    record_values = [record.max_system_cpu_rate for record in records_list]
+  elif target == "mean_resident_memory_mib":
+    record_values = [record.mean_resident_memory_mib for record in records_list]
+  elif target == "max_resident_memory_mib":
+    record_values = [record.max_resident_memory_mib for record in records_list]
+  return [(v, r.server_timestamp.seconds * 1000)
```
max-vogler, line 155: Your function returns milliseconds, but uses only second-precision. Please extract the millisecond part from `server_timestamp.nanos`.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -0,0 +1,185 @@
...
+          for (v, r) in zip(record_values, records_list)]
+
+
+def _ExtractClientIdFromVariable(req: JSONRequest) -> Text:
+  """Extracts the client ID from a Grafana JSON request."""
+  scoped_vars_dict = req["scopedVars"]
```
max-vogler, line 161: This variable can be inlined in the next line, e.g.

```
return req["scopedVars"]["ClientID"]["value"]
```

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,22 @@ def GetLabelsFromFleetspeak(client_id):
       grr_labels.append(fs_label.label)
 
   return grr_labels
+
+
+def FetchClientResourceUsageRecordsFromFleetspeak(
```
max-vogler, line 118: The module already contains "fleetspeak", I suggest to simplify this very long function name to: `FetchClientResourceUsageRecords`.

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -16,6 +16,7 @@
 from fleetspeak.src.common.proto.fleetspeak import common_pb2 as fs_common_pb2
 from fleetspeak.src.common.proto.fleetspeak import system_pb2 as fs_system_pb2
 from fleetspeak.src.server.proto.fleetspeak_server import admin_pb2
+from fleetspeak.src.server.proto.fleetspeak_server.resource_pb2 import ClientResourceUsageRecord
```
max-vogler, line 19: Per pystyle, do not import individual members from modules, instead import the whole module `resource_pb2` and reference the member `resource_pb2.ClientResourceUsageRecord`.

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -112,3 +113,22 @@ def GetLabelsFromFleetspeak(client_id):
...
+    client_id: Id of the client to fetch Fleetspeak resource usage records for.
+    limit: Max number of resource usage records to retrieve.
+
+  Returns:
+    A list of client resource usage records retrieved from Fleetspeak.
+  """
+  res = fleetspeak_connector.CONN.outgoing.FetchClientResourceUsageRecords(
+      admin_pb2.FetchClientResourceUsageRecordsRequest(
+          client_id=GRRIDToFleetspeakID(client_id), limit=limit))
+  if not res or not res.records or not res.records[0]:
```
max-vogler, line 132: In what edge case does the response have at least one record, but the record would evaluate to false? Please clarify in a comment in the code and add a test case to trigger this edge case.

max-vogler (changes requested):

grr/server/grr_response_server/bin/grrafana.py
tsehori: Small question - what do we do about lines where enforcing this limit is non-trivial? As in, you have to rename an import statement or a variable.
For example:

```python
conn.outgoing.FetchClientResourceUsageRecords.return_value = admin_pb2.FetchClientResourceUsageRecordsResponse(
```

grr/server/grr_response_server/fleetspeak_utils.py
tsehori: I agree that it is quite trivial that it reflects retrieving data from Fleetspeak (and that this function name is long); on the other hand, it follows a similar convention that existed in the rest of the module, for instance:
https://github.com/google/grr/blob/b2969c7ed6f5c6e62919fa66625d992e7b5e0b34/grr/server/grr_response_server/fleetspeak_utils.py#L86-L87
What do you think?

grr/server/grr_response_server/bin/grrafana.py
tsehori: The `__call__` method makes the `Grrafana` object callable, which otherwise causes a TypeError. Also see the [discussion here](https://github.com/google/grr/pull/832#discussion_r477474772).
By inlining the method, do you mean as follows?
```python
  __call__ = lambda self, environ, start_response: self.WsgiApp(environ, start_response)
```

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -44,7 +44,21 @@ class JSONRequest(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Request):
...
+      status=None,
+      headers=None,
+      mimetype=None,
+      content_type=None,
+      direct_passthrough=False,
+  ):
+    try:
+      json_response = json.dumps(response)
+    except TypeError:
+      json_response = response
```
panhania, line 60: That... loks very questionable. Why is this needed? Cannot we just have a well-typed constructor?

grr/server/grr_response_server/bin/grrafana.py
tsehori: I agree that `JSONMixin` does not affect the response, but its property `json` is utilized later within the tests for both endpoints `/search` and `/query`. Also, not using `json` will cause us to decode the JSON bytes ourselves from the response.
Therefore, I think it is wise to leave the inheritance as is and get this property from the mixin.

Re overriding `__init__` and modifying the relevant parameters: I agree. PTAL on this and let me know what you think.

grr/server/grr_response_server/bin/grrafana.py
tsehori: It is needed because while testing, Werkzeug also uses a [ClosingIterator](https://werkzeug.palletsprojects.com/en/1.0.x/wsgi/#werkzeug.wsgi.ClosingIterator) object if I understand correectly. So, when a [Response](https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/#werkzeug.wrappers.Response) (specifically a JSONResponse) object is constructed with the parameter [`response`](https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/#werkzeug.wrappers.BaseResponse.response) being a `ClosingIterator` object, then `json.dumps()` fails.

grr/server/grr_response_server/bin/grrafana.py
tsehori: I changed `TargetWithDatapoints` to a namedtuple like you suggested. The only caveat is that Grafana cannot interpret correctly the JSON that is extracted from this namedtuple, so I use [`as_dict()`](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict) in the method `_OnQuery`.

Let me know if you think this is the right way to go about this, because it feels like a slight workaround.

grr/server/grr_response_server/fleetspeak_utils.py
tsehori: You're right, I forgot that empty lists in Python will [evaluate as false](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty) :)

grr/server/grr_response_server/bin/grrafana.py
max-vogler: panhania's point is valid: I suggest probing for ClosingIterator (or other edge cases) specifically instead of using error handling. In your code, errors during json conversion would be ignored and the raw object would be returned with a HTTP 200 code. This leads to hard-to-debug runtime issues.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -152,14 +184,16 @@ def _CreateDatapointsForTarget(
     record_values = [record.mean_resident_memory_mib for record in records_list]
   elif target == "max_resident_memory_mib":
```
max-vogler, line 185: `target` comes from the HTTP request and is not validated to be exclusively mean_user_cpu_rate, max_user_cpu_rate, or max_resident_memory_mib. When provided any other value ,this function would fail in a cryptic way (because `record_values` is not initialized). Please add an else branch to throw a better exception.

max-vogler (changes requested):

grr/server/grr_response_server/bin/grrafana.py
tsehori: As far as [Werkzeug's tutorial](https://werkzeug.palletsprojects.com/en/1.0.x/tutorial/#step-2-the-base-structure) goes, you're right; it is being implemented the way it is so that `WsgiApp` can be wrapped to apply [middleware](https://werkzeug.palletsprojects.com/en/0.14.x/middlewares/#module-werkzeug.wsgi).

grr/server/grr_response_server/fleetspeak_utils.py
```diff
@@ -129,6 +130,6 @@ def FetchClientResourceUsageRecordsFromFleetspeak(
   res = fleetspeak_connector.CONN.outgoing.FetchClientResourceUsageRecords(
       admin_pb2.FetchClientResourceUsageRecordsRequest(
           client_id=GRRIDToFleetspeakID(client_id), limit=limit))
-  if not res or not res.records or not res.records[0]:
+  if not res.records:
```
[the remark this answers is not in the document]
tsehori: Yes, `res.records` is a repeated field.

I thought that this can solve a problem where the response returned from the gRPC service is not what we expect; from instance, [an error that rises](https://github.com/google/fleetspeak/blob/616a079d4c55e4637574ddda1804ce36d68cde18/fleetspeak/src/server/admin/admin.go#L253) from Fleetspeak's side. So if we get an [`idErr`](https://github.com/google/fleetspeak/blob/298e9ae7573cf0c2f29b0c1e68a0a11370259c53/fleetspeak/src/common/ids.go#L41), then `FetchClientResourceUsageRecords` on GRR's side will essentially not crash.

I tried researching it a little bit, but at the moment did not arrive at a final conclusion; is it possible that such an error within the gRPC service can crash the HTTP server? That's a situation we want to avoid in a case where a Grafana user accidentally enters a wrong client ID. My intention was that checking whether `res.records` is falsey will imply that it is either empty (in this case it does not matter) or it does not exist as a property of the `res`, in which case, it isn't a valid instance of `FetchClientResourceUsageRecordsResponse`.

I agree with you that in a case where it is not possible, then this if statement is unnecessary.
What do you think?

grr/server/grr_response_server/bin/grrafana.py
tsehori: I agree; I added an `else` branch with a `NameError` exception, and also added a specific test case for it.

Just to have it written here a side note; the reason we chose to have it as a large if-else branch and not using `getattr` was to not use metaprogramming, since we know that the available resource usage metrics are rather fixed and not likely to change often.

tsehori:
I addressed most comment reviews and left notes on some I'm unsure about.
PTAL

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -46,20 +47,12 @@ class JSONRequest(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Request):
...
-      mimetype=None,
-      content_type=None,
-      direct_passthrough=False,
-  ):
-    try:
-      json_response = json.dumps(response)
-    except TypeError:
-      json_response = response
-    super().__init__(response=json_response, mimetype=JSON_MIME_TYPE)
+  def __init__(self, *args, **kwargs) -> None:
```
max-vogler, line 50: Specify `response=None` as first argument before *args and **kwargs to support the common `Response` initialization where the contents are passed as first argument.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -46,20 +47,12 @@ class JSONRequest(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Request):
...
+    kwargs["mimetype"] = JSON_MIME_TYPE
+    response = kwargs.get("response")
+    if response and not isinstance(response, werkzeug_wsgi.ClosingIterator):
```
max-vogler, line 53: Use `response is not None` instead of `response`. `if response` is `False` for `''` which would be a valid response.

max-vogler (changes requested):

max-vogler (approved):

grr/server/grr_response_server/fleetspeak_utils.py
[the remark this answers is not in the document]
max-vogler: Protobuf are strictly typed - so even if the RPC fails, if the response parses as your expected Protobuf message, you can expect `res.records` to be a list (possibly empty).

If the RPC fails, `FetchClientResourceUsageRecords` should fail as well - it's only a wrapper around the RPC. There is no benefit in returning an empty list if the underlying call failed due to some error. This can actually make debugging issues very hard, because underlying issues are concealed.

PR #833, opened 2020-08-26
Client details drawer

danielberbece:
**Note:** This PR also contains the changes from PR #826 and should be reviewed **only** after that one is merged.

Implemented a drawer to host the client details information. The drawer uses MatDrawer component from the Angular Material library. The drawer toggles between open and closed when clicking on the client details button, in the client overview panel, hovering over the approval and flows entries, providing a full view of the client. When the drawer is open, the scrolling is restricted only to the drawer panel. Also, opening the drawer also updates the URL, by appending `/details` to the previous URL. Closing the drawer removes the `/details` from the URL. That way, the client details page can be shared easily. To support this, I also added the route on which the client page opens together with the details drawer in the `CLIENT_ROUTES` array.

Unit tests for this behaviour were also added, but one expectation didn't work on the jasmine framework, but worked on manual testing. I commented the expectation and also added an explanation for what the problem seems to be, for anyone brave enough and ready for a challenge.


grr/server/grr_response_server/gui/ui/components/client/client.ts
```diff
@@ -46,6 +51,36 @@ export class Client implements OnInit, OnDestroy {
...
 
+  ngAfterViewInit() {
+    this.clientDetailsDrawers.closedStart.subscribe(() => {
+      const urlTokens = this.location.path().split('/');
+      this.location.go(urlTokens.slice(0, -1).join('/'));
+    });
+
+    this.location.onUrlChange(url => {
+      const urlTokens = url.split('/');
+      if (urlTokens[urlTokens.length - 1] === 'details') {
```
max-vogler, line 62: Please introduce and re-use a function that receives url segments, checks for 'details' and opens the drawer. This function can be used in `onUrlChange` and the following `route.url.pipe`.

grr/server/grr_response_server/gui/ui/components/client/client_test.ts
```diff
@@ -172,49 +180,103 @@ describe('Client Component', () => {
...
-    snackbarDivButton!.dispatchEvent(new MouseEvent('click'));
-    fixture.detectChanges();
-    fixture.whenRenderingDone().then(() => {
-      expect(facade.addClientLabel).toHaveBeenCalledWith('testlabel');
-    });
+    expect(drawerOpenSpy).toHaveBeenCalled();
+    detailsDrawer.close();
   });
+
+  it('correctly updates URL for client details drawer usage', fakeAsync(() => {
```
max-vogler, line 226: This test is too long. If it breaks, the reader will need to inspect 60 LOC. Please split this test up into smaller tests.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/client/client.ts
```diff
@@ -46,6 +51,36 @@ export class Client implements OnInit, OnDestroy {
...
+        this.clientDetailsDrawers.open();
+      }
+    });
+
+    this.route.url.pipe(map(url => url[url.length - 1]), take(1))
```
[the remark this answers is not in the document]
danielberbece: There is no need to take more than the first emitted value. We want to see the URL when the component loads, in order to know whether or not to open the drawer. After that, there is no need to still receive updates, so we can remove the subscription instance.

grr/server/grr_response_server/gui/ui/components/client/client.ts
danielberbece: This will require to make a global function, because in the next PR the logic is split between the `client` component and the `client-overview` component. Is that OK?

max-vogler: No, let's keep it separate functions for now then.

grr/server/grr_response_server/gui/ui/components/client/client.ts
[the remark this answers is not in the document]
max-vogler: After that, there **is** a need to receive updates, but this is handled by your `this.location.onUrlChange` code. Manipulating the URLs both via Router and Location seems like a code smell to me. Ideally, we would use Router/ActivatedRoute for both. Possibly, registering `/details` as a child route would allow us to keep the `Client` page loaded and load the drawer without needing to manipulate URLs so much ourselves.

This refactoring is not blocking for this PR, but please look into it, if you find the time.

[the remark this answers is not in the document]
danielberbece: I agree, having only one function to handle the URL/Route change is preferred. I tried using only Router at first, but it reloaded the page, which wasn't what we wanted. I'll look into child routes, see if it works that way. I'll address this issue in #836.

max-vogler (approved):

PR #837, opened 2020-09-01
Detect optional path arguments in GRR API routes and present them in the OpenAPI description

alexandrucosminmihai:
## Introduction
In this PR I introduce code that figures out if there are multiple routes associated to a singe router method which actually just describe a single route with optional path arguments. The implementation is a bit more generic than what is currently necessary for the routes described in `api_call_router.py`, in the sense that it also treats multiple routes with different HTTP methods associated with a single router method, as well as multiple routes with the same HTTP method but different path components and different required / optional path arguments for the same router method. I implemented it this way because it covers more theoretically possible cases (but which aren't necessarily a good practice, such as using the same router method for routes with different HTTP methods) and because it allows for easier testing of the generated OpenAPI description in `metadata_test.py` (for example, there are tests that do use routes with different HTTP methods in order to avoid the overhead of creating separate router methods for minor differences).

## Implementation details
The idea behind the implementation is to create a trie of all the routes components (meaning the HTTP method and the URL components which can be "fixed" or which can represent Werkzeug path parameters) for each router method. Using this trie, we can find all the routes that share a common "stem" path and add optional path arguments in its continuation.

The main idea behind the problem of finding these compatible routes that can be reduced to one is that **sequences of consecutive terminal path args can be reduced to a single route with optional path args**. By terminal I mean that the path arg is the last component in a route associated with the router method.

In the following lines I use * after a component to note the fact that there is a route that ends with that component.

### Examples
  
  Example 1: `_prefix_/fixed*/arg1*/arg2*/arg3*` can be reduced to the stem `_prefix_/fixed*` with:
    - required path arguments = `extract_path_arguments(_prefix_`)
    - optional path arguments = [arg1, arg2, arg3].
  
  Example 2: `_prefix_/arg1*/arg2*/arg3*` can be reduced to the stem `_prefix_/arg1*` with:
    - required path arguments = `extract_path_arguments(_prefix_`) + [arg1]
    - optional path arguments = [arg2, arg3].

### Notes

Note 1: If there is a terminal component, followed by an unterminal path argument and a terminal path argument, we can't reduce them:
  ex.: `_prefix_/fixed1*_or_arg1*/arg2/arg3*` will result in 2 routes: stem `_prefix_/fixed1*_or_arg1*` and stem `_prefix_/fixed1*_or_arg1*/arg2/arg3*` and not stem `_prefix_/fixed1*_or_arg1*` with optional path args [arg2, arg3] because arg2 can only appear if arg3 appears.

Note 2: Whenever there is a terminal fixed component, a new route is defined (which may or may not have optional path arguments that follow to its right):
  ex.1: `_prefix_/fixed1*/fixed2*` will result in 2 routes: `_prefix_/fixed1*` and `_prefix_/fixed2*`
  ex.2: `_prefix_/fixed1*/arg1*/arg2*/fixed3*/arg3*` will result in 2 routes:
        - stem `_prefix_/fixed1*` with optional path args = [arg1, arg2], required path args = `extract_path_arguments(_prefix_`)
        - stem `_prefix_/fixed1/arg1/arg2/fixed3*` with optional args = [arg3], required path args = `extract_path_arguments(_prefix_`) + [arg1, arg2]

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -1081,7 +1081,7 @@ def _GroupRoutesByStem(
 
 def _GetGroupedRoutes(routes: List[List[str]]) -> List[RouteInfo]:
   comps_trie_root = _CreateTrie(routes)
-  grouped_routes_stems = dict()
+  grouped_routes_stems: Dict[str, Dict[str, List[ComponentTrieNode]]] = dict()
   _GroupRoutesByStem(comps_trie_root, [], None, grouped_routes_stems)
 
   grouped_routes = []
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -742,7 +742,7 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     paths_obj: DefaultDict[str, Dict] = collections.defaultdict(dict)
 
     router_methods = self.router.__class__.GetAnnotatedMethods()
-    for router_method in router_methods:
+    for router_method in router_methods.values():
       # To extract optional path arguments, all the routes associated with this
       # router method must be analysed and grouped.
       ungrouped_routes = []
@@ -781,11 +781,9 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
           elif path_param.name in opt_path_params_names:
             opt_path_params.append(path_param)
           else:
-            # TODO(alexandrucosminmihai): Simplify exception message.
             raise AssertionError(
-              f"Path parameter {path_param.name} not found among required path "
-              f"parameters: {req_path_params_names}, nor among optional path "
-              f"parameters: {opt_path_params_names}."
+              f"Path parameter {path_param.name} was not classified as "
+              f"required/optional."
             )
 
         normalized_path = _NormalizePath(path)
@@ -1010,7 +1008,10 @@ def _IsMapField(field_descriptor: FieldDescriptor) -> bool:
 class ComponentTrieNode:
   def __init__(self, component, parent_path):
     self.component = component
-    self.path = f"{parent_path}/{component}"
+    if parent_path:
+      self.path = f"{parent_path}/{component}"
+    else:
+      self.path = component
     self.is_path_arg = component.startswith("<") and component.endswith(">")
     self.is_route_end = False
     self.children = dict()
@@ -1068,7 +1069,7 @@ def _GroupRoutesByStem(
       "optional": [],
     }
 
-  for child in curr_node.children:
+  for child in curr_node.children.values():
     _GroupRoutesByStem(child, path_args, new_stem_node, grouped_routes_stems)
 
   # We'll go back to the parent of this node, so we remove the current path arg
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -1007,7 +1007,11 @@ def _IsMapField(field_descriptor: FieldDescriptor) -> bool:
 
 
 class ComponentTrieNode:
-  def __init__(self, component, parent_path):
+  def __init__(
+      self,
+      component: str,
+      parent_path: str,
+  ) -> None:
     self.component = component
     if parent_path:
       self.path = f"{parent_path}/{component}"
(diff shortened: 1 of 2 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -496,21 +497,25 @@ def _GetDescribedSchema(
 
   def _GetParameters(
       self,
-      path_params: List[FieldDescriptor],
+      required_path_params: List[FieldDescriptor],
+      optional_path_params: List[FieldDescriptor],
```
panhania, line 501: I don't think you care about this being a list, `Iterable` should be enough. Try to be as liberal in parameter type as possible.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -692,6 +697,10 @@ def _GetOperationDescription(
       http_method: str,
       path: str,
       router_method: Any,
+      required_path_params: List[FieldDescriptor],
+      optional_path_params: List[FieldDescriptor],
+      query_params: List[FieldDescriptor],
+      body_params: List[FieldDescriptor],
```
panhania, line 703: Ditto.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -952,3 +1003,116 @@ def _GetMapFieldKeyValueTypes(
 def _IsMapField(field_descriptor: FieldDescriptor) -> bool:
   """Checks that a `FieldDescriptor` is of a map type."""
   return _GetMapFieldKeyValueTypes(field_descriptor) is not None
+
+
+class ComponentTrieNode:
+  def __init__(self, component, parent_path):
```
panhania, line 1009: Missing type annotations.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -952,3 +1003,116 @@ def _GetMapFieldKeyValueTypes(
...
+    if parent_path:
+      self.path = f"{parent_path}/{component}"
+    else:
+      self.path = component
+    self.is_path_arg = component.startswith("<") and component.endswith(">")
+    self.is_route_end = False
+    self.children = dict()
+
+
+def _CreateTrie(routes: List[List[str]]) -> ComponentTrieNode:
```
panhania, line 1020: I would make it a pseudo-constructor (a `@classmethod` that returns a class instance) called `FromRoutes`.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -952,3 +1003,116 @@ def _GetMapFieldKeyValueTypes(
...
+    routes: A list of routes where each route is represented as a list where
+      the first element is a string representing the HTTP method of the route
+      and the following elements are strings representing the path components.
+
+  Returns:
+    A list of `RouteInfo`s which are tuples consisting of three elements: a list
+    with the HTTP method and the components of the largest path that covers
+    multiple of the initial routes given as argument, a list of path components
+    which represent the required path arguments and a list of path components
+    which represent the optional path arguments.
```
panhania, line 1099: This shows the problem with using tuples. If you were to use a class with named fields (e.g. through `NamedTuple`), you could put these descriptions to appropriate fields making it much more obvious and better to read.

In general: avoid tuples at all costs. I don't see any value in using tuples in any production-ready code. They can be useful for quick run-and-forget scripts, though.

panhania (changes requested):

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Noted. Thank you for pointing this out, I made the change!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Oops, I must have missed it. Thank you!

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Made the change. Thank you! A note about the `curr_node` local variable: I annotated it as a `ComponentTrieNode` so I can keep `ComponentTrieNode.children` as a dictionary of `ComponentTrieNode`. This way, if we ever subclass ComponentTrieNode, the `FromRoutes` class method would still work, while allowing tries with nodes of different subclasses of `ComponentTrieNode` to exist. I don't know how useful/probable this is to happen, but I believe this annotation makes sense. Please let me know if you think differently.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: You're right. I converted `RouteInfo` to be a `NamedTuple` instead of a type alias. Thank you!

mbushkov (commented):
There's some impressive type annotation work in this PR - had to go to stackoverflow to understand why `@classmethods` are annotated this way.
Overall - looks good to me - will defer to @panhania to formally LGTM.

panhania (approved):
Huh, I thought I merged this already...

PR #839, opened 2020-09-03
Entry history components (part 1)

danielberbece:
Introducing `entry-history-button` and `entry-history-dialog` components, to provide individual overview of the changes in a client details entry. This PR doesn't support showing history for the user/interface/volume list objects, because they will be added in a future PR. The source code also contains a few `TODO` comments related to the ClientDetailsStore implementation, which will be available soon.


grr/server/grr_response_server/gui/ui/components/client_details/client_details.ts
```diff
@@ -20,8 +20,9 @@ export class ClientDetails {
   readonly INITIAL_NUM_VOLUMES_SHOWN = 2;
 
   readonly client$ = this.clientPageFacade.selectedClient$;
+
+  // TODO Move this to ClientDetailsStore
```
max-vogler, line 24: Use `TODO(danielberbece): Move this to ClientDetailsStore.`

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
 }
+
+function getFirstStringsJoinedPath(path: any[]): string {
+  let tokens: string[] = [];
+  path.forEach(val => {
+    if (typeof val !== 'string') return;
+    tokens.push(val);
+  });
+
+  return tokens.join('.');
```
max-vogler, line 213: Use functional programming for better readability:

```
return path.filter(val => typeof val === 'string').join('.')
```

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
 
   return clientChanges;
 }
+
+function getFirstStringsJoinedPath(path: any[]): string {
```
max-vogler, line 206: Don't use `any`, specify more precise types.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+
+  differences.forEach(diffItem => {
+    if (diffItem.path === undefined) return;
+    const joinedPath = getFirstStringsJoinedPath(diffItem.path);
+    if (!RELEVANT_ENTRIES_LABEL_MAP.has(joinedPath)) return;
+
+    pathsSetOfChangedEntries.add(joinedPath);
+  });
+
+  return [...pathsSetOfChangedEntries.values()];
```
max-vogler, line 228: Use `Array.from(pathsSetOfChangedEntries)`.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+    if (typeof val !== 'string') return;
+    tokens.push(val);
+  });
+
+  return tokens.join('.');
+}
+
+function getPathsOfChangedEntries(differences: Diff<Client, Client>[]):
+    string[] {
+  const pathsSetOfChangedEntries: Set<string> = new Set();
```
max-vogler, line 218: This is a complex name that also contains the type (set), which is redundant. Why not use `changedPaths`?

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+
+  differences.forEach(diffItem => {
+    if (diffItem.path === undefined) return;
+    const joinedPath = getFirstStringsJoinedPath(diffItem.path);
+    if (!RELEVANT_ENTRIES_LABEL_MAP.has(joinedPath)) return;
+
+    pathsSetOfChangedEntries.add(joinedPath);
```
max-vogler, line 225: Use functional programming instead of side-effects for better readability.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+  path.forEach(val => {
+    if (typeof val !== 'string') return;
+    tokens.push(val);
+  });
+
+  return tokens.join('.');
+}
+
+function getPathsOfChangedEntries(differences: Diff<Client, Client>[]):
+    string[] {
```
max-vogler, line 217: Use `ReadonlyArray<string>` as return type - we prefer immutability.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+
+  return [...pathsSetOfChangedEntries.values()];
+}
+
+/**
+ * Returns relevant entry paths mapped to the client snapshots in which the
+ * changes were introduced, reverse chronologically ordered. A property path may
+ * not be inside the map if no changes were applied to it since it's creation
+ * @param clientSnapshots an array of chronologically reverse ordered client
+ *     snapshots
```
max-vogler, line 236: Provide a one-liner description of what the function does in the first line of JSDoc. Then after a blank line possible further explanations.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -202,3 +202,59 @@ export function getClientVersions(clientSnapshots: Client[]):
...
+ */
+export function getClientEntriesChanged(clientSnapshots: Client[]):
+    Map<string, ReadonlyArray<Client>> {
+  const clientChangedEntries: Map<string, Client[]> = new Map();
+
+  for (let i = 0; i < clientSnapshots.length - 1; i++) {
+    const differences = diff(clientSnapshots[i + 1], clientSnapshots[i]);
```
max-vogler, line 243: This accesses `clientSnapshots[clientSnapshots.length]` and will return `undefined`. Is this intended? In general, this code mixes low-level iteration logic and high-level diffing logic. I suggest extracting the logic for pairwise iteration into a generic helper function, e.g. `function pairwise<T>(arr: T[]) => ReadonlyArray<[T, T]>`.

Also, use functional programming like `filter(difference => difference !== undefined)` over branching and side-effects.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,54 @@
+import {Component, Inject} from '@angular/core';
+import {MAT_DIALOG_DATA} from '@angular/material/dialog';
+import {Client} from '@app/lib/models/client';
+
+export type EntryType =
```
max-vogler, line 5: Provide JSDoc for all exported members.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,54 @@
...
+    'primitive'|'timestamp'|'size'|'user-list'|'interface-list'|'volume-list';
+
+export interface EntryHistoryDialogParams {
+  path: string;
+  type: EntryType;
+  clientVersions: Client[];
```
max-vogler, line 11: In your interface definitions, make all properties immutable if possible: Use `readonly` and `ReadonlyArray`.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+interface EntryHistoryTableRow {
+  time: Date;
+  version: any;
```
max-vogler, line 16: Don't use `any` - use more precise types.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  private initTableRows(data: EntryHistoryDialogParams) {
+    data.clientVersions.forEach((client) => {
+      let property: any = client;
+      data.path.split('.').forEach((token) => {
+        property = property[token];
+      });
+
+      if (property === undefined) {
+        throw new Error('Wrong "path" provided');
```
max-vogler, line 45: This error message is not helpful. Include `data` in the error to make debugging easier.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+
+      this.tableRows?.push({
```
max-vogler, line 48: Remove `?`.

grr/server/grr_response_server/gui/ui/components/client_details/module.ts
```diff
@@ -11,6 +12,8 @@ import {HumanReadableSizeModule} from '../human_readable_size/module';
 import {TimestampModule} from '../timestamp/module';
 
 import {ClientDetails} from './client_details';
+import {EntryHistoryButton} from './entry_history_button/entry_history_button';
+import {EntryHistoryDialog} from './entry_history_dialog/entry_history_dialog';
```
max-vogler, line 16: Create an own module for EntryHistoryButton/Dialog.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -231,6 +231,9 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
       filter(
           (clientVersions): clientVersions is Client[] =>
               clientVersions !== undefined),
+      // Reverse snapshots to provide reverse chronological order
+      map(snapshots => snapshots.map(
+              (item, idx) => snapshots[snapshots.length - 1 - idx])),
```
max-vogler, line 236: Use a function from the standard library or lodash to reverse.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
danielberbece: I updated it in the second part: #842

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
danielberbece: I agree that having `any` as type is not preferred. Here `version` could be any kind of object (Date, string, number, StorageVolume, NetworkInterface, etc.). Having a custom type for this is more rigid and must be changed when a new custom type will define the property. Should we change anyway?

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
danielberbece: `any` is the type of the provided argument, always. This is coming from the `deep-diff` implementation. Is there a workaround for this type of situation?

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
danielberbece: This method implementation was changed in the following PR: #842

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
danielberbece: The `i` variable will take values in the range `[0, clientSnapshots.length - 2]`, because of the condition inside the for statement: `i < clientSnapshots.length - 1`. This means that `i + 1` will be in the range `[1, clientSnapshots.length - 1]`. Knowing this, I don't understand how it accesses `clientSnapshots[clientSnapshots.length]`. On the other hand, the `pairwise` helper function is a good idea, I will make the changes.

grr/server/grr_response_server/gui/ui/components/client_details/module.ts
danielberbece: I believe it would be better to keep them inside the `ClientDetailsModule`, as they are pretty related and not used elsewhere. Also, in the next PR there are 3 more components that are used in the `ClientDetails`, as well as in the `EntryHistoryDialog`. Where would they fit?

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
danielberbece: I updated the implementation here as well.

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
max-vogler: Fair. Please add a comment on the preceeding line:
```
// tslint:disable-next-line:no-any Provided by deep-diff typing.
```

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
max-vogler: Use generics:
```
interface EntryHistoryTableRow<T> {
  ....
  version: T;
```

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
```diff
@@ -86,6 +87,7 @@ describe('Entry History Dialog', () => {
 
     expect(component.tableRows).toEqual(expectedTableRows);
 
+    // @ts-ignore
```
max-vogler, line 90: Why do we need `@ts-ignore`?

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
```diff
@@ -205,51 +205,67 @@ export function getClientVersions(clientSnapshots: Client[]):
...
   let tokens: string[] = [];
-  path.forEach(val => {
-    if (typeof val !== 'string') return;
-    tokens.push(val);
-  });
+
+  for (let i = 0; i < path.length; i++) {
+    if (typeof path[i] !== 'string') break;
+    tokens.push(path[i]);
+  }
```
max-vogler, line 212: Please explain why you stop appending after the first non-string. What is a typical non-string value here? Why are strings after the non-string not relevant?

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
danielberbece: Because the property is `readonly`, changing it after it was assigned would signal an error, so I had to add `@ts-ignore` to bypass it. I thought about creating the whole object right there, but I would have to also configure the testing module. Do you think this would be a better approach, in order to inject the data?

grr/server/grr_response_server/gui/ui/components/client_details/module.ts
max-vogler: Our build system really pushes to having one module per folder. It's your choice to either keep EntryHistoryDialog in the sub-folder and add a module, or move the files one folder up and have them in the parent module.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
max-vogler: Please avoid disabling the error handling with @ts-ignore. This violates basic invariances (readonly attributes are never overwritten) and is unexpected for the person reading the code. Please refactor the test to create new objects and inject them.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/client_details/client_diff.ts
danielberbece: A typical non-string element would be a number [0,1,..n] for array access, because `deep-diff` can find differences between individual elements of an array. Unfortunately, showing detailed versions of array individual elements is not currently supported by us in the UI. In case the path contains a number, it means that the algorithm found a difference within an array, so we truncate it right there to obtain only the path to the array object. That way we can gather all the changes within the array under a bigger umbrella, like `List changes`. That's why we need the trucated path.

max-vogler: That's super insightful and important information. Please put that in comments.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
```diff
@@ -57,66 +58,97 @@ describe('Entry History Dialog', () => {
...
     })();
-  }));
-
-  afterEach(() => {
-    overlayContainer.ngOnDestroy();
-  });
+  }
 
   it('is created successfully', () => {
+    configureTestingModule({
```
max-vogler, line 68: Is it possible to still call configureTestingModule in beforeEach, but let it inject EntryHistoryDialogParams using `useFactory: () => providedData` and change `providedData` in each test function before `TestBed.createComponent` is called?

If the injection is only executed when TestBed.createComponent is called, this would still allow setting up the TestBed in beforeEach.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -28,7 +28,7 @@ interface EntryHistoryTableRow {
 })
 export class EntryHistoryDialog {
   readonly entryType: EntryType;
-  readonly tableRows: EntryHistoryTableRow[] = [];
+  readonly tableRows: EntryHistoryTableRow<any>[] = [];
```
max-vogler, line 31: I see that looping through generics everywhere might be out of scope for this PR. Please add a comment that disables ts-lint's any detection for this line.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
danielberbece: Trying that resulted in this error. I tried searching for a solution, but wasn't very lucky.
```
Error: Cannot call Promise.then from within a sync test.
    at SyncTestZoneSpec.push../node_modules/zone.js/dist/zone-testing.js.SyncTestZoneSpec.onScheduleTask (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-testing.js:368:1)
    at ZoneDelegate.scheduleTask (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:378:1)
    at Zone.scheduleTask (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:210:1)
    at Zone.scheduleMicroTask (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:230:1)
    at scheduleResolveOrReject (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:847:1)
    at ZoneAwarePromise.then (http://localhost:9876/_karma_webpack_/node_modules/zone.js/dist/zone-evergreen.js:979:1)
    at ApplicationInitStatus.runInitializers (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/core.js:40425:1)
    at R3TestBedCompiler.finalize (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/testing.js:1898:82)
    at TestBedRender3.get testModuleRef [as testModuleRef] (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/testing.js:3253:1)
    at TestBedRender3.inject (http://localhost:9876/_karma_webpack_/node_modules/@angular/core/__ivy_ngcc__/fesm2015/testing.js:3110:1)
```

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
panhania: Is it though? It is newly added code and I don't think there is a good reason not make it properly typed from the very beginning.

Here, the `EntryHistoryDialog` can hold only is expected to hold history of a particular type. It seems easy enough to make it generic as well.

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
```diff
@@ -0,0 +1,60 @@
...
+import {MAT_DIALOG_DATA} from '@angular/material/dialog';
+import {Client} from '@app/lib/models/client';
+
+/** Entry type */
+export type EntryType =
+    'primitive'|'timestamp'|'size'|'user-list'|'interface-list'|'volume-list';
+
+/** Parameters required to open an EntryHistoryDialog */
+export interface EntryHistoryDialogParams {
+  readonly path: string;
```
panhania, line 11: Judging from the code that uses it (`data.path.split('.')`) this is some list of components. Is there a reason to store it as a string rather than using a more structured way from the very beginning as a list of components?

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog_test.ts
max-vogler: Can you post a gist with the code that triggers the error?

danielberbece: I managed to find the issue and solve it

grr/server/grr_response_server/gui/ui/components/client_details/entry_history_dialog/entry_history_dialog.ts
danielberbece: I agree it is not preferred to use `any` as type. In order to use a specific type for the array, we could use a generic type for the `EntryHistoryDialog`, just like you suggested. Unfortunately, Angular doesn't consider the generic type if we do that, and it would be of no use. I am not able to find a solution for this problem and if you or @max-vogler have any other suggestions, please reach out.

panhania: Discussed offline. For now, seems like we have to leave it as it is.

panhania (approved):

PR #843, opened 2020-09-09
Extract type from protobuf semantic field option and use it in OpenAPI description

alexandrucosminmihai:
In this PR I extract the [`SemanticDescriptor` `sem_type`](https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/grr/proto/grr_response_proto/semantic.proto#L8) protobuf field option and use its `type` field to describe message fields and parameters in the generated OpenAPI description of the GRR API.

I use a dictionary of supported RDF types for which we have an OpenAPI schema including a description which specifies what that type actually represents. The RDF types included are all the ones defined in [`grr_response_core/lib/rdfvalue.py`](https://github.com/google/grr/blob/master/grr/core/grr_response_core/lib/rdfvalue.py) and [`grr_response_core/lib/rdfvalues/paths.py`](https://github.com/google/grr/blob/master/grr/core/grr_response_core/lib/rdfvalues/paths.py) which are used at least once as the `type` in a `sem_type` protobuf field option. Please let me know if any types should be added or removed.

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -137,59 +137,49 @@ rdf_types_schemas: Dict[str, Schema] = {
   "RDFDatetime": {
     "type": "string",
     "format": "uint64",
-    "description": "RDF type is `RDFDatetime` and it represents "
-                   "the number of microseconds since epoch to a timestamp.",
+    "description": "the number of microseconds since epoch to a timestamp.",
   },
   "RDFDatetimeSeconds": {
     "type": "string",
     "format": "uint64",
-    "description": "RDF type is `RDFDatetimeSeconds` and it represents "
-                   "the number of seconds since epoch to a timestamp.",
+    "description": "the number of seconds since epoch to a timestamp.",
   },
   "Duration": {
     "type": "string",
     "format": "uint64",
-    "description": "RDF type is `Duration` and it represents "
-                   "the number of microseconds between two timestamps.",
+    "description": "the number of microseconds between two timestamps.",
   },
   "DurationSeconds": {
     "type": "string",
     "format": "uint64",
-    "description": "RDF type is `DurationSeconds` and it represents "
-                   "the number of seconds between two timestamps.",
+    "description": "the number of seconds between two timestamps.",
   },
   "RDFBytes": {
     "type": "string",
     "format": "byte",
-    "description": "RDF type is `RDFBytes` and it represents "
-                   "a buffer of bytes.",
+    "description": "a buffer of bytes.",
   },
   "HashDigest": {
     "type": "string",
     "format": "byte",
-    "description": "RDF type is `HashDigest` and it represents "
-                   "a binary hash digest with hex string representation.",
+    "description": "a binary hash digest with hex string representation.",
   },
   "GlobExpression": {
     "type": "string",
-    "description": "RDF type is `GlobExpression` and it represents "
-                   "a glob expression for a client path.",
+    "description": "a glob expression for a client path.",
   },
   "ByteSize": {
     "type": "string",
     "format": "uint64",
-    "description": "RDF type is `ByteSize` and it represents "
-                   "a size for bytes allowing standard unit prefixes.",
+    "description": "a size for bytes allowing standard unit prefixes.",
   },
   "RDFURN": {
     "type": "string",
-    "description": "RDF type is `RDFURN` and it represents "
-                   "an object to abstract URL manipulation.",
+    "description": "an object to abstract URL manipulation.",
   },
   "SessionID": {
     "type": "string",
-    "description": "RDF type is `SessionID` and it represents "
-                   "an rdfvalue object that represents a session_id.",
+    "description": "an rdfvalue object that represents a session_id.",
   },
 }
 
@@ -262,6 +252,15 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
     """Adds the OpenAPI schemas for RDF types."""
     if self.schema_objs is None:
       raise AssertionError("OpenAPI type schemas not initialized.")
+
+    for rdf_type_name in rdf_types_schemas:
+      old_description = rdf_types_schemas[rdf_type_name]['description']
+      new_prefix = f"RDF type is `{rdf_type_name}` and it represents"
+
+      if not old_description.startswith(new_prefix):
+        rdf_types_schemas[rdf_type_name]["description"] = (
+          f"{new_prefix} {old_description}"
+        )
 
     self.schema_objs.update(rdf_types_schemas)
 
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -137,49 +137,49 @@ rdf_types_schemas: Dict[str, Schema] = {
   "RDFDatetime": {
     "type": "string",
     "format": "uint64",
-    "description": "the number of microseconds since epoch to a timestamp.",
+    "description": "the number of microseconds since epoch to a timestamp",
   },
   "RDFDatetimeSeconds": {
     "type": "string",
     "format": "uint64",
-    "description": "the number of seconds since epoch to a timestamp.",
+    "description": "the number of seconds since epoch to a timestamp",
   },
   "Duration": {
     "type": "string",
     "format": "uint64",
-    "description": "the number of microseconds between two timestamps.",
+    "description": "the number of microseconds between two timestamps",
   },
   "DurationSeconds": {
     "type": "string",
     "format": "uint64",
-    "description": "the number of seconds between two timestamps.",
+    "description": "the number of seconds between two timestamps",
   },
   "RDFBytes": {
     "type": "string",
     "format": "byte",
-    "description": "a buffer of bytes.",
+    "description": "a buffer of bytes",
   },
   "HashDigest": {
     "type": "string",
     "format": "byte",
-    "description": "a binary hash digest with hex string representation.",
+    "description": "a binary hash digest with hex string representation",
   },
   "GlobExpression": {
     "type": "string",
-    "description": "a glob expression for a client path.",
+    "description": "a glob expression for a client path",
   },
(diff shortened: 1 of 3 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -132,6 +133,66 @@ class RouteInfo(NamedTuple):
   },
 }
 
+rdf_types_schemas: Dict[str, Schema] = {
```
panhania, line 136: I am not a native speaker but double pluralization sounds awkward to me. Shouldn't it be `rdf_type_schemas`?

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -132,6 +133,66 @@ class RouteInfo(NamedTuple):
...
+  "RDFDatetime": {
+    "type": "string",
+    "format": "uint64",
+    "description": "RDF type is `RDFDatetime` and it represents "
```
panhania, line 140: I find this _"RDF type is `X` and it represents Y"_ to be quite cumbersome. Why not just _"Y"_ and move the prefix to where you actually generate the message?

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -197,6 +258,13 @@ def _AddPrimitiveTypesSchemas(self) -> None:
       cast(Dict[str, Dict[str, str]], primitive_types_schemas)
     )
 
+  def _AddRDFTypesSchemas(self) -> None:
+    """Adds the OpenAPI schemas for RDF types."""
+    if self.schema_objs is None:
+      raise AssertionError("OpenAPI type schemas not initialized.")
+
+    self.schema_objs.update(rdf_types_schemas)
```
panhania, line 266: This has more assertion logic than actual content. I think it is a right thing to do in this PR for consistency with the other initialization method, but I would keep in mind for the future to refactor these two (`schema_objs` is pretty static IIUC, why not make it a top-level constant)?

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -474,15 +543,17 @@ def _GetDescribedSchema(
         f"Only one field per oneof should be present."
       )
 
-    # `protobuf.map` related description.
-    if _IsMapField(field_descriptor):
-      if description:
-        description += " "
+    # Add the description of the field's type which is stored with the type's
+    # schema. Examples of such descriptions are the ones of RDFTypes, the ones
```
panhania, line 547: "RDF types"

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I am not a native speaker either, but the way I thought about it when I named the variable was that the dictionary will hold different `schemas` for different `rdf_types`, whereas `rdf_type_schemas` sounds to me like a structure for multiple `schemas` associated with the same, single `rdf_type`. Please let me know if I should change it.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you for pointing this out! I will refactor them in a future PR. 

Regarding `schema_objs`, I kept it tied to the handler instance because it depends on the api router which is passed as an argument to the handler's constructor and from which the router method's args types and result types are traversed during schemas creation. I believe, but I'm not completely sure, that multiple routers can be used "at the same time" for requests from different users from what I've read in `http_api.py` [[1]](https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/grr/server/grr_response_server/gui/http_api.py#L423) [[2]](https://github.com/google/grr/blob/b472626b6ff4d05b24269a7ce52419d00a847607/grr/server/grr_response_server/gui/http_api.py#L191). If we had a single, top-level `schema_objs`, the OpenAPI description might expose too many/few API methods depending on the user maybe? Please let me know if I should move it to top-level.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you! I updated it.

grr/server/grr_response_server/gui/api_plugins/metadata.py
panhania: Maybe @bgalehouse can provide some expertise here.

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: Thank you! I moved the prefix mentioning RDF type name in `_AddRDFTypesSchemas` so it can be added to the OpenAPI components descriptions as well, not only to the description of fields that have this type.

grr/server/grr_response_server/gui/api_plugins/metadata.py
bgalehouse: I think you want rdf_type_schemas.

It is a semi-corner case in english that nouns used as adjectives lose any plurality marker, even if the noun would otherwise be trivially plural.  "5 second rule", "2 foot drop", _not_ "5 seconds rule" or "2 feet drop".

alexandrucosminmihai: Thank you @panhania and @bgalehouse ! There are multiple instances of this wrong naming scheme, even for parts of code outside of this PR so I think I should create a separate PR fixing all occurrences. Please let me know if I should change this specific one in this PR.

panhania: I think it is fine to do it in a separate PR.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -197,6 +248,22 @@ def _AddPrimitiveTypesSchemas(self) -> None:
...
+      raise AssertionError("OpenAPI type schemas not initialized.")
+
+    for rdf_type_name in rdf_types_schemas:
+      old_description = rdf_types_schemas[rdf_type_name]['description']
+      new_prefix = f"RDF type is `{rdf_type_name}` and it represents"
+
+      if not old_description.startswith(new_prefix):
+        rdf_types_schemas[rdf_type_name]["description"] = (
+          f"{new_prefix} {old_description}"
+        )
```
panhania, line 263: Modifying a global variable is never a good idea—here it is even worse, because you get incorrect descriptions if someone instantiates `ApiGetOpenApiDescriptionHandler` multiple times.

grr/server/grr_response_server/gui/api_plugins/metadata.py
bgalehouse: BTW, my suggestion gives the impression that you a number of schemas, each of which is connected/related to one or more rdf_type values - each entry could be called an rdf_type schema.

If the schemas and the rdf_types just happen to be attached, something like rdf_type_to_schema could work as a name for a dict/map.

grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -474,15 +542,17 @@ def _GetDescribedSchema(
...
+      type_description = type_schema.get("description")
+      if type_description is not None:
+        if description:
+          description += " "
 
-      map_type_schema = self.schema_objs[type_name]
-      description += cast(
-        str, map_type_schema.get("description", "")
-      )
+        description += cast(str, type_description)
```
panhania, line 555: This is where the description generation happens.

panhania (changes requested):

grr/server/grr_response_server/gui/api_plugins/metadata.py
alexandrucosminmihai: I agree, it's not a clean solution, but there are some details about this. Initially I added it, as you pointed out, in `_GetDescribedSchema`, but the problem is that that function is called only for parameters and fields, whereas this description is strictly type related, not field/parmeter related (compared to `protobuf.oneof` descriptions) and this would cause that the `description` field in the definitions of the schemas of the types in the `components` field of the resulting JSON to be different from what is displayed for fields, altough the description is related to the type. The verification for the prefix should solve the problem about multiple instantiations (an alternative would be to use a global boolean flag). But if there are multiple threads, there will be a race condition. Will there be multiple threads accessing the same memory?

On the other hand, having different descriptions for the component definition (which would lack the prefix) and for field/parameter definitions is not necessarily a bad thing since the OpenAPI component definition includes the name of the type.

Please let me know if I should move forward with different descriptions for the component definition (the actual definition of the type's schema) and the fields/parameters (where a reference to the type's schema is used and the description is overriden).

panhania: > I agree, it's not a clean solution. (...)

There are two jarring problems with the current approach.

The first one is modifying a global constant. A global constant should be a global constant, it should never change. It is problematic not only because of multi-threading issues, but also because some parts of code might make assumptions about it and the bugs are then very hard to track.

The other one is mixing the data and presentation. Here, your data is the type, the format and a short description. Combining the three to present to the user is not something that should be ever persisted. When needed, you should take the data, generate a user-facing string, give it to the user and forget it. You have all the data you need to re-create it if needed, don't waste the memory. Imagine that you would like to add internationalization to your code. If you mix data and presentation, it would be just impossible.

> But if there are multiple threads, there will be a race condition. Will there be multiple threads accessing the same memory?

It's a wrong question. Requirements change. Maybe it isn't now (but maybe it is, request handlers is something that I imagine is the most basic parallelization unit for web servers), but what if it will become in the future? I doubt anyone will remember that there is some piece of code that assumed that there is no multi-threading and it should be adjusted now.

> On the other hand, having different descriptions for the component definition (which would lack the prefix) and for field/parameter definitions is not necessarily a bad thing

Then let's do it for now and adjust if there is a need for it.

alexandrucosminmihai: Thank you for the clear explanation! I made the change.

panhania (approved):

PR #847, opened 2020-09-10
Introduce `casing` module

alexandrucosminmihai:
Implement `SnakeToCamel` and `CamelToSnake` functions in a new `casing` module, as well as tests for them.


grr/core/grr_response_core/lib/casing_test.py
```diff
@@ -0,0 +1,32 @@
+# Lint as: python3
+"""Tests for casing functions."""
+
+from grr_response_core.lib import casing
+
+from grr.test_lib import test_lib
+
+
+class CasingTests(test_lib.GRRBaseTest):
```
panhania, line 9: `GRRBaseTest` is evil, `absltest.TestCase` is the way to go unless your test has no control over its dependencies and requires a tons of GRR-specific setup.

panhania, line 9: There should be a test class per class or function. In your case, it should be `SnakeToCamelTest` and `CamelToSnakeTest`.

grr/core/grr_response_core/lib/casing_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+  def testSnakeToCamelWorksOnOneWordStrings(self):
+    self.assertEqual("abcd", casing.SnakeToCamel("abcd"))
+    self.assertEqual("a", casing.SnakeToCamel("a"))
+
+  def testSnakeToCamelWorksOnRegularStrings(self):
+    self.assertEqual("thisIsACamel", casing.SnakeToCamel("this_is_a_camel"))
+    self.assertEqual("aCamelThisIs", casing.SnakeToCamel("a_camel_this_is"))
+    self.assertEqual("aBCD", casing.SnakeToCamel("a_b_c_d"))
+
```
panhania, line 21: I would also test scenarios on strings with:

  * double underscores
  * underscore prefixes and suffixes
  * only underscores
  * uppercase characters
  * unicode characters

grr/core/grr_response_core/lib/casing.py
```diff
@@ -0,0 +1,25 @@
...
+
+def SnakeToCamel(snake_str: str) -> str:
+  """Convert a string in snake_case to lowerCamelCase."""
+  if not snake_str:
+    return snake_str
+
+  snake_components = snake_str.split("_")
+  lower_camel_str = (
+    f"{snake_components[0][:1].lower() + snake_components[0][1:]}"
+    f"{''.join([comp.capitalize() for comp in snake_components[1:]])}"
```
panhania, line 14: `map(str.capitalize, snake_components[1:])`

panhania, line 14: `f"{str_x + str_y}{str_z}"` looks awkward. Why not just `str_x + str_y + str_z`?

grr/core/grr_response_core/lib/casing.py
```diff
@@ -0,0 +1,25 @@
+# Lint as: python3
+"""Functions to convert strings between different case styles."""
+import re
+
+
+def SnakeToCamel(snake_str: str) -> str:
+  """Convert a string in snake_case to lowerCamelCase."""
```
panhania, line 7: I would make a note that these functions assume that the given string is a single identifier and the behaviour for string with spaces is undefined.

grr/core/grr_response_core/lib/casing_test.py
```diff
@@ -0,0 +1,32 @@
...
+    self.assertEqual("", casing.CamelToSnake(""))
+
+  def testCamelToSnakeWorksOnOneWordStrings(self):
+    self.assertEqual("abcd", casing.CamelToSnake("abcd"))
+    self.assertEqual("a", casing.CamelToSnake("a"))
+
+  def testCamelToSnakeWorksOnRegularStrings(self):
+    self.assertEqual("this_is_a_snake", casing.CamelToSnake("thisIsASnake"))
+    self.assertEqual("a_snake_this_is", casing.CamelToSnake("aSnakeThisIs"))
+    self.assertEqual("a_b_c_d", casing.CamelToSnake("aBCD"))
```
panhania, line 32: Similarly here, I would add tests for:

  * only uppercase characters
  * unicode characters (especially uppercase unicode characters; you can fine some especially beautiful examples [here](https://en.wikipedia.org/wiki/Polish_alphabet))
  * underscores

panhania (changes requested):

grr/core/grr_response_core/lib/casing_test.py
alexandrucosminmihai: Ow, thank you pointing this out! I made the change here and I'll keep it in mind in the future.

alexandrucosminmihai: Thank you! I split the tests into the two new test classes.

grr/core/grr_response_core/lib/casing_test.py
alexandrucosminmihai: Thank you for this suggestion! I implemented the new tests.

grr/core/grr_response_core/lib/casing.py
alexandrucosminmihai: Thank you for the functional suggestion! I made the change.

alexandrucosminmihai: The code changed a bit, but I still made the transition to `str + str`. Thank you!

grr/core/grr_response_core/lib/casing_test.py
alexandrucosminmihai: I added the new tests. Thank you for the suggestion. Please let me know if you have suggestions for other tests.

grr/core/grr_response_core/lib/casing.py
```diff
@@ -0,0 +1,73 @@
...
+  # Extract the words from the snake_case string.
+  words = list(filter(lambda word: len(word) > 0, snake_str.split("_")))
+  if not words:
+    return ""
+
+  words[:] = map(str.lower, words)
+
+  lower_camel_str = words[0] + ''.join(map(str.capitalize, words[1:]))
+
+  return lower_camel_str
```
panhania, line 28: `lower_camel_str` is kind of useless, why not just `return`?

grr/core/grr_response_core/lib/casing.py
```diff
@@ -0,0 +1,73 @@
...
+
+  temp_snake_str = "_".join(map(str.lower, words))
+
+  # Remove duplicate underscores, if any.
+  snake_str_list = []
+  prev_char = ""
+  for char in temp_snake_str:
+    if not char == prev_char == "_":
+      snake_str_list.append(char)
+    prev_char = char
```
panhania, line 65: How about `re.sub(r"_+", "_", temp_snake_str)`? :D

grr/core/grr_response_core/lib/casing.py
alexandrucosminmihai: Thank you for your suggestion, I made the change.

grr/core/grr_response_core/lib/casing.py
alexandrucosminmihai: Nice! Made the change. Thank you!

grr/core/grr_response_core/lib/casing.py
```diff
@@ -54,20 +53,15 @@ def CamelToSnake(lower_camel_str: str) -> str:
...
-      snake_str_list.append(char)
-    prev_char = char
+  snake_str = re.sub(r"_+", "_", snake_str)
 
   # Remove the underscore from the start/end, if any.
-  if len(snake_str_list) > 0 and snake_str_list[0] == "_":
-    snake_str_list.pop(0)
-  if len(snake_str_list) > 0 and snake_str_list[-1] == "_":
-    snake_str_list.pop()
+  if len(snake_str) > 0 and snake_str[0] == "_":
```
panhania, line 62: `snake_str.startswith("_")`

grr/core/grr_response_core/lib/casing.py
```diff
@@ -54,20 +53,15 @@ def CamelToSnake(lower_camel_str: str) -> str:
...
+    snake_str = snake_str[1:]
+  if len(snake_str) > 0 and snake_str[-1] == "_":
```
panhania, line 64: `snake_str.endswith("_")`

grr/core/grr_response_core/lib/casing.py
alexandrucosminmihai: Sorry about that, I should have remembered to use the `startswith` and `endswith` methods! Thank you for pointing this out!

panhania (approved):

PR #852, opened 2020-09-15
On query support for time range

tsehori:
Following a discussion on google/fleetspeak#281, we would like to pass a time range to the `/query` API endpoint rather than a specified `limit`. This way, we can retrieve much more meaningful and more accurate datapoints from Fleetspeak's database.

Note that this PR depends on google/fleetspeak#294, and will work only once google/fleetspeak#294 is merged.

### DIFF
--- a/grr/server/grr_response_server/bin/grrafana.py
+++ b/grr/server/grr_response_server/bin/grrafana.py
@@ -2,11 +2,13 @@
 # Lint as: python3
 from absl import app, flags
 import collections
+import dateutil
 import json
 import os
 from typing import Any, cast, Dict, List, Text, Tuple, Iterable
 
 from fleetspeak.src.server.proto.fleetspeak_server import resource_pb2
+from google.protobuf import timestamp_pb2
 
 from grr_response_core import config
 from grr_response_core.config import server as config_server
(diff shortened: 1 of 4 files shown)


grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -155,6 +162,13 @@ def _FetchDatapointsForTargets(
   return response
 
 
+def _GrafanaTimestampToTimestampObj(
```
mbushkov, line 165: Given that we explicitly test this function in the testing module, I'd suggest not making it module-private (i.e. removing the leading "_").

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -155,6 +162,13 @@ def _FetchDatapointsForTargets(
...
+    grafana_time: Text) -> timestamp_pb2.Timestamp:
```
mbushkov, line 166: `Text` was only needed for Python 2->3 migration. Please use `str`.

grr/server/grr_response_server/bin/grrafana_test.py
```diff
@@ -198,7 +202,25 @@ def testQuery(self):
...
+        _TEST_CLIENT_RESOURCE_USAGE_RECORD_1,
+        _TEST_CLIENT_RESOURCE_USAGE_RECORD_2
+    ])
+    with mock.patch.object(fleetspeak_connector, "CONN", conn):
+      self.assertRaises(NameError,
+                        self.client.post,
+                        "/query",
+                        json=_TEST_INVALID_TARGET_QUERY)
+
+  def testGrafanaTimestampToTimestampObj(self):
```
mbushkov, line 217: This is a standalone function that is used by Grrafana HTTP server, but has no direct relation to it. The current class (`GrrafanaTest`) is a set of tests for the HTTP server. I'd put this test in a separate test suite (call it GrafanaTimestampToTimestampObjTest).

mbushkov (changes requested):
Looks good! A few minor comments.

max-vogler (approved):

PR #853, opened 2020-09-17
Use camelCase for property names in OpenAPI description

alexandrucosminmihai:
In this PR I convert protobuf message field names from snake_case to camelCase when using them in the generated OpenAPI description. These field names are used when specifying the expected parameter and the expected result schemas.


mbushkov (commented):
Looks good to me! Will let @panhania formally approve.

panhania (approved):

PR #854, opened 2020-09-17
Use `/api/v2/...` when describing route paths in OpenAPI description

alexandrucosminmihai:
In this PR I replace the `/api/...` path component from the paths annotated to router methods with `/api/v2/...` in the generated OpenAPI description in order to point users to routes of the API that return the described fields with the described level of nesting.

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -908,6 +908,9 @@ def _NormalizePath(path: str) -> str:
   normalized_components = [
     _NormalizePathComponent(component) for component in components
   ]
+  if normalized_components[1] == "api":
+    # We describe the v2 API in the OpenAPI description.
+    normalized_components.insert(2, "v2")
 
   normalized_path = "/".join(normalized_components)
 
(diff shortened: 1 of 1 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -908,6 +908,12 @@ def _NormalizePath(path: str) -> str:
   normalized_components = [
     _NormalizePathComponent(component) for component in components
   ]
+  if (
+      normalized_components[1] == "api"
+      and (len(normalized_components) == 2 or normalized_components[2] != "v2")
+  ):
+    # We describe the v2 API in the OpenAPI description.
+    normalized_components.insert(2, "v2")
```
panhania, line 916: What if the given `path` is empty? Is such scenario possible?

alexandrucosminmihai: I don't know if there's any upstream code that prevents the path to be empty, but I don't think that an empty path makes sense for a HTTP route. So I considered the shortest possible route to be the `/` root so when calling `"/".split("/")` there would be a list with 2 empty strings. Also, I believe that all of GRR API's paths should practically begin with `/api`, which is the case for all the routes defined in `api_call_router.py`. On the other hand, I don't think that an extra check would hurt, but should the `_NormalizePath` return the empty string or should it raise an exception when an empty path is given?

panhania: Okay, I think it is reasonable assumptions and it is fine not to do anything in this case.

panhania (approved):

panhania:
Use v2 API paths in the OpenAPI description.

PR #855, opened 2020-09-18
Fix bug in OpenAPI description generation related to trailing slashes in routes

alexandrucosminmihai:
A [couple of routes](https://github.com/google/grr/blob/77c87c29b4f5c76cbe26ed7a48a37a32a2c12c24/grr/server/grr_response_server/gui/api_call_router.py#L350-L363) have a trailing "/" in their path. This caused an extra empty string to be added to the list of path components when using `path.split("/")` and in turn, this caused the optional path parameter extraction algorithm to fail because the route adding a new path parameter would actually add 2 new path components, instead of one (the actual path parameter).

This PR solves this bug and adds to an existing test in order to verify this case too,

### DIFF
--- a/grr/server/grr_response_server/gui/api_plugins/metadata.py
+++ b/grr/server/grr_response_server/gui/api_plugins/metadata.py
@@ -804,6 +804,10 @@ class ApiGetOpenApiDescriptionHandler(api_call_handler_base.ApiCallHandler):
       ungrouped_routes = []
       for http_method, path, strip_root_types in router_method.http_methods:
         path_components = path.split("/")
+        # Remove trailing empty strings from the list of path components.
+        while path_components and len(path_components[-1]) == 0:
+          path_components.pop()
+
         ungrouped_routes.append([http_method] + path_components)
 
       grouped_routes = _GetGroupedRoutes(ungrouped_routes)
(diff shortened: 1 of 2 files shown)


grr/server/grr_response_server/gui/api_plugins/metadata.py
```diff
@@ -804,6 +804,10 @@ def _GetPaths(self) -> Dict[str, Dict]:
       ungrouped_routes = []
       for http_method, path, strip_root_types in router_method.http_methods:
         path_components = path.split("/")
+        # Remove trailing empty strings from the list of path components.
+        while path_components and len(path_components[-1]) == 0:
+          path_components.pop()
```
panhania, line 809: Do we need to only remove trailing slashes? If we would filter all non-empty strings, wouldn't it be better? Empty path components do not have any meaning anyway, do they?

alexandrucosminmihai: You're right! Currently there is one empty string in all lists caused by the root "/" when calling `str.split`, sort of representing the root component, but it doesn't help us in any way and is present for all routes so it shouldn't affect the path argument extraction. I'll filter out all the empty strings. Thank you!

alexandrucosminmihai: Done. I needed to add the root "/" in `_NormalizePath` because previously it was added using `"/".join(...)` relying on that empty string caused by the `str.split("/")`.

panhania (approved):

PR #863, opened 2020-09-25
Set Travis up to generate OpenAPI description and documentation of GRR API

alexandrucosminmihai:
In this PR I introduce the code and modifications to the Travis configuration that take care of:

1. generating the OpenAPI description JSON of the GRR API and then the HTML documentation file based on it;
2. uploading these two generated files to the `autobuilds-grr-openapi` GCS Bucket.


travis/get_openapi_description.py
```diff
@@ -0,0 +1,40 @@
...
+
+OPENAPI_JSON_FOLDER_NAME = os.environ.get(
+  "OPENAPI_JSON_FOLDER_NAME",
+  default="generated_description"
+)
+OPENAPI_JSON_FILE_NAME = os.environ.get(
+  "OPENAPI_JSON_FILE_NAME",
+  default="openapi_description.json"
+)
+HOME_PATH = os.environ["HOME"]
```
panhania, line 19: I don't think this variable buys us much.

travis/get_openapi_description.py
```diff
@@ -0,0 +1,40 @@
...
+from grr_response_server.gui.api_plugins import metadata as metadata_plugin
+
+OPENAPI_JSON_FOLDER_NAME = os.environ.get(
+  "OPENAPI_JSON_FOLDER_NAME",
+  default="generated_description"
+)
+OPENAPI_JSON_FILE_NAME = os.environ.get(
+  "OPENAPI_JSON_FILE_NAME",
+  default="openapi_description.json"
+)
```
panhania, line 18: Using flags is a cleaner solution, I think. Also, not just make this whole thing take the direct path to the file to write to? Seems much simpler.

travis/upload_build_results_to_gcs.py
```diff
@@ -125,22 +132,48 @@ def _DecryptGCPServiceFileTo(service_file_path: str):
...
 
 
-def _UploadBuildResults(gcs_bucket: storage.Bucket, gcs_build_results_dir: str):
-  """Uploads all build results to Google Cloud Storage."""
-  logging.info("Will upload build results to gs://%s/%s.",
-               os.environ[_GCS_BUCKET], gcs_build_results_dir)
-
-  for build_result in os.listdir(flags.FLAGS.build_results_dir):
-    path = os.path.join(flags.FLAGS.build_results_dir, build_result)
+def _UploadDirectory(local_dir:str, gcs_bucket: storage.Bucket, gcs_dir: str):
```
panhania, line 135: Nit: missing space after `local_dir:`.

travis/upload_build_results_to_gcs.py
```diff
@@ -125,22 +132,48 @@ def _DecryptGCPServiceFileTo(service_file_path: str):
...
+  """Upload the contents of a local directory to a GCS Bucket."""
+  for file_name in os.listdir(local_dir):
+    path = os.path.join(local_dir, file_name)
     if not os.path.isfile(path):
       logging.info("Skipping %s as it's not a file.", path)
       continue
     logging.info("Uploading: %s", path)
-    gcs_blob = gcs_bucket.blob("{}/{}".format(gcs_build_results_dir,
-                                              build_result))
+    gcs_blob = gcs_bucket.blob("{}/{}".format(gcs_dir, file_name))
```
panhania, line 143: Why not f-strings?

travis/upload_build_results_to_gcs.py
```diff
@@ -216,6 +249,14 @@ def main(argv):
...
     gcs_build_results_dir = _GetGCSBuildResultsDir()
     _UploadBuildResults(gcs_bucket, gcs_build_results_dir)
+
+    # Upload the generated OpenAPI description and the generated documentation.
+    if len(flags.FLAGS.openapi_json_dir) > 0:
+      gcs_bucket_openapi = (
+        gcs_client.get_bucket(os.environ[_GCS_BUCKET_OPENAPI])
+      )
+      _UploadOpenApiJson(gcs_bucket_openapi, "openapi_description")
+      _UploadDocumentation(gcs_bucket_openapi, "documentation")
```
panhania, line 259: Should these two values be hardcoded? Everywhere else they seem to be passed either by flags or by environmental variables.

alexandrucosminmihai: These are the names of the directories used in the GCS bucket to upload the VM-local description JSON and documentation HTML, the previous flags and environment variables were used for the paths of the VM-local files. We can definitely use flags for the GCS bucket folders as well, but when I wrote the script I thought they should be relatively static. Please let me know if you want me to get their names through flags as well.

panhania: Makes sense, they can be static.

travis/get_openapi_description.py
alexandrucosminmihai: You're right. Thank you for pointing this out! I removed it and used `os.environ["HOME"]` directly.

travis/get_openapi_description.py
alexandrucosminmihai: Thank you for your suggestion! I replaced my usage of environmental variables with flags and I used the entire file path.

travis/upload_build_results_to_gcs.py
alexandrucosminmihai: Thanks for pointing this out! I added it.

travis/upload_build_results_to_gcs.py
alexandrucosminmihai: The `str.format(...)` was inherited from the previous version of the code. I replaced this instance with an f string and I will replace the other instances in this file in a separate PR. Thank you!

panhania (approved):

PR #870, opened 2020-10-21
GRRafana: Add support to query Client Breakdown Statistics

tsehori:
This PR extends GRRafana HTTP server introduced in #832 by adding new metrics that can be fetched.
These new metrics are the ones that appear in GRR's Statistics page (for reference, you can check out one of the metrics fetched within this page [here](https://github.com/google/grr/blob/0e03976c5c5b694c210c0f392581364ddefa1f4b/grr/server/grr_response_server/gui/api_plugins/report_plugins/client_report_plugins.py#L194)).

The metrics introduced to the server (and Grafana, in turn) are:
- OS Release Version Breakdown
- OS Platform Breakdown
- Client Version

All metrics are available in the interface for 1, 7, 14 and 30 days, and are shown in a table format (similar to the way it currently is in the Statistics page within GRR AdminUI).

In order to support this new metric (which unlike Client Load Stats in #859, which must be provided with a GRR Client ID), we GRRafana to support multiple types of metrics. So we have introduced the abstract Metric class to encapsulate the appropriate logic.

Note that this PR may make the different [cron jobs](https://github.com/google/grr/blob/0e03976c5c5b694c210c0f392581364ddefa1f4b/grr/server/grr_response_server/flows/cron/system.py#L82) that build [`FleetStats`](https://github.com/google/grr/blob/b2969c7ed6f5c6e62919fa66625d992e7b5e0b34/grr/server/grr_response_server/fleet_utils.py#L29) objects and push them to the datastore obsolete. As Grafana issues a request to GRRafana, the latter will call [`CountClientPlatformReleasesByLabel`](https://github.com/google/grr/blob/9a96b96bec76fd06e4f43e3a46cb7450c06e4b88/grr/server/grr_response_server/databases/db.py#L1071) (for example) the build the corresponding `FleetStats` object that will encapsulate the data. Then GRRafana will process it and send it back to Grafana.

Also note that in a future PR, we'd like to cache the results of the queries within GRRafana, as they may get quite time consuming with many labels and data.

### DIFF
--- a/grr/server/grr_response_server/bin/grrafana.py
+++ b/grr/server/grr_response_server/bin/grrafana.py
@@ -129,7 +129,7 @@ AVAILABLE_METRICS_LIST = [
     ClientResourceUsageMetric(
         "Max System CPU Rate", lambda records_list:
         [record.max_system_cpu_rate for record in records_list]),
-    # We multiply to convert from MiB to MB
+    # Converting MiB to MB
     ClientResourceUsageMetric(
         "Mean Resident Memory MB", lambda records_list:
         [record.mean_resident_memory_mib * 1.049 for record in records_list]),
--- a/grr/server/grr_response_server/bin/grrafana.py
+++ b/grr/server/grr_response_server/bin/grrafana.py
@@ -22,16 +22,8 @@ from werkzeug import routing as werkzeug_routing
 from werkzeug import serving as werkzeug_serving
 from werkzeug import wrappers as werkzeug_wrappers
 from werkzeug import wsgi as werkzeug_wsgi
-from werkzeug.wrappers import json as werkzeug_wrappers_json
+from werkzeug.wrappers import json as werkzeug_wrappers_json  # type: ignore
 
-AVAILABLE_METRICS = [
-    "Mean User CPU Rate",
-    "Max User CPU Rate",
-    "Mean System CPU Rate",
-    "Max System CPU Rate",
-    "Mean Resident Memory MB",
-    "Max Resident Memory MB",
-]
 
 JSON_MIME_TYPE = "application/json"
 
@@ -40,6 +32,11 @@ flags.DEFINE_bool(
     default=False,
     allow_override=True,
     help="Print the GRR console version number and exit immediately.")
+
+_Datapoint = Tuple[float, int]
+_Datapoints = List[_Datapoint]
+_TargetWithDatapoints = collections.namedtuple("TargetWithDatapoints",
+                                               ["target", "datapoints"])
 
 
 class JSONRequest(werkzeug_wrappers_json.JSONMixin, werkzeug_wrappers.Request):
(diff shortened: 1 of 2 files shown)


grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +64,128 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+    return _TargetWithDatapoints(target=self.name, datapoints=datapoints)
+
+
+class ClientsStatisticsMetric(Metric):
+
+  def __init__(self, name: str, get_fleet_stats_fn: Callable,
+               days_active: int) -> None:
+    super().__init__(name)
+    self.get_fleet_stats_fn = get_fleet_stats_fn
+    self.days_active = days_active
```
panhania, line 116: Ditto.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +64,128 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+  """A single metric that can be fetched from a
+  Fleetspeak-based GRR deployment."""
+
+  @abstractmethod
+  def __init__(self, name: str) -> None:
+    """Initializes a new metric."""
+    self.name = name
+
+  @abstractmethod
+  def ProcessQuery(self, req: JSONRequest):
```
panhania, line 83: Missing return type.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +64,128 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+        data_store.REL_DB.CountClientPlatformReleasesByLabel(buckets), n_days)
+    for n_days in _FLEET_BREAKDOWN_DAY_BUCKETS
+])
+AVAILABLE_METRICS_LIST.extend([
+    ClientsStatisticsMetric(
+        f"Client Version Strings - {n_days} Day Active", lambda buckets:
+        data_store.REL_DB.CountClientVersionStringsByLabel(buckets), n_days)
+    for n_days in _FLEET_BREAKDOWN_DAY_BUCKETS
+])
+AVAILABLE_METRICS_DICT = {
```
panhania, line 171: I think `AVAILABLE_METRICS_BY_NAME` would be more descriptive.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +64,128 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+        f"OS Release Version Breakdown - {n_days} Day Active", lambda buckets:
+        data_store.REL_DB.CountClientPlatformReleasesByLabel(buckets), n_days)
+    for n_days in _FLEET_BREAKDOWN_DAY_BUCKETS
+])
+AVAILABLE_METRICS_LIST.extend([
+    ClientsStatisticsMetric(
+        f"Client Version Strings - {n_days} Day Active", lambda buckets:
+        data_store.REL_DB.CountClientVersionStringsByLabel(buckets), n_days)
+    for n_days in _FLEET_BREAKDOWN_DAY_BUCKETS
+])
```
panhania, line 170: I think these cram too much stuff into single expressions and are hard to parse for human. Rather than using `extend` with list comprehensions, try to write this with classic `for` loop and `append`s.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -98,115 +214,38 @@ def _OnRoot(self, request: JSONRequest) -> JSONResponse:
...
-    response.append(
-        _TargetWithDatapoints(target=target,
-                             datapoints=datapoints_for_single_target))
-  return response
-
-
 def timeToProtoTimestamp(
     grafana_time: str) -> timestamp_pb2.Timestamp:
-  date = dateutil.parser.parse(grafana_time)
+  date = dateutil.parser.parse(grafana_time)  # type: ignore
```
panhania, line 244: All `type: ignore` should be accompanied by a comment (preferably linking to a particular bug).

grr/server/grr_response_server/bin/grrafana_test.py
```diff
@@ -209,11 +277,26 @@ def testQueryInvalidRequest(self):
         _TEST_CLIENT_RESOURCE_USAGE_RECORD_2
     ])
     with mock.patch.object(fleetspeak_connector, "CONN", conn):
-      self.assertRaises(NameError,
+      self.assertRaises(KeyError,
                         self.client.post,
                         "/query",
                         json=_TEST_INVALID_TARGET_QUERY)
```
panhania, line 283: Prefer to use context manager `assertRaises`:

```python
with self.assertRaises(KeyError):
  self.client.post("/query", json=_TEST_INVALID_TARGET_QUERY)
```

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -22,25 +24,36 @@
...
 JSON_MIME_TYPE = "application/json"
+_FLEET_BREAKDOWN_DAY_BUCKETS = frozenset([1, 7, 14, 30])
 
 flags.DEFINE_bool(
     "version",
     default=False,
     allow_override=True,
     help="Print the GRR console version number and exit immediately.")
 
+_Datapoint = Tuple[float, int]
```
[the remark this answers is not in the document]
tsehori: Do you mean alias the two elements of the tuple?
For instance, 

```python
_Time = float
```

?

grr/server/grr_response_server/bin/grrafana.py
tsehori: That's something I was very divided about. The two implementations of `Metric` (`ClientResourceUsageMetric` and `ClientsStatisticsMetric`) have different return types in the method `ProcessQuery`.
I tried creating a parent class for the named tuples `_TargetWithDatapoints` and `_TableQueryResult`, but [results I found online](https://github.com/python/typing/issues/427) do not suggest there is a good way of doing this (besides using dataclasses).

What do you think?

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -1,18 +1,20 @@
 #!/usr/bin/env python
 # Lint as: python3
+from abc import ABC, abstractmethod
 from absl import app, flags
```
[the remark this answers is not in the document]
tsehori: I see. Then, one should implement a class with no constructor (or that cannot be instantiated) and then inherit from it?
By the way @mbushkov I think we discussed this a couple of weeks ago. I'd appreciate your opinion as well.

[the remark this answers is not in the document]
mbushkov: I think @panhania 's point is not to get rid of abc, but to import it like:
```
import abc
```

and then use its functions as `@abc.abstractmethod`, etc. So you still very much should use the `abc` package, but just import it differently.

grr/server/grr_response_server/bin/grrafana.py
mbushkov: Can you use [Union](https://docs.python.org/3/library/typing.html#typing.Union)?

grr/server/grr_response_server/bin/grrafana.py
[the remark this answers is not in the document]
panhania: No, no. Rather, use a `NamedTuple` with proper names for both fields rather than just `Tuple` that values have to be accessed by `[0]` and `[1]`..

grr/server/grr_response_server/bin/grrafana.py
panhania: > Can you use Union?

I would refrain from using `Union`, if possible. They can be useful for easy typifying of existing code, but usually they indicate poor design.

Here, if you look at how your abstract method is used it is easy to understand how to type it:

```python
targets_with_datapoints = [
    AVAILABLE_METRICS_DICT[target].ProcessQuery(json_data)
    for target in requested_targets
]
response = [t._asdict() for t in targets_with_datapoints]
return JSONResponse(response=response)
```

You don't care about a particular type, you care that you can call `_asdict()` on the results. Hence, it is enough to say that `ProcessQuery` should return a `NamedTuple` instance. Whether it is `ClientResourceUsageMetric` or `ClientsStatisticsMetric` doesn't matter.

tsehori: That makes sense, the problem I get when I try to specify `NamedTuple` as a return type for `ProcessQuery`, mypy generates the following errors:
```
grrafana.py:112: error: Return type "_TargetWithDatapoints" of "ProcessQuery" incompatible with return type "NamedTuple" in supertype "Metric"
grrafana.py:136: error: Return type "_TableQueryResult" of "ProcessQuery" incompatible with return type "NamedTuple" in supertype "Metric"
```
It's weird as both `_TargetWithDatapoints` and `_TableQueryResult` inherit from `NamedTuple`.

tsehori:
PTAL @mbushkov @panhania

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -98,115 +214,38 @@ def _OnRoot(self, request: JSONRequest) -> JSONResponse:
...
-  for target in targets:
-    datapoints_for_single_target = _CreateDatapointsForTarget(
-        target, records_list)
-    response.append(
-        _TargetWithDatapoints(target=target,
-                             datapoints=datapoints_for_single_target))
-  return response
-
-
 def timeToProtoTimestamp(
```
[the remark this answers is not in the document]
tsehori: I'll change it to PascalCase (as of the rest of the code, as discussed in #832) in a separate PR.

grr/server/grr_response_server/bin/grrafana.py
panhania: Hmm, you are right. Looks like this is a known issue: python/mypy#3915, python/typing#431. The code seems to work under pytype, though. I would still try to annotate it with `NamedTuple`, add `# type: ignore[return-value]` to suppress the warnings and add a `TODO` comment with links to the mentioned tracking issues.

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +73,134 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+  Fleetspeak-based GRR deployment."""
+
+  @abc.abstractmethod
+  def __init__(self, name: str) -> None:
+    """Initializes a new metric."""
+    self._name = name
+
+  def GetName(self) -> str:
+    """Returns the name of the metric."""
+    return self._name
```
panhania, line 93: This would look more naturally as an accessor (`@property`).

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +73,134 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
 
 
+class Metric(abc.ABC):
+  """A single metric that can be fetched from a
+  Fleetspeak-based GRR deployment."""
+
+  @abc.abstractmethod
+  def __init__(self, name: str) -> None:
+    """Initializes a new metric."""
+    self._name = name
```
panhania, line 89: Methods are either abstract or have some implementation. In general, constructors should not be abstract (since interfaces specify the that instances should fulfill but not how they should be constructed).

grr/server/grr_response_server/bin/grrafana.py
```diff
@@ -51,22 +73,134 @@ class JSONResponse(werkzeug_wrappers_json.JSONMixin,
...
+                                   r.server_timestamp.nanos // 1000000)
+                  for (v, r) in zip(record_values, records_list)]
+    return _TargetWithDatapoints(target=self._name, datapoints=datapoints)
+
+
+class ClientsStatisticsMetric(Metric):
+
+  def __init__(self,
+               name: str,
+               get_fleet_stats_fn: Callable[[frozenset], fleet_utils.FleetStats],
```
panhania, line 130: Prefer not to use raw `frozenset` in type annotations, but rather generic [`FrozenSet`][1] with specific type.

[1]: https://docs.python.org/3/library/typing.html#typing.FrozenSet

panhania (changes requested):

grr/server/grr_response_server/bin/grrafana.py
tsehori: > Hmm, you are right. Looks like this is a known issue: [python/mypy#3915](https://github.com/python/mypy/issues/3915), [python/typing#431](https://github.com/python/typing/issues/431). The code seems to work under pytype, though. I would still try to annotate it with `NamedTuple`, add `# type: ignore[return-value]` to suppress the warnings and add a `TODO` comment with links to the mentioned tracking issues.

Will do. Just to clarify; the error is actually an [override](https://mypy.readthedocs.io/en/stable/error_code_list.html#check-validity-of-overrides-override) code and not a return-value code.

tsehori:
PTAL @panhania

panhania (approved):

PR #871, opened 2020-10-22
MVP details component for Osquery

simstoykov:


simstoykov:
The component looks like this:
Screenshot 2020-10-22 at 5 08 03 PM

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -0,0 +1,34 @@
...
+  <div *ngIf="resultTable$ | async as resultTable">
+    <div class="results-query">
+      <div class="results-query-label">Query:</div>&nbsp;
+      <div class="results-query-text">{{resultTable.query}}</div>
+    </div>
+
+    <div class="table-container">
+      <table>
+        <thead>
+          <th *ngFor="let column of resultTable.header.columns">{{column.name}}</th>
```
max-vogler, line 15: add `trackBy: fn` and track the entries by a primitive, not a POJO. Since objects are unequal, unless they are the same reference, Angular would re-render this table even if the data is actually the same.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -0,0 +1,34 @@
...
+        </thead>
+        <tbody>
+          <tr *ngFor="let rowValues of resultTable.rows">
```
max-vogler, line 18: Add `trackBy`.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+@Component({
+  selector: 'osquery-details',
+  templateUrl: './osquery_details.ng.html',
+  styleUrls: ['./osquery_details.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OsqueryDetails extends Plugin implements OnDestroy {
+  private readonly onDestroyHappened$ = new Subject<void>();
+
+  hasLoadedResults = false;
```
max-vogler, line 20: Make members like this private if possible.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  flowRunning$ = this.flagByState(FlowState.RUNNING);
+  flowCompleted$ = this.flagByState(FlowState.FINISHED);
+  flowError$ = this.flagByState(FlowState.ERROR);
```
max-vogler, line 24: Make observable members `readonly`.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  args$: Observable<OsqueryArgs> = this.flowListEntry$.pipe(
+      map((flowListEntry) => flowListEntry.flow.args as OsqueryArgs)
+  );
+
+  osqueryResult$: Observable<OsqueryResult> = this.flowListEntry$.pipe(
+    flatMap((listEntry) => listEntry.resultSets),
+    flatMap((singleResultSet) => singleResultSet?.items),
+    map((singleItem) => singleItem?.payload as OsqueryResult)
```
max-vogler, line 33: This casts nullable values to a non-nullable type, which can trigger hard-to-trace runtime errors. Instead of casting away the `undefined` and `null`, please use `filter(isNonNullable)`.

Alternatively, cast to `OsqueryResult | undefined` and use this type on the member definition as well.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Component that displays the details (status, errors, results) for a particular Osquery Flow.
+ */
+@Component({
+  selector: 'osquery-details',
+  templateUrl: './osquery_details.ng.html',
+  styleUrls: ['./osquery_details.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OsqueryDetails extends Plugin implements OnDestroy {
+  private readonly onDestroyHappened$ = new Subject<void>();
```
max-vogler, line 18: We typically call this `unsubscribe$` in our code base.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+  flagByState(targetState: FlowState): Observable<boolean> {
+    return this.flowListEntry$.pipe(
+      map((listEntry) => listEntry.flow.state === targetState)
+    );
+  }
+
+  constructor() {
+    super();
+    this.completeOnDestroy(this.flowCompleted$)
+      .subscribe(() => this.loadResults());
```
max-vogler, line 53: I'd suggest inlining `.pipe(takeUntil(this.unsubscibe$))`, because `completeOnDestroy()` is used only once and makes reading the code harder.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  completeOnDestroy<T>(obs: Observable<T>): Observable<T> {
+    return obs.pipe(takeUntil(this.onDestroyHappened$));
+  }
+
+  loadResults() {
+    if (!this.hasLoadedResults) {
```
max-vogler, line 61: Instead of keeping `hasLoadedResults` state, add `take(1)` into the `this.flowCompleted$.pipe()`, so the subscriber that calls `loadResults` is only called once. Then, you can inline `this.queryFlowResults()` into the lambda subscriber:

```
this.flowCompleted$.pipe(
  takeUntil(this.unsubscribe$),
  filter(completed => completed),
  take(1),
).subscribe(() => {
  this.queryFlowResults(...);
});
```

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
max-vogler, line 53: When `flowCompleted$` emits `false`, `loadResults()` is still triggered. This looks like a bug to me.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details_test.ts
```diff
@@ -0,0 +1,203 @@
...
+import { FlowState, FlowListEntry, FlowResultSet, FlowResultSetState } from '@app/lib/models/flow';
+import { By } from '@angular/platform-browser';
+import { DebugElement, Component } from '@angular/core';
+import { OsqueryResult, OsqueryTable } from '@app/lib/api/api_interfaces';
+
+
+initTestEnvironment();
+
+/** Helper class to build a FlowListEntry objects in a declarative manner */
+class FlowListEntryBuilder {
```
max-vogler, line 16: The builder pattern is very Java-y, coming from Java's less powerful, stricter, nominal type system. We typically use TypeScript's structural typing for less verbose code. Please re-use `model_test_util.newFlowListEntry()`. Your `withResult` and `withTable` functions are very helpful though, please do keep them!

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
simstoykov: I looked into that, but I couldn't understand exactly under what circumstances our component could get different objects for the same value. Thinking about it now, I guess this could happen if we request data in batches, more than once?

This reminded me I should add support for getting all the data, as currently I'm just getting the first batch.

Also I'm not sure at the moment how often will this function be called. If we have 1m rows, I should probably think about the efficiency of the function? A thought that comes to mind is to use the row content's hash. Chances for collision are small even with 1m rows.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details_test.ts
simstoykov: Could we have a short VC once you are back and discuss that? I spent some time thinking about it and playing with various different options, and the builder still seems like the cleanest way to do it. The only improvement I could think of was to set `private flowListEntry: any = newFlowListEntry();`, and I'm not sure that's the cleanest way to do it because I'm throwing away type information.

simstoykov:
@max-vogler can you PTAL

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
max-vogler: You should be able to remove `?` from `singleItem?.payload` after `filter(isNonNull)`.

grr/server/grr_response_server/gui/ui/lib/models/model_test_util.ts
```diff
@@ -96,3 +96,23 @@ export function newClientApproval(args: Partial<ClientApproval> = {}):
     ...args,
   };
 }
+
+/**
+ * A helper function to construct a FlowResultSet object.
+ * Feel free to extend it with more default arguments.
```
max-vogler, line 102: "Feel free to extend it with more default arguments." is a comment about an invariant that might not age well. I suggest removing it.

grr/server/grr_response_server/gui/ui/lib/models/model_test_util.ts
```diff
@@ -96,3 +96,23 @@ export function newClientApproval(args: Partial<ClientApproval> = {}):
...
+ */
+export function newResultSet(payload = { }): FlowResultSet {
```
max-vogler, line 104: Please rename to `newFlowResultSet`.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -12,12 +12,12 @@
...
         <thead>
-          <th *ngFor="let column of resultTable.header.columns">{{column.name}}</th>
+          <th *ngFor="let column of resultTable.header.columns; trackBy: trackByIndex">{{column.name}}</th>
         </thead>
         <tbody>
-          <tr *ngFor="let rowValues of resultTable.rows">
-            <td *ngFor="let value of rowValues.values">
-              {{value}}
+          <tr *ngFor="let row of resultTable.rows; trackBy: trackByIndex">
+            <td *ngFor="let cellValue of row.values; trackBy: trackByIndex">
```
max-vogler, line 19: `trackBy` is not needed for primitives (which I assume `cellValue` is because you print it?)

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -12,12 +12,12 @@
     <div class="table-container">
       <table>
         <thead>
-          <th *ngFor="let column of resultTable.header.columns">{{column.name}}</th>
+          <th *ngFor="let column of resultTable.header.columns; trackBy: trackByIndex">{{column.name}}</th>
```
max-vogler, line 15: We typically trackBy an object key (see https://stackoverflow.com/questions/54478273/angular-trackby-returning-the-index-in-ngfor-vs-item-id). Tracking by an index seems to be fine in this scenario, no change needed.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
simstoykov: I'm curious, what "key" exactly? "key" in the sense of a unique value that identifies the content of the thing and depends on the case?

max-vogler: `column.name` would be a sensible key. On top of the performance benefits through DOM reuse from tracking by key/index (see [so](https://stackoverflow.com/questions/54478273/angular-trackby-returning-the-index-in-ngfor-vs-item-id)), tracking by a key works better when list items are re-ordered. We don't really re-order items, so it does not matter too much for us.

simstoykov:
@max-vogler PTAL

grr/server/grr_response_server/gui/ui/lib/api/api_interfaces.ts
```diff
@@ -627,3 +627,45 @@ export declare interface OsqueryArgs {
...
+export declare interface OsqueryResult {
+  readonly table?: OsqueryTable;
+  readonly stderr?: string;
+}
+
+/** OsqueryTable proto mapping */
+export declare interface OsqueryTable {
+  readonly query?: string;
+  readonly header?: OsqueryHeader;
+  readonly rows?: OsqueryRow[];
```
max-vogler, line 641: Always use `ReadonlyArray<...>`.

grr/server/grr_response_server/gui/ui/lib/models/model_test_util.ts
```diff
@@ -96,3 +96,22 @@ export function newClientApproval(args: Partial<ClientApproval> = {}):
     ...args,
   };
 }
+
+/**
+ * A helper function to construct a FlowResultSet object.
```
max-vogler, line 101: This JSDoc is unneeded - it adds no new information. For the whole file, we disabled the linter error of missing JSDoc in order to reduce the visual clutter. Please remove.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -0,0 +1,34 @@
+<div *ngIf="flowRunning$ | async" class="in-progress">
+  {{ (args$ | async).query }}
+</div>
+
+<div *ngIf="flowCompleted$ | async" class='results'>
+  <div *ngIf="resultTable$ | async as resultTable">
+    <div class="results-query">
+      <div class="results-query-label">Query:</div>&nbsp;
```
max-vogler, line 8: Are you sure this nbsp works as intended (since there is whitespace surrounding it). Also, instead of using `<div>`s with `display: inline` in the style, prefer using `<span>` which is `display: inline` per default.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
simstoykov: Basically I want some margin around the label and the query. So maybe I should use the margin style?

Didn't know about `span`s, I'm gonna use them from now on.

simstoykov:
@max-vogler PTYAL

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
max-vogler: Yes, please use explicit styling (padding/margin) instead of repurposing special characters. Haven't seen &nbsp; for styling in a looooong time :D

max-vogler (approved):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
simstoykov: Hmmmmm: https://github.com/google/grr/blob/22db4ed1a6020437fff7b84b5f32ece23db9b9c9/grr/server/grr_response_server/gui/ui/components/flow_details/helpers/file_results_table.ng.html#L17

max-vogler: That's the intended use: There should be a space between the path and the copy button, but the copy button should stay on the same line without line-break: non-breaking space.

simstoykov: Indeed that's the only one. Should we use styling there too?

simstoykov: Oh, or maybe you mean we actually need the non-breaking feature, and we won't be getting that with just setting the margin style?

PR #873, opened 2020-10-27
Fix error in queryFlowResults functionality

simstoykov:
The old queryFlowResults implementation would discard requests for more data if another one is currently in progress, even if the new requests are coming from different flows. This caused only the first flow which requests results to get them. 

The new implementation groups requests by flows so that different flows don't affect each other. Additionaly, within the same group, it effectively creates a queue of size 1 for any pending requests, which ensures that the last request that was made will always be satisfied.

### DIFF
--- a/grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
+++ b/grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
@@ -1,0 +1,28 @@
+import { ObservableInput, OperatorFunction, ObservedValueOf, pipe, EMPTY } from 'rxjs';
+import { concatMap, tap } from 'rxjs/operators';
+
+/** TODO */
+export function queuedExhaustMap<T, O extends ObservableInput<any>> (
+  project: (value: T, index: number) => O
+): OperatorFunction<T, ObservedValueOf<O>> {
+
+  function recordLastRequest(value: T): void {
+
+  }
+  function isLastRequest(value: T): boolean {
+    return true;
+  }
+
+  return pipe(
+    tap((value) => {
+      recordLastRequest(value);
+    }),
+    concatMap((value: T, index: number) => {
+      if (isLastRequest(value)) {
+        return project(value, index);
+      } else {
+        return EMPTY;
+      }
+    }),
+  );
+}
--- a/grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
+++ b/grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
@@ -1,24 +1,43 @@
 import { ObservableInput, OperatorFunction, ObservedValueOf, pipe, EMPTY } from 'rxjs';
 import { concatMap, tap } from 'rxjs/operators';
 
-/** TODO */
+class NotImplementedError extends Error { };
+
+class RequestTracker<T> {
+  lastValue?: T;
+  isLastValueSet = false;
+
+  recordLast(value: T): void {
+    this.lastValue = value;
+    this.isLastValueSet = true;
+  }
+
+  shouldProcess(value: T): boolean {
+    return this.isLastValueSet && this.lastValue === value;
+  }
+}
+
+/**
+ * Similar to exhaustMap, but buffers the specified number of most recent source values
+ * in a queue of limited size. The source values that are not discarded are projected to
+ * Observables and merged in the output Observable in a serialized fashion, waiting for
+ * each one to complete before merging the next.
+ */
 export function queuedExhaustMap<T, O extends ObservableInput<any>> (
-  project: (value: T, index: number) => O
+  project: (value: T, index: number) => O, queueSize: 1 = 1
 ): OperatorFunction<T, ObservedValueOf<O>> {
-
-  function recordLastRequest(value: T): void {
-
+  if (queueSize !== 1) {
+    throw new NotImplementedError('Queue size different than 1 is not implemented yet.');
   }
-  function isLastRequest(value: T): boolean {
-    return true;
-  }
+
+  const requestTracker = new RequestTracker<T>();
 
   return pipe(
     tap((value) => {
-      recordLastRequest(value);
+      requestTracker.recordLast(value);
     }),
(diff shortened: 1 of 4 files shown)


simstoykov:
I mostly implemented the features required for the fix and I wanted to get an early review. 
What's left is to tidy the code into separate functions/effects and add tests for the client page facade.

grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map_test.ts
```diff
@@ -0,0 +1,61 @@
...
+  }));
+});
+
+function getProcessedValues<T>(observable$: Observable<T>) {
+  const processedValues = Array();
+  observable$.subscribe((value) => {
+    processedValues.push(value)
+  });
+  return processedValues;
+}
```
simstoykov, line 61: Add a test for emitting the same value.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -355,9 +356,18 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
...
-          exhaustMap(([query, clientId]) => {
+        takeUntil(this.selectedClientIdChanged$),
+        withLatestFrom(this.selectedClientId$),
+        // Below we are grouping the requests by flowId, tag and type, and
+        // for each group we use a queuedExhaustMap with queue size 1 to send http request
+        // for results.
+        // TODO: Think about how we should split the code so it's cleaner
+        // TODO: Add tests to ensure the intended functionality
+        groupBy(([query, clientId]) => {
+          return [query.flowId, query.withTag, query.withType].join('-');
```
simstoykov, line 367: Separate this into a function.

Also, URI-encode the characters and use a special delimiter which is encoded to a different value.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
```diff
@@ -355,9 +356,18 @@ export class ClientPageStore extends ComponentStore<ClientPageState> {
...
+        }),
+        mergeMap(group$ => group$.pipe(
+          queuedExhaustMap(([query, clientId]) => {
```
simstoykov, line 370: Try different approaches for making this cleaner. But it's also not-that-awful this way. We could leave it like this for now, and make it better as part of the general refactoring of `client_page_facade.ts`.

grr/server/grr_response_server/gui/ui/store/client_page_facade.ts
simstoykov: Also provide a small test for the function.

simstoykov:
@mbushkov PTAL

grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map_test.ts
```diff
@@ -0,0 +1,88 @@
...
+        return of(value).pipe(
+          delay(blockTimeMs),
+        );
+      }, 1),
+    );
+
+    const processedValues = getProcessedValues(manyValuesWithDelay$);
+
+    tick(blockTimeMs);
+    expect(processedValues).toEqual(['initial']);
```
mbushkov, line 48: Here it gets a bit confusing. There's an assumption that processedValues will change behind the scenes, since its "captured" in a closure inside getProcessedValues. That's maybe ok for tests, but is somehow hard to follow.

What about changing the interface of getProcessedValues() to:
```
getProcessedValues<T>(observable$: Observable<T>): () => ReadonlyArrat<T>
```

Basically, getProcessedValues() will return a callable. This callable will be returning all the emitted elements from the last call.

Then your code here would look like:
```
const processedValuesFn = getProcessedValues(manyValuesWithDelay$);

tick(blockTimeMs);
expect(processedValuesFn()).toEqual(['initial']);

tick(blockTimeMs);
expect(processedValuesFn()).toEqual(['lastSaved']);
```

Then what previously was happening behind-the-scenes is now explicit and connected to a function call. What do you think?

mbushkov (changes requested):
Looks really good! Just a few minor comments in the testing code and we're good to go.

grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
```diff
@@ -0,0 +1,69 @@
...
+
+class NotImplementedError extends Error { };
+
+interface RequestTracker<T> {
+  recordLastTag(tag: T): void;
+  shouldProcess(thisTag: T): boolean;
+}
+
+class SingleIndexTracker implements RequestTracker<number> {
+  private lastIndex = -1;
```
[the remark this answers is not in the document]
simstoykov: What about renaming `tag` and `thisTag` parameters below to `index` and `thisIndex`? In this `SingleIndexTracker` implementation all tags are indices, so I wanted to emphasize that.

grr/server/grr_response_server/gui/ui/lib/queued_exhaust_map.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Similar to exhaustMap, but buffers the specified number of most recent source values
+ * in a queue of limited size. The source values that are not discarded are projected to
+ * Observables and merged in the output Observable in a serialized fashion, waiting for
+ * each one to complete before merging the next.
+ *
+ * Important notes:
+ * - Currently only queue of size 1 is supported.
+ * - If the source produces values faster than this operation discards them, no values will
```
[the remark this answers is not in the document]
simstoykov: Yeap you are right, that's a different issue I forgot to describe, and there is also a warning in the [RxJS concatMap page](https://rxjs-dev.firebaseapp.com/api/operators/concatMap):
> Warning: if source values arrive endlessly and faster than their corresponding inner Observables can complete, it will result in memory issues as inner Observables amass in an unbounded buffer waiting for their turn to be subscribed to.

What about the "If the source produces values faster...": do you think that's a valid note?

mbushkov (approved):

PR #876, opened 2020-10-29
Add progress info for the Osquery Flow (Backend)

simstoykov:
Report the first 10 rows, together with the total number of rows as "Progress".
This PR only contains the backend changes.

### DIFF
--- a/grr/server/grr_response_server/flows/general/osquery.py
+++ b/grr/server/grr_response_server/flows/general/osquery.py
@@ -27,8 +27,8 @@ class OsqueryFlow(flow_base.FlowBase):
 
   def Start(self):
     super(OsqueryFlow, self).Start()
-
     self.state.progress = rdf_osquery.OsqueryProgress()
+
     self.CallClient(
         server_stubs.Osquery,
         request=self.args,
@@ -38,16 +38,21 @@ class OsqueryFlow(flow_base.FlowBase):
     if not responses.success:
       raise flow_base.FlowError(responses.status)
 
+    self.UpdateProgress(responses)
+
     for response in responses:
-      self.UpdateProgress(response.table)
       self.SendReply(response)
 
   def GetProgress(self) -> rdf_osquery.OsqueryProgress:
     return self.state.progress
 
-  def UpdateProgress(self, table: rdf_osquery.OsqueryTable) -> None:
-    self.state.progress.partialTable = TruncateRows(table)
-    self.state.progress.totalRowsCount = len(table.rows)
+  def UpdateProgress(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
+    # XXX: Less than TRUNCATED_ROWS_COUNT will be used if the original table is split into chunks smaller than that 
+    first_chunk = responses.responses[0]
+    table = first_chunk.table
+
+    self.state.progress.partial_table = TruncateRows(table)
+    self.state.progress.total_rows_count = len(table.rows)
 
 
 def TruncateRows(table: rdf_osquery.OsqueryTable) -> rdf_osquery.OsqueryTable:
--- a/grr/server/grr_response_server/flows/general/osquery.py
+++ b/grr/server/grr_response_server/flows/general/osquery.py
@@ -14,6 +14,15 @@ from grr_response_server import flow_responses
 
 TRUNCATED_ROWS_COUNT = 10
 
+def _TruncateRows(table: rdf_osquery.OsqueryTable) -> rdf_osquery.OsqueryTable:
+  result = rdf_osquery.OsqueryTable()
+
+  result.query = table.query
+  result.header = table.header
+  result.rows = table.rows[:TRUNCATED_ROWS_COUNT]
+
+  return result
+
 
 class OsqueryFlow(flow_base.FlowBase):
   """A flow mixin wrapping the osquery client action."""
@@ -25,6 +34,19 @@ class OsqueryFlow(flow_base.FlowBase):
   args_type = rdf_osquery.OsqueryArgs
   progress_type = rdf_osquery.OsqueryProgress
 
+  def _UpdateProgress(
+    self,
+    responses: flow_responses.Responses[rdf_osquery.OsqueryResult]
+  ) -> None:
+    # XXX: Less than TRUNCATED_ROWS_COUNT will be used if the original table is
+    # split into chunks smaller than that. It is expected that the limit will
+    # always be smaller than the chunk size, so this shouldn't happen.
+    first_chunk = responses.responses[0]
+    table = first_chunk.table
+
+    self.state.progress.partial_table = _TruncateRows(table)
+    self.state.progress.total_rows_count = len(table.rows)
+
   def Start(self):
     super(OsqueryFlow, self).Start()
     self.state.progress = rdf_osquery.OsqueryProgress()
@@ -34,32 +56,17 @@ class OsqueryFlow(flow_base.FlowBase):
         request=self.args,
         next_state=compatibility.GetName(self.Process))
 
-  def Process(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
+  def Process(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult]
+  ) -> None:
     if not responses.success:
       raise flow_base.FlowError(responses.status)
 
-    self.UpdateProgress(responses)
+    self._UpdateProgress(responses)
 
     for response in responses:
       self.SendReply(response)
 
   def GetProgress(self) -> rdf_osquery.OsqueryProgress:
     return self.state.progress
-
-  def UpdateProgress(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
-    # XXX: Less than TRUNCATED_ROWS_COUNT will be used if the original table is split into chunks smaller than that 
-    first_chunk = responses.responses[0]
-    table = first_chunk.table
-
-    self.state.progress.partial_table = TruncateRows(table)
-    self.state.progress.total_rows_count = len(table.rows)
-
-
-def TruncateRows(table: rdf_osquery.OsqueryTable) -> rdf_osquery.OsqueryTable:
-  result = rdf_osquery.OsqueryTable()
-
-  result.query = table.query
-  result.header = table.header
-  result.rows = table.rows[:TRUNCATED_ROWS_COUNT]
-
-  return result
--- a/grr/server/grr_response_server/flows/general/osquery.py
+++ b/grr/server/grr_response_server/flows/general/osquery.py
@@ -9,6 +9,10 @@ from grr_response_core.lib.rdfvalues import osquery as rdf_osquery
 from grr_response_core.lib.util import compatibility
 from grr_response_server import flow_base
 from grr_response_server import server_stubs
+from grr_response_server import flow_responses
+
+
+TRUNCATED_ROWS_COUNT = 10
 
 
 class OsqueryFlow(flow_base.FlowBase):
@@ -19,17 +23,38 @@ class OsqueryFlow(flow_base.FlowBase):
   behaviours = flow_base.BEHAVIOUR_BASIC
 
   args_type = rdf_osquery.OsqueryArgs
+  progress_type = rdf_osquery.OsqueryProgress
 
   def Start(self):
     super(OsqueryFlow, self).Start()
+
+    self.state.progress = rdf_osquery.OsqueryProgress()
     self.CallClient(
         server_stubs.Osquery,
         request=self.args,
         next_state=compatibility.GetName(self.Process))
 
-  def Process(self, responses):
+  def Process(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
     if not responses.success:
       raise flow_base.FlowError(responses.status)
 
     for response in responses:
+      self.UpdateProgress(response.table)
       self.SendReply(response)
+
+  def GetProgress(self) -> rdf_osquery.OsqueryProgress:
+    return self.state.progress
+
+  def UpdateProgress(self, table: rdf_osquery.OsqueryTable) -> None:
+    self.state.progress.partialTable = TruncateRows(table)
+    self.state.progress.totalRowsCount = len(table.rows)
+
+
+def TruncateRows(table: rdf_osquery.OsqueryTable) -> rdf_osquery.OsqueryTable:
+  result = rdf_osquery.OsqueryTable()
+
+  result.query = table.query
+  result.header = table.header
+  result.rows = table.rows[:TRUNCATED_ROWS_COUNT]
+
+  return result
--- a/grr/server/grr_response_server/flows/general/osquery_test.py
+++ b/grr/server/grr_response_server/flows/general/osquery_test.py
@@ -9,12 +9,17 @@ import io
 import os
 import time
 
+from typing import List
+from collections import OrderedDict
+
 from absl import app
 
 from grr_response_client.client_actions import osquery as osquery_action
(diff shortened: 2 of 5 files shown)


grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +23,43 @@ class OsqueryFlow(flow_base.FlowBase):
...
         server_stubs.Osquery,
         request=self.args,
         next_state=compatibility.GetName(self.Process))
 
-  def Process(self, responses):
+  def Process(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
     if not responses.success:
       raise flow_base.FlowError(responses.status)
 
+    self.UpdateProgress(responses)
```
mbushkov, line 41: I would make UpdateProgress protected: _UpdateProgress.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +23,43 @@ class OsqueryFlow(flow_base.FlowBase):
...
+
     for response in responses:
       self.SendReply(response)
+
+  def GetProgress(self) -> rdf_osquery.OsqueryProgress:
+    return self.state.progress
+
+  def UpdateProgress(self, responses: flow_responses.Responses[rdf_osquery.OsqueryResult]) -> None:
```
mbushkov, line 49: Throughout the codebase autoformat with 80 characters max line length. Please use this line length limit throughout the file (do you think you can configure your autoformatter with the same setting?)

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +23,43 @@ class OsqueryFlow(flow_base.FlowBase):
   behaviours = flow_base.BEHAVIOUR_BASIC
 
   args_type = rdf_osquery.OsqueryArgs
+  progress_type = rdf_osquery.OsqueryProgress
```
mbushkov, line 26: I don't see a change in rdfvalues/osquery.py included in to this PR.

mbushkov (changes requested):
Looks really good! Please see a few minor comments below.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -148,47 +153,101 @@ def testFailure(self):
...
+        client_id=self.client_id,
+        token=self.token,
+        query=query)
+    return session_id
+
+  def _RunQuery(self, query):
+    session_id = self._NewTestFlowStartAndRun(query)
+    return flow_test_lib.GetFlowResults(self.client_id, session_id)
+
+  def _AssertTablesMatch(self, table: rdf_osquery.OsqueryTable, expected: List[OrderedDict]) -> None:
```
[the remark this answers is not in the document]
simstoykov: I agree in this particular case - my code should be consistent with the rest of the code base.

In general, though, what are the benefits of that? I guess this might be a left-over from languages like C in which you need to declare things before using them.

If instead protected methods are put at the end, they will make way for the public methods to be first. The public methods are the interface exposed, so if anyone needs to look at the class they will be presented with the core high-level functionality first. This in my opinion is better than first exposing low-level helper functions?

simstoykov:
@mbushkov can you PTAL

grr/server/grr_response_server/flows/general/osquery_test.py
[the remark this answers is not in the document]
mbushkov: > If instead protected methods are put at the end, they will make way for the public methods to be first. The public methods are the interface exposed, so if anyone needs to look at the class they will be presented with the core high-level functionality first. This in my opinion is better than first exposing low-level helper functions?

This is a good question. I don't have a strong opinion on this as I think that practical differences between different ways of grouping methods together are negligible. It's normally useful to have them grouped (even though sometimes it actually makes sense to put a protected method right next to the public method using it), but whether it's better to have public first, protected second or protected first, public second - feels like a subjective preference. Developers have different backgrounds, use different IDEs, etc.

I personally find it useful to have the implementation details grouped together, and since often private object properties are defined at the top of the class, I see the point of grouping them with private/protected methods. Then, when I read the class definition I skim through all the private/protected stuff, getting an idea of what is there, and then I can read through public implementation with that additional context in mind. Again, this is very subjective, so the only real guideline in this question is code consistency - whichever approach we choose, makes sense to stick to it.

mbushkov (approved):

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -147,48 +152,108 @@ def testFailure(self):
...
-    stdout = """
-    [
-      { "foo": "quux", "bar": "norf", "baz": "thud" },
-      { "foo": "blargh", "bar": "plugh", "baz": "ztesch" }
+    test_table = [
+      OrderedDict([ ("foo", "quux"), ("bar", "norf"), ("baz", "thud") ]),
+      OrderedDict([ ("foo", "blargh"), ("bar", "plugh"), ("baz", "ztesch") ]),
     ]
-    """
+    stdout = json.Dump(test_table)
```
panhania, line 209: Why? This contains more logic and is more complicated (it has more structure, yes, but this is just test data, do we care?).

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -147,48 +152,108 @@ def testFailure(self):
...
-class FakeOsqueryFlowTest(flow_test_lib.FlowTestsBaseclass):
+def _CreateDictTable(column_number, row_number) -> List[OrderedDict]:
+  def CellValue(col_number, row_number):
+    return (f"col-{col_number}", f"col-{col_number}, row-{row_number}")
+
+  return [
+    OrderedDict(
+      CellValue(col_number, row_number) for col_number in range(column_number)
+    ) for row_number in range(row_number)
+  ]
```
panhania, line 163: This makes the test code hard to follow—avoid using any helpers in the test code if possible. Here, you don't care about the values at all (whether they are unique or not) and the number of columns is not too high, so you can prepare a readable test data yourself, e.g. `[{ "foo": "quux", "baz": "thud", "bar": "norf" }] * 42`.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +32,41 @@ class OsqueryFlow(flow_base.FlowBase):
...
+
+  def _UpdateProgress(
+    self,
+    responses: flow_responses.Responses[rdf_osquery.OsqueryResult]
+  ) -> None:
+    # XXX: Less than TRUNCATED_ROWS_COUNT will be used if the original table is
```
panhania, line 41: What is `XXX`?

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +32,41 @@ class OsqueryFlow(flow_base.FlowBase):
...
+
     self.CallClient(
         server_stubs.Osquery,
         request=self.args,
         next_state=compatibility.GetName(self.Process))
 
-  def Process(self, responses):
+  def Process(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult]
```
panhania, line 61: Missing trailing comma (it is needed to force auto-formatter not to fold the result type to the same line).

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -19,17 +32,41 @@ class OsqueryFlow(flow_base.FlowBase):
...
     if not responses.success:
       raise flow_base.FlowError(responses.status)
 
+    self._UpdateProgress(responses)
+
     for response in responses:
       self.SendReply(response)
+
+  def GetProgress(self) -> rdf_osquery.OsqueryProgress:
+    return self.state.progress
```
panhania, line 72: This can be an accessor.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -9,6 +9,19 @@
 from grr_response_core.lib.util import compatibility
 from grr_response_server import flow_base
 from grr_response_server import server_stubs
+from grr_response_server import flow_responses
+
+
+TRUNCATED_ROWS_COUNT = 10
+
+def _TruncateRows(table: rdf_osquery.OsqueryTable) -> rdf_osquery.OsqueryTable:
```
panhania, line 17: I would move this to a method on `OsqueryTable`. If you do, then `Truncate` sounds very imperative, as if it was doing this update in-place. For example, in Python `foo.sort()` will sort in-place, but `sorted(foo)` will construct a new, sorted object. If you apply both this suggestions, then the code would look like:

```python
foo = bar.Truncated()
```
rather than

```python
bar = _TruncateTable(foo)
```

which is just prettier IMO.

I would also make the number of rows to truncate an optional parameter (otherwise it would be impossible to test it).

panhania (changes requested):

grr/server/grr_response_server/flows/general/osquery_test.py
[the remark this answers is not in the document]
panhania: > Then, when I read the class definition I skim through all the private/protected stuff, getting an idea of what is there, and then I can read through public implementation with that additional context in mind.

Here, I agree with @simstoykov. Why should you need to read the private and protected parts of the class? Public interface is all you in general should care about. Sometimes it is not (if the code is poorly documented), but we should write code with assumption that we write good code, not a bad one.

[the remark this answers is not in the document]
max-vogler: IMHO, both approaches make sense, both "public members first" and "c-style bottom-up definition". Since we do not have a guideline or style guide regarding this, I would favour accepting any sensible grouping of class members in this PR.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -147,48 +152,108 @@ def testFailure(self):
...
+
+  def _RunQuery(self, query):
+    session_id = self._NewTestFlowStartAndRun(query)
     return flow_test_lib.GetFlowResults(self.client_id, session_id)
 
+  def _AssertTablesMatch(
+    self,
+    table: rdf_osquery.OsqueryTable,
+    expected: List[OrderedDict]
+  ) -> None:
```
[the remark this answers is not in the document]
simstoykov: Ultimately I think it's a trade-off, and it's a matter of taste which one is better. 
Benefits of "AssertTablesMatch":
1. To me the test bodies are more readable with "AssertTablesMatch" as opposed to having a few assertions. 
2. If a test fails, the stack trace will include "AssertTablesMatch", so it is obvious that the problem is in the tables. Whereas with plain assertions we get no semantic context.

Downsides:
1. If a test fails, you need to look into the implementation of AssertTablesMatch, and figure out what is failing.
I'm biased but it's not difficult to decode `self.assertEqual(table_col.name, expected_col_name)` or `self.assertEqual(table_col_values, expected_col_vlues)`. (Jokes on me, I have a mistype)

I don't feel too strongly about it, though, so I can revert your code to how it used to be.

For mine, however, I will need a minimum of 10 additional assertions in the non-truncating test, and 11 in the other one. And what if we decide to increase the truncating length to 20? I need the automation.

What are your thoughts?

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: Another thing is that the `[{ "foo": "quux", "baz": "thud", "bar": "norf" }] * 42` sample input wouldn't capture errors such as reordering of rows, as all of them are the same.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -147,48 +152,108 @@ def testFailure(self):
...
+  ]
 
+
+class FakeOsqueryFlowTest(flow_test_lib.FlowTestsBaseclass):
   def setUp(self):
     super(FakeOsqueryFlowTest, self).setUp()
     self.client_id = self.SetupClient(0)
 
-  def _RunQuery(self, query):
+  def _NewTestFlowStartAndRun(self, query):
```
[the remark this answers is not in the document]
simstoykov: Sorry I overlooked - I didn't rename `_RunQuery`, I separated some of its functionality in `_NewTestFlowStartAndRun`. But it's true that it wasn't obvious to me in the beginning that it starts a new flow. Does renaming it to `_RunQueryInNewFlow` sound good to you, or is it too wordy?

In general the scope of the two functions is not big, so I agree that maybe shorter names would be a better fit. I just prefer to be more explicit in the names, so that the code reads like a story. What do you think would be the best names for the two functions?

grr/server/grr_response_server/flows/general/osquery.py
panhania: In our codebase the convention is to use `TODO` comments. The format is `TODO(username)` (where `username` is the person that created or should deal with the issue at some point) and `TODO(link)` where `link` is a hyperlink to a bug report or an issue.

grr/server/grr_response_server/flows/general/osquery.py
panhania: Yeah, in that case it should stay as it is. Resolved.

grr/server/grr_response_server/flows/general/osquery_test.py
panhania: > However, what if a test fails because some cell values disagree?

If values are relevant they should not be introduced next to the place where you make assertions, otherwise you make assumption on what the format the helper is using, forcing the reader to familiarize themselves with the helper code (which is bad for the reasons explained in the next paragraph).

> Isn't it readable enough if the complexity is hidden behind helper functions?

If the reader has to jump around the code in order to understand what it is happening, it is not readable. Tests should read like linear scenario and be obvious. Here, I had to spend significant amount of time to understand what this helper is doing.

Moreover, it is not only about the readability. The more complex the helper is, the more likely you are to end up in a scenario where your generated data is invalid, e.g. it is empty and your tests passes trivially. Ultimately, in such cases, you should write tests for the helper itself (and at this point you not only introduce complexity, but also you have written more code than just inlining some input in the test code itself).

> Another thing is that the [{ "foo": "quux", "baz": "thud", "bar": "norf" }] * 42 sample input wouldn't capture errors such as reordering of rows, as all of them are the same.

Tests should be minimal and written for a particular purpose. If you want to test whether your code preserves the order, you should write small test case with pre-made order-sensitve input. If you want to test whether the code works with large input, then you can do `[single_row] * 42` and it still tests what it should.

grr/server/grr_response_server/flows/general/osquery_test.py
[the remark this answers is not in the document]
panhania: > To me the test bodies are more readable with "AssertTablesMatch" as opposed to having a few assertions.

I would not count it as a benefit, because it is not objective. For me it is less readable and forces me to execute code in my head and jump around the code instead of reading in in a linear way.

> If a test fails, the stack trace will include "AssertTablesMatch", so it is obvious that the problem is in the tables. Whereas with plain assertions we get no semantic context.

The name of the test case should be everything that you need. You are testing tables, so obviously the problem will be with the tables. If you test more then one thing, then your test is wrong and should be split.

> I don't feel too strongly about it, though, so I can revert your code to how it used to be.

I kind of to, so I would prefer to revert to the old code.

> For mine, however, I will need a minimum of 10 additional assertions in the non-truncating test, and 11 in the other one

10 or 11 is within reason, IMO.

>  And what if we decide to increase the truncating length to 20? I need the automation.

That's kind of the problem here that you depend on an implementation detail. Truncating length should be a parameter and in the tests you should cover two scenarios:

  * Use some small truncating length to write tests with particular values that are asserted.
  * Use some big truncating length to assert have some vague assertion that the "shape" is still preserved.

PR #879, opened 2020-10-30
Add support for displaying the first table rows from the OsqueryProgress

simstoykov:


simstoykov:
Sorry, this PR turned out larger than expected. Let me know if you want to talk about the changes if that will make reviewing easier.

simstoykov:
Screenshot 2020-10-30 at 12 54 45 PM

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
```diff
@@ -0,0 +1,208 @@
...
+    columns: ReadonlyArray<OsqueryColumn>,
+  },
+  rows: ReadonlyArray<OsqueryRow>
+};
+
+function newOsqueryTable(): FullMutableOsqueryTable {
+  return {
+    query: '',
+    header: {
+      columns: [] as ReadonlyArray<OsqueryColumn>,
```
mbushkov, line 20: Why explicit casting to ReadonlyArray is necessary here?

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_results_table.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+@Component({
+  selector: 'osquery-results-table',
+  templateUrl: './osquery_results_table.ng.html',
+  styleUrls: ['./osquery_results_table.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OsqueryResultsTable {
+  @Input()
+  tableToDisplay?: OsqueryTable;
```
max-vogler, line 15: `toDisplay` looks redundant on first glance.

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
```diff
@@ -0,0 +1,208 @@
...
+    },
+    rows: [] as ReadonlyArray<OsqueryRow>,
+  };
+}
+
+export class OsqueryTableBuilder {
```
mbushkov, line 26: Conceptually using a builder concept feels nice, but the size of the builder classes is scary. It's 150 lines - comparable to the size of the test itself. Also, the level of verbosity/complexity feels somewhat javaesque to me. I think we can rely on Javascript/Typescript features to express a very similar pattern.

What you can do is define:
```
const EMPTY_OSQUERY_TABLE: OsqueryTable {
    query: '',
    header: {
      columns: [],
    },
    rows: [],
};
```

For example:
```
const tableBuilder = new OsqueryTableBuilder()
      .withColumns(columns)
      .withValues(values);
```

can be expressed as

```
{
   ...EMPTY_OSQUERY_TABLE,
   header: {
      columns,
   },
   rows: values.map((rowValues) => ({values: rowValues}))
}
```

if you introduce a single helper function `valuesToRows`, it will look even better:

```
{
   ...EMPTY_OSQUERY_TABLE,
   header: {
      columns,
   },
   rows: valuesToRows(values)
}
```

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
```diff
@@ -0,0 +1,208 @@
...
+import { newFlowListEntry, newFlowResultSet } from '@app/lib/models/model_test_util';
+import { DeepMutable } from '@app/lib/type_utils';
+import { FlowListEntry, FlowState, FlowResultSet } from '@app/lib/models/flow';
+import { DebugElement } from '@angular/core';
+import { By } from '@angular/platform-browser';
+
+type FullMutableOsqueryTable = {
+  query: string,
+  header: {
+    columns: ReadonlyArray<OsqueryColumn>,
```
max-vogler, line 11: QQ: Why does `FullMutable*` contain `ReadonlyArray`?

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -1,34 +1,31 @@
...
+    <osquery-results-table
+      [tableToDisplay]="resultsTable">
+    </osquery-results-table>
   </div>
-</div>
+
+  <ng-template #noResultsTable>
+    <div *ngIf="progressTable$ | async as progressTable" class="progress-table">
+      <div *ngIf="additionalRowsAvailable$ | async as additionalRowsAvailable" class="show-additional">
+        <span class="show-additional-text">{{additionalRowsAvailable}} row(s) hidden.</span>
```
max-vogler, line 22: Use [`i18nPlural`](https://angular.io/api/common/I18nPluralPipe) instead of `row(s)`.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -15,58 +15,63 @@ import { isNonNull } from '@app/lib/preconditions';
...
+    map(progress => progress.partialTable),
+  );
+
+  readonly additionalRowsAvailable$ = this.osqueryProgress$.pipe(
+    map(progress => {
+      const progressTableLength = progress.partialTable?.rows?.length;
+      const fullTableLength = progress.totalRowsCount;
+
+      if (isNonNull(progressTableLength) && isNonNull(fullTableLength)) {
+        return BigInt(fullTableLength) - BigInt(progressTableLength);
```
max-vogler, line 57: BigInt is only needed if your numbers are bigger than `Number.MAX_SAFE_INT` (~9P). In this case it is safe to parse the decimal string `progress.totalRowsCount` as `Number` instead of `BigInt`.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -15,58 +15,63 @@ import { isNonNull } from '@app/lib/preconditions';
...
-  ngOnDestroy() {
-    this.unsubscribe$.next();
-    this.unsubscribe$.complete();
+  private loadCompleteResults() {
+    this.queryFlowResults({offset: 0, count: 1}); // TODO: Fetch more chunks if present
   }
 
-  trackByIndex(index: number, item: unknown): number {
-    return index;
+  fullTableRequested() {
```
max-vogler, line 74: Use imperative over past tense, you could also call `loadCompleteResults()` directly?

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  buildWithQueryIfDefined(query: string): FullMutableOsqueryTable | undefined {
+    this.withQueryIfDefined(query);
+    return this.build();
+  }
+}
+
+/** Helper class to build a FlowListEntry objects in a declarative manner */
+export class OsqueryFlowListEntryBuilder {
```
mbushkov, line 76: The following code

```
const testFlowListEntry = new OsqueryFlowListEntryBuilder()
  .withFlowState(FlowState.FINISHED)
  .withProgressTable([testColName], [[testCellValue]])
  .withProgressRowsCount(1)
  .withQuery(testQuery);
```

should effectively be equivalent to:
```
{
  ...newFlowListEntry({
    state: FlowState.FINISHED,
    progress: {
      totalRowsCount: 1,
      partialTable: {
        ...EMPTY_OSQUERY_TABLE,
        header: {
          [testColName],
        },
        rows: [{values: [testCellValue]}]
      }
    }
  }),
  
  query: testQuery,
}
```

This is slightly more complicated than the builder variant but also much more explicit and you don't need the 150 lines of untested (and fairly complicated) builder code. 

Probably we can define newOsqueryResultsTable(t: Partial<OsqueryResultsTable>) helper function to save a few lines of code, but I wouldn't go far beyond that. TypeScript's ellipsis (`...`) operator - something that doesn't exist in Java (last time I checked) should make building new test objects generally easy and reduces the need for the super-verbose builder pattern.

mbushkov (changes requested):
The code itself looks good! My main concern is bloated testing code. Please see my comments below.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
```diff
@@ -0,0 +1,208 @@
...
+  private includeProgress(): void {
+    const progress = {
+      totalRowsCount: this.progressRowsCount,
+      partialTable: this.progressTableBuilder.buildWithQueryIfDefined(this.query),
+    };
+    this.flowListEntry.flow.progress = progress;
+  }
+}
+
+export function elementBySelector(selector: string, root: DebugElement): DebugElement {
```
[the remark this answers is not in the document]
simstoykov: To me it's a bit awkward with or without the function, but I will remove it so that the code is consistent with the rest of the codebase. Also using `element.query(By.css('selector'))` is a general practice in testing, so I expect that the reader of the code knows what's going on and a separate function wouldn't add to the clarity too much.

For the record, I think there are two (subjective) benefits:
1. When we have many uses of it, `elementBySelector('foo bar', this.root)` stands out and the intent is clearer, whereas `this.root.query(By.css(selector))` initially looks like a random sequence of accessors (so it blends in with stuff like `.innerText` and others).
2. The `root` might be undefined in some cases, so we need optional chaining `?.`. This clutters the code a bit and there's some danger to miss it.

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
simstoykov: `[]` has inferred type `never[]`, because as far as Typescript is concerned, it is always an empty array.

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_test_utils.ts
simstoykov: Yeap I see your point; I will refactor this and remove the builder. But can we appreciate for the last time how beautiful, declarative and concise tests are? I'm probably too attached to this code.

simstoykov:
@max-vogler @mbushkov thanks for the reviews! Can you PTAL? 

I also refactored the osquery_details template trying to improve error handling: for example I noticed that if the progress table is undefined the user has no way of requesting the full table. While this shouldn't happen, I figured it's good to leave the button there in case something goes wrong.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details_test.ts
```diff
@@ -1,111 +1,75 @@
...
+ * Builds an OsqueryTable
+ * @param query The Osquery query which produced this table
+ * @param columns Column names of the table
+ * @param rows Array of arrays containing row values
+ */
+export function newOsqueryTable(
+  query: string,
+  columns: ReadonlyArray<string>,
+  rows: ReadonlyArray<ReadonlyArray<string>>,
+) {
```
mbushkov, line 23: Probably makes sense to specify the return type here, so that the type checker can catch the errors early.

mbushkov: Another note - if a test function is shared between two tests, let's put it in a third file. So maybe, for this purpose, we'd need to resurrect the osquery_test_utils.ts that was in this PR earlier. Primary reason: more rigid build systems, like Bazel, are sometimes not happy when a single source file has a dual purpose (a test and a library file with exported functions).

Alternatively, we can put this function into https://github.com/google/grr/blob/master/grr/server/grr_response_server/gui/ui/lib/models/model_test_util.ts

mbushkov (commented):
Looks great! Thanks for reworking the code. LGTM, but please see my 2 comments below. 
Letting @max-vogler to formally LGTM and actually merge as soon as he's happy with the code too.

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_results_table_test.ts
```diff
@@ -1,8 +1,24 @@
...
-import { OsqueryTableBuilder, ParsedDetailsTable } from './osquery_test_utils';
 import { DebugElement } from '@angular/core';
 import { OsqueryResultsTable } from './osquery_results_table';
+import { By } from '@angular/platform-browser';
+import { OsqueryTable } from '@app/lib/api/api_interfaces';
+import { newOsqueryTable } from '../plugins/osquery_details_test';
+
+/** Helper data structure to parse an osquery_results_table */
+class OsqueryResultsTableDOM {
+  queryDiv = this.rootElement.query(By.css('.results-query-text'));
```
max-vogler, line 11: Generally use `readonly` if possible. Restricting mutability is desirable to make code easier to understand.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
```diff
@@ -6,26 +6,38 @@
...
+
+    <ng-template #noResultsTable>
+      <div *ngIf="progressTable$ | async as progressTable; else noTables" class="progress-table">
+        <osquery-results-table
+          [table]="progressTable">
+        </osquery-results-table>
+
+        <ng-container *ngIf="(additionalRowsAvailable$ | async) !== 0">
+          <div
+          *ngIf="additionalRowsAvailable$ | async as additionalCount; else progressCountUndefined"
```
max-vogler, line 28: This ngIf is duplicated and looks a bit like code smell. Maybe you can change additionalRowsAvailable$ to never be undefined?

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ng.html
simstoykov: I'm actually using the undefined value - the case that I try to cover is if for some reason `totalRowCount` in the progress is undefined, and the values in `additionalRowsAvailable$` are too, in which case something went wrong and we still want to display the "View all" button.

I'm not sure what would be a cleaner approach.

max-vogler: I suggest simplifying the code: Remove #progressCountUndefined to re-use one `button` for both cases. This button can then show `({{additionalCount}} more)` if they exist or hide the "more" count, if 0 or undefined.

From a users point of view "Results might be missing or incomplete" does not provide any new actionable information.

grr/server/grr_response_server/gui/ui/components/flow_details/helpers/osquery_results_table.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+@Component({
+  selector: 'osquery-results-table',
+  templateUrl: './osquery_results_table.ng.html',
+  styleUrls: ['./osquery_results_table.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OsqueryResultsTable {
+  @Input()
+  table?: OsqueryTable;
```
max-vogler, line 15: Please handle the nullability of `table`. This input shows table as potentially undefined, but the template expects it to be always set.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -15,58 +15,59 @@ import { isNonNull } from '@app/lib/preconditions';
...
+    flatMap(singleResultSet => singleResultSet?.items),
     filter(isNonNull),
-    map((singleItem) => singleItem.payload as OsqueryResult),
+    map(singleItem => singleItem.payload as OsqueryResult),
   );
 
-  readonly resultTable$ = this.osqueryResult$.pipe(
-    map((result) => result?.table),
+  readonly resultsTable$ = this.osqueryResults$.pipe(
+    map(result => result?.table),
```
max-vogler, line 35: Please properly handle nullability: `osqueryResults$` emits non-nullable values only, but is accessed with `?` here and below.

max-vogler (changes requested):

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -31,36 +27,53 @@ export class OsqueryDetails extends Plugin {
...
-  readonly resultsStderr$ = this.osqueryResults$.pipe(
-    map(result => result?.stderr),
-  );
-
   readonly osqueryProgress$: Observable<OsqueryProgress> = this.flowListEntry$.pipe(
     map(listEntry => listEntry.flow.progress as OsqueryProgress),
     filter(isNonNull),
   );
 
+  readonly resultsTable$ = this.osqueryResults$.pipe(
```
max-vogler, line 35: Please mark all possible members as private, to make the public interface of the component class apparent to the reader of your code.

grr/server/grr_response_server/gui/ui/components/flow_details/plugins/osquery_details.ts
```diff
@@ -31,36 +27,53 @@ export class OsqueryDetails extends Plugin {
...
-  readonly additionalRowsAvailable$ = this.osqueryProgress$.pipe(
-    map(progress => {
-      const progressTableLength = progress.partialTable?.rows?.length;
-      const fullTableLength = progress.totalRowCount;
+  readonly displayTable$ = concat(
+    this.progressTable$.pipe(takeUntil(this.resultsTable$)),
+    this.resultsTable$,
+  );
+
+  readonly totalAndDisplayedRowCount = combineLatest([
```
max-vogler, line 50: `combineLatest` only emits if all inner observables also emitted at least once. If this is unexpected, use `startWith()` to provide a starting value.

max-vogler, line 50: `totalAndDisplayedRowCount` is only used in `additionalRowsAvailable$`. Please inline it to make the code more readable.

max-vogler (changes requested):

max-vogler (approved):

PR #880, opened 2020-11-03
Adding backend support for truncated Osquery result in progress (second)

simstoykov:


grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -12,16 +12,14 @@
...
-  result.header = table.header
-  result.rows = table.rows[:TRUNCATED_ROWS_COUNT]
-
-  return result
+def _GetTotalRowCount(
+  responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
+) -> int:
+  project_length = lambda cur_response: len(cur_response.table.rows)
+  row_lengths = map(project_length, responses)
+  return sum(row_lengths)
```
panhania, line 22: Initially, I read `project_length` as "length of a project" rather than "make a projection of length". I would rename it to `row_count` (or `table_row_count`) and then remove `row_lengths` variable.

Also, no need for the `cur_` prefix, in the lambda parameter (since there is no "previous", or "next" or anything alike to make it more meaningful).

```python
  row_count = lambda response: response.table.rows
  return sum(map(row_count, responses))
```

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -244,16 +245,36 @@ def testProgressTruncates20RowsTo10(self):
...
     self._AssertTablesMatch(progress.partial_table, test_table[:10])
 
+  def testTotalRowCountIncludesAllChunks(self):
+    row_count = 100
+    split_pieces = 10
+
+    # Not using _CreateDictTable here in order to control the exact size of the
+    # test table in bytes, and thus control the number of chunks
+    cell_value = 'fixed'
+    table = [{'column1': cell_value} for _ in range(row_count)]
```
panhania, line 265: This can be simplified to` [{'column1': cell_value}] * row_count` (it's fine, since it is used as immutable, constant value).

simstoykov:
@panhania PTAL

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -152,15 +150,12 @@ def testFailure(self):
...
-    OrderedDict(
-      CellValue(col_number, row_number) for col_number in range(column_number)
-    ) for row_number in range(row_number)
-  ]
+@contextmanager
+def PatchTruncationLimitTo(new_value: int):
+  old_value = osquery_flow.TRUNCATED_ROW_COUNT
+  osquery_flow.TRUNCATED_ROW_COUNT = new_value
+  yield
+  osquery_flow.TRUNCATED_ROW_COUNT = old_value
```
panhania, line 158: No need for this helper, `mock.patch.object` has this functionality (e.g. see [here][1]).

[1]: https://github.com/google/grr/blob/fe2563069b0a8c6fdea3fe24a3c938c16391a5e3/grr/server/grr_response_server/artifact_test.py#L517-L518

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -14,14 +14,35 @@
...
-  project_length = lambda cur_response: len(cur_response.table.rows)
-  row_lengths = map(project_length, responses)
+  get_row_lengths = lambda response: len(response.table.rows)
+  row_lengths = map(get_row_lengths, responses)
   return sum(row_lengths)
 
 
+def _GetTruncatedTable(
+  responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
+) -> rdf_osquery.OsqueryTable:
```
panhania, line 28: This function has a lot of not obvious logic, a docstring is mandatory.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -169,7 +164,7 @@ def setUp(self):
     super(FakeOsqueryFlowTest, self).setUp()
     self.client_id = self.SetupClient(0)
 
-  def _NewTestFlowStartAndRun(self, query):
+  def _InitializeFlow(self, query):
```
panhania, line 167: `_InitializeFlow` and `_RunQuery` are very much alike (except that the other one also collects results). How about renaming them to `_InitializeOsqueryFlow` and `_RunOsqueryFlow`? Then both have consistent naming hinting similar responsibilities (with the second one doing a bit more).

We might want to drop the `Osquery` part (since it is kind of obvious in tests for osquery), then `_RunQuery` would just become `_RunFlow`.

What do you think?

panhania (approved):

PR #885, opened 2020-11-20
Export Osquery Flow results as CSV

simstoykov:


api_client/python/grr_api_client/flow.py
```diff
@@ -105,6 +106,14 @@ def GetCollectedTimeline(
         client_id=self.client_id, flow_id=self.flow_id, format=fmt)
     return self._context.SendStreamingRequest("GetCollectedTimeline", args)
 
+  def GetOsqueryResults(
+      self,
+      fmt: osquery_pb2.ApiGetOsqueryResultsArgs.Format
```
panhania, line 111: Missing trailing comma. Without the internal auto-formatter can fold it in awkward ways.

grr/core/grr_response_core/lib/rdfvalues/osquery.py
```diff
@@ -99,6 +99,13 @@ class OsqueryResult(rdf_structs.RDFProtoStruct):
   protobuf = osquery_pb2.OsqueryResult
   rdf_deps = [OsqueryTable]
 
+  def GetTableColumns(self):
+    return [column.name for column in self.table.header.columns]
+  
+  def GetTableRows(self):
+    return [row.values for row in self.table.rows]
```
panhania, line 106: Prefer to use type annotations in all newly written code. Also, do you need to return a list? Wouldn't iterator suffice and let the caller decide whether to allocate a list out of it?

grr/proto/grr_response_proto/api/osquery.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto2";
+
+package grr;
+
+import "grr_response_proto/semantic.proto";
+
+message ApiGetOsqueryResultsArgs {
+  enum Format {
+    CSV = 0;
```
panhania, line 9: The default ("0") value of an enum has to be always named "UNKNOWN", "UNDEFINED", "DEFAULT" or alike.

grr/server/grr_response_server/gui/api_call_router.py
```diff
@@ -725,6 +726,19 @@ def UnscheduleFlow(
     """Unschedules and deletes a previously scheduled flow."""
     raise NotImplementedError()
 
+  @Category("Flows")
+  @ArgsType(api_osquery.ApiGetOsqueryResultsArgs)
+  @ResultBinaryStream()
+  @Http("GET", 
+      "/api/clients/<client_id>/flows/<flow_id>/osquery-results/<format>")
```
panhania, line 733: The previous method aligns arguments, I think this should follow the already established style.

grr/server/grr_response_server/gui/api_call_router_without_checks.py
```diff
@@ -219,6 +220,13 @@ def UnscheduleFlow(
   ) -> api_flow.ApiUnscheduleFlowHandler:
     return api_flow.ApiUnscheduleFlowHandler()
 
+  def GetOsqueryResults(
+      self,
+      args: api_osquery.ApiGetOsqueryResultsArgs,
+      context: Optional[api_call_context.ApiCallContext] = None,
+  ):
+    return api_osquery.ApiGetOsqueryResultsHandler()
```
panhania, line 228: Missing:

```python
del args, context  # Unused.
```

grr/server/grr_response_server/gui/api_integration_tests/osquery_test.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
```
panhania, line 1: I don't think this is needed.

grr/server/grr_response_server/gui/api_integration_tests/osquery_test.py
```diff
@@ -0,0 +1,101 @@
...
+# Lint as: python3
+"""Integration tests for the Osquery flow, its api client and api endpoints."""
```
panhania, line 3: If you capitalize "osquery" then I think "API" also should be written in uppercase.

grr/server/grr_response_server/gui/api_integration_tests/osquery_test.py
```diff
@@ -0,0 +1,101 @@
...
+from grr.test_lib import test_lib
+from grr.test_lib import action_mocks
+
+from grr_response_server.flows.general import osquery as osquery_flow
+
+from grr_response_client.client_actions import osquery as osquery_action
+
+from grr_response_proto.api import osquery_pb2 as api_osquery_pb2
+
+from grr_api_client import utils
```
panhania, line 16: I would not put so many blank lines. I guess you tried to group by package but the convention within our codebase is grouping into 3 groups:

  * standard library imports
  * third-party imports
  * GRR imports

grr/server/grr_response_server/gui/api_plugins/osquery.py
```diff
@@ -0,0 +1,115 @@
...
+  added_columns = False
+
+  for result in osquery_results:
+    if not added_columns:
+      added_columns = True
+      columns = result.GetTableColumns()
+      yield _ListToCsvBytes(columns)
+
+    row_bytes = [_ListToCsvBytes(row) for row in result.GetTableRows()]
+    yield from row_bytes
```
panhania, line 105: No need to allocate a list only to yield from it:

```python
yield from map(_ListToCsvBytes, result.GetTableRows())
```

grr/server/grr_response_server/gui/api_plugins/osquery.py
```diff
@@ -0,0 +1,115 @@
...
+
+    for datum in data_fetched:
+      if type(datum.payload) != rdf_osquery.OsqueryResult:
+        raise ValueError(f"Incorrect payload type: {type(datum.payload)}")
+      yield datum.payload
+
+
+def _ParseToCsvBytes(
+    osquery_results: Generator[rdf_osquery.OsqueryResult, None, None]
+) -> Generator[bytes, None, None]:
```
panhania, line 95: `Iterator[bytes]`

grr/server/grr_response_server/gui/api_plugins/osquery.py
```diff
@@ -0,0 +1,115 @@
...
+    next_to_fetch += last_fetched_count
+
+    for datum in data_fetched:
+      if type(datum.payload) != rdf_osquery.OsqueryResult:
+        raise ValueError(f"Incorrect payload type: {type(datum.payload)}")
+      yield datum.payload
+
+
+def _ParseToCsvBytes(
+    osquery_results: Generator[rdf_osquery.OsqueryResult, None, None]
```
panhania, line 94: `Iterator[rdf_osquery.OsqueryResult]`

grr/server/grr_response_server/gui/api_plugins/osquery.py
```diff
@@ -0,0 +1,115 @@
...
+      columns = result.GetTableColumns()
+      yield _ListToCsvBytes(columns)
+
+    row_bytes = [_ListToCsvBytes(row) for row in result.GetTableRows()]
+    yield from row_bytes
+
+
+def _ListToCsvBytes(values: List[str]) -> bytes:
+  csv_text = ','.join(values) + '\n'
+  return csv_text.encode('utf-8')
```
panhania, line 110: That looks pretty dangerous, what about escaping? Why not use the standard `csv` module?

panhania (changes requested):

grr/server/grr_response_server/gui/api_plugins/osquery.py
simstoykov: `Iterator` is a super-type of `Generator` so `Iterator[bytes]` is technically correct, however this function actually returns a `Generator` from my experiments (`<generator object _ParseToCsvBytes at 0x7f89525b8740>`), and also the [Python docs](https://docs.python.org/3.6/glossary.html#term-generator).

grr/server/grr_response_server/gui/api_call_router_without_checks.py
simstoykov: Why do we do that exactly?

grr/server/grr_response_server/gui/api_plugins/osquery.py
simstoykov: Okay I agree that `Iterator[bytes]` is the right choice here. To answer my own question - yes the exact type is `Generator[bytes, None, None]`, however this is just an implementation detail; the intended exposed interface only needs an `Iterator` to iterate over.

grr/server/grr_response_server/gui/api_integration_tests/osquery_test.py
```diff
@@ -0,0 +1,96 @@
...
+      { "🇬 🇷 🇷": "🔝🔝🔝"}
+    ]
+    """
+
+    results_iterator = self._RunOsqueryExportResults(stdout)
+    csv_bytes = next(results_iterator)
+
+    expected_text = "🇬 🇷 🇷\r\n🔝🔝🔝\r\n"
+    expected_bytes = expected_text.encode("utf-8")
+    self.assertEqual(expected_bytes, csv_bytes)
```
panhania, line 77: The assertions would read better and, in case of a failure, the error messages are human-friendly if you decode CSV bytes rather than encoding expectations:

```python
    output = next(results_iterator).decode("utf-8")
    self.assertEqual(output, "🇬 🇷 🇷\r\n🔝🔝🔝\r\n")
```

This remark applies also below.

grr/server/grr_response_server/gui/api_plugins/osquery_test.py
```diff
@@ -0,0 +1,50 @@
...
+from grr_response_core.lib.rdfvalues import osquery as rdf_osquery
+
+
+class UtilsTest(unittest.TestCase):
+
+  def testListToCSVBytes(self):
+    received_bytes = api_osquery._LineToCsvBytes(["a", "b", "c", "d"])
+
+    csv_correct_bytes = "a,b,c,d\r\n".encode("utf-8")
+    self.assertEqual(received_bytes, csv_correct_bytes)
```
panhania, line 15: Ditto (here and below).

simstoykov:
@panhania PTAL

grr/server/grr_response_server/gui/api_plugins/osquery.py
```diff
@@ -0,0 +1,118 @@
...
+      offset=next_to_fetch,
+      count=_RESULTS_TO_FETCH_AT_ONCE,
+      client_id=client_id,
+      flow_id=flow_id)
+
+    last_fetched_count = len(data_fetched)
+    next_to_fetch += last_fetched_count
+
+    for datum in data_fetched:
+      if type(datum.payload) != rdf_osquery.OsqueryResult:
```
panhania, line 88:
```python
if not isinstance(datum.payload, rdf_osquery.OsqueryResult)
```

grr/server/grr_response_server/gui/api_integration_tests/osquery_test.py
```diff
@@ -0,0 +1,93 @@
...
+        token=self.token,
+        query="doesn't matter")
+      result_flow = self.api.Client(client_id=client_id).Flow(flow_id)
+      result_flow.WaitUntilDone()
+
+    format_csv = api_osquery_pb2.ApiGetOsqueryResultsArgs.Format.CSV
+    return result_flow.GetOsqueryResults(format_csv)
+
+  def testExportSomeResults(self):
+
```
panhania, line 36: Unnecessary newline.

panhania (commented):
Looks almost good, only two nitpicky comments left.

panhania (approved):

mbushkov (approved):

PR #891, opened 2020-11-25
Added server and port configurations for GRRafana server

tsehori:
Made the server and port of GRRafana server configurable. For reference, this is similar to how the [default crash limit](https://github.com/google/grr/blob/0e03976c5c5b694c210c0f392581364ddefa1f4b/grr/server/grr_response_server/rdfvalues/hunts.py#L83) of hunts [are configured](https://github.com/google/grr/blob/fe2563069b0a8c6fdea3fe24a3c938c16391a5e3/grr/core/grr_response_core/config/server.py#L246).

I do have one small question though - do we want to let these settings be set up during `grr_config_updater initialize` as well, similar to the way the [MySQL port](https://github.com/google/grr/blob/fe2563069b0a8c6fdea3fe24a3c938c16391a5e3/grr/server/grr_response_server/bin/config_updater.py#L127) is set up?

### DIFF
--- a/grr/core/grr_response_core/config/server.py
+++ b/grr/core/grr_response_core/config/server.py
@@ -282,6 +282,15 @@ config_lib.DEFINE_integer(
     "If the average network usage per client becomes "
     "greater than this limit, the hunt gets stopped.")
 
+# GRRafana HTTP Server settings.
+config_lib.DEFINE_string("GRRafana.server",
+                         default="localhost",
+                         help="The GRRafana server address.")
+
+config_lib.DEFINE_integer("GRRafana.port",
+                          default=5000,
+                          help="The GRRafana server port.")
+
 # Fleetspeak server-side integration flags.
 config_lib.DEFINE_string(
     "Server.fleetspeak_message_listen_address", "",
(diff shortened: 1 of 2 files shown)


grr/core/grr_response_core/config/server.py
```diff
@@ -282,6 +282,15 @@
     "If the average network usage per client becomes "
     "greater than this limit, the hunt gets stopped.")
 
+# GRRafana HTTP Server settings.
+config_lib.DEFINE_string("GRRafana.server",
```
mbushkov, line 286: I'd suggest using "bind" here as is in https://github.com/google/grr/blob/252af5d85f02121a75f8d83ebbee8e3d7e4f5fc9/grr/test/grr_response_test/test_data/grr_test.yaml#L238.

mbushkov, line 286: I'm also increasingly concerned about the name. Basically: `grrafana` and `grafana` are indistinguishable when reading the text and relying on upper casing to highlight the difference feels a bit odd. Even though I like the name, I think we should go with something conventional, like `GrafanaSourceServer` or `GrafanaDataSourceServer`. (Not for this PR, but let's have a discussion on this)

mbushkov (changes requested):

tsehori:
PTAL
Let's discuss the naming in sync.

mbushkov (approved):

PR #894, opened 2020-11-27
Adding a basic editor with proper code editor font, line numbers, basic SQL syntax highlighting

simstoykov:
## What the user sees initially:
image

## Autocompletion pop-up:
Screenshot 2020-11-27 at 11 19 32 AM

## Syntax highlighting:
Screenshot 2020-11-27 at 11 21 38 AM

### DIFF
--- a/grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
+++ b/grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
@@ -11,6 +11,10 @@ import {
 } from '@angular/core';
 
 import * as CodeMirror from 'codemirror';
+// importing sql-hint is needed for SQL syntax highlighting
+import 'codemirror/addon/hint/sql-hint.js';
+// importing show-hint is needed for the autocomplete pop-up
+import 'codemirror/addon/hint/show-hint.js';
 
 
 /** Displays a code editor. */
@@ -34,7 +38,7 @@ export class CodeEditor implements AfterViewInit {
 
   private initializeEditor(): void {
     const editor = CodeMirror.fromTextArea(this.editorTarget.nativeElement, {
-      value: this.initialValue,
+      value: '',
       mode: 'text/x-sqlite',
       theme: 'idea',
       extraKeys: {'Ctrl-Space': 'autocomplete'},
@@ -45,5 +49,7 @@ export class CodeEditor implements AfterViewInit {
     editor.on('change', () => {
       this.latestValue.emit(editor.getValue());
     });
+
+    editor.setValue(this.initialValue);
   }
 }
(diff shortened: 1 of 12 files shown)


mbushkov (changes requested):
Looks really good! Please see a few of my comments below.

grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
```diff
@@ -0,0 +1,55 @@
...
+  templateUrl: './code_editor.ng.html',
+  styleUrls: ['./code_editor.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  encapsulation: ViewEncapsulation.None,
+})
+export class CodeEditor implements AfterViewInit {
+  @ViewChild('editorTarget')
+  private readonly editorTarget!: ElementRef;
+
+  @Input() initialValue = '';
```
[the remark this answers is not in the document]
simstoykov: Yeap I was planning to extend it with this functionality in the next PR containing the query helper, but we can also do it here.

It looks like the `propName` and `propNameChange` convention is useful when I want to bind a **variable** in the parent component to some property in the child component. However, here I need the stream of changes (not just the latest value in a variable), because I am calling `this.form.patchValue({ query: latestValue });` (and I didn't find a way to use two-way binding with form controls, something like `[(value)]="form.query"`?).

Can we make it so the parent component sets the value of the child code editor through a Subject like `[value]="setEditorValue$ | async"`, which is initialized in the following way: `readonly setEditorValue$ = new BehaviorSubject(this.defaultQueryDisplayed);`. 
On the other hand, the child code editor exposes a `valueChange` event emitter, to which the parent component listens to and updates the `form.query` value: `(valueChange)="updateQuery($event)"`. In the child component we can update the values as follows:

```
private latestValue = '';
  @Input()
  set value(newText: string) {
    this.latestValue = newText;

    this.valueChange.emit(this.latestValue);
  }
  get value(): string {
    return this.latestValue;
  }

  @Output() valueChange = new EventEmitter<string>();
```

grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
```diff
@@ -0,0 +1,84 @@
...
+export class CodeEditor implements AfterViewInit, ControlValueAccessor {
+  @ViewChild('editorTarget')
+  private readonly editorTarget!: ElementRef;
+  private editor?: CodeMirror.Editor;
+
+  private latestOverwrite = '';
+  @Input()
+  set editorValue(newValue: string) {
+    this.latestOverwrite = newValue;
+    this.editor?.setValue(newValue);
```
mbushkov, line 45: Do I understand it correctly that editor might be not set when this function is called? Let's add a comment why that might be the case.

grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+  get editorValue() {
+    return this.editor?.getValue() || '';
```
mbushkov, line 48: `??` is normally preferred over `||` in GRR's codebase.

mbushkov (approved):
Looks good, please see 2 minor comments below.

PR #899, opened 2020-12-07
Upgrade editor and prepare for the osquery helper

simstoykov:
This is how the final result looks like:
image

### DIFF
--- a/grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
+++ b/grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
@@ -7,10 +7,17 @@ import {
     ViewEncapsulation,
     Input,
     forwardRef,
+    OnDestroy,
+    HostBinding,
 } from '@angular/core';
-import { NG_VALUE_ACCESSOR, ControlValueAccessor } from '@angular/forms';
+import {NG_VALUE_ACCESSOR, ControlValueAccessor, NgControl} from '@angular/forms';
+import {MatFormFieldControl} from '@angular/material/form-field';
+import {Subject} from 'rxjs';
+import {FocusMonitor} from '@angular/cdk/a11y';
+import {takeUntil} from 'rxjs/operators';
+import {isNonNull} from '@app/lib/preconditions';
 
-import * as CodeMirror from 'codemirror';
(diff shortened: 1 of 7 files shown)


grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
```diff
@@ -74,14 +60,155 @@ export class CodeEditor implements AfterViewInit, ControlValueAccessor {
...
+
+  private readonly unsubscribe$ = new Subject<void>();
+
+  readonly controlType = 'code-editor';
+
+  @HostBinding()
+  get id(): string {
+    const uniqueId = `${this.controlType}-${EditorWithCvaAndMffc.uniqueNumber}`;
+    EditorWithCvaAndMffc.uniqueNumber += 1;
+    return uniqueId;
```
max-vogler, line 107: I'd expect a getter to return the same value always and not update internal state when it is called. If this assumption is false, please document in a comment.

Depending on how IDs are used, you can also use a GUID to avoid keeping `uniqueNumber` as state.

grr/server/grr_response_server/gui/ui/components/code_editor/code_editor.ts
```diff
@@ -74,14 +60,155 @@ export class CodeEditor implements AfterViewInit, ControlValueAccessor {
...
+  }
+  registerOnTouched(): void { }
+}
+
+/**
+ * Extends the editor with MatFormFieldControl functionality.
+ * This allows the editor to be used inside <mat-form-field></mat-form-field>
+ * tags.
+ */
+class EditorWithCvaAndMffc extends EditorWithCva
```
max-vogler, line 95: It's a bit hard to read to have 4 classes, 3 of them are only used once as base class for the next higher up class.  E.g. CodeEditor is only a minimal layer on EditorWithCvaAndMffc and EditorWithCva is only a minimal layer on CoreEditor. I suggest unifying at least those two to simplify reading this code and simplify the naming of the classes.

max-vogler (changes requested):

simstoykov:
@max-vogler can you PTAL

max-vogler (approved):

PR #900, opened 2020-12-07
Osquery helper to browse through tables and overwrite the query

simstoykov:
This is how the final result looks like:
Screenshot 2020-12-07 at 10 56 46 AM
Screenshot 2020-12-07 at 10 57 01 AM
Screenshot 2020-12-07 at 10 59 03 AM
Screenshot 2020-12-07 at 10 59 26 AM


grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper_test.ts
```diff
@@ -0,0 +1,61 @@
+import {newOsqueryTableSpec, newOsqueryColumnSpec} from './osquery_table_specs';
+import {QueryComposer} from './osquery_query_helper';
+
+fdescribe('QueryComposer', () => {
```
mbushkov, line 4: Let's make sure we don't submit fdescribe accidentally.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/table_info_item/table_info_item.ng.html
```diff
@@ -0,0 +1,26 @@
+<div class="item-header">
```
mbushkov, line 1: I think it's ok to put table_info_item.* files into osquery_query_helper/ folder without creating an additional subfolder. At the moment we don't have a rule of "one folder for one component".

mbushkov (changes requested):
Generally looks good - please see my comments below.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper.ts
```diff
@@ -0,0 +1,320 @@
...
+      return tableSpec;
+    });
+
+    return new TableCategory(categoryName, tableSpecs);
+  }
+}
+
+declare interface ValueWithMatcher {
+  readonly value: string;
+  readonly matcher: SingleElementFuzzyMatcher;
```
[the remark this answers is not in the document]
simstoykov: Actually this particular thing, apart from making it ineffective, is also what over-complicates the code in my opinion too. I thought there's no way to have 1 matcher and map back matches to table names, descriptions and columns, because `match(1)` returns match results not in the order that subjects were passed (and doesn't return match results for subjects which were didn't have at least 1):

I thought the dictionary approach wouldn't work if there are repeating values (e.g. column names can easily repeat), because we wouldn't be able to uniquely map matches to original subjects. Thinking of it now, if two subjects are equal, they will have the same matches, so there's no issue at all?

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper_test.ts
simstoykov: Good spot!

simstoykov:
@mbushkov can you PTAL. If we are generally happy with the structure of the code, I will proceed with writing tests.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_helper_interfaces.ts
```diff
@@ -0,0 +1,96 @@
+import {OsqueryTableSpec, nameToTable, OsqueryColumnSpec} from './osquery_table_specs';
```
mbushkov, line 1: `osquery_helper_interface.ts` sounds too generic (you could just as well say `osquery_helper_classes.ts` or `osquery_helper_functions.ts` :) )

You can call this `osquery_model.ts`, for example - here "model" stands for "data model". What do you think?

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_table_specs.ts
```diff
@@ -0,0 +1,52 @@
+import tablesJson from '../../../../static/osquery_table_specs_4_5_1.json';
```
mbushkov, line 1: Unfortunately this won't work with Bazel/Closure compiler. Can we wrap the data to make it a legitimate TS file?

grr/server/grr_response_server/gui/ui/tsconfig.json
```diff
@@ -29,7 +29,9 @@
     ],
     "paths": {
       "@app/*": ["./*"]
-    }
+    },
+    "resolveJsonModule": true,
```
mbushkov, line 33: Please see my comment above about the direct JSON import. Unfortunately we can't use this feature.

grr/server/grr_response_server/gui/ui/tsconfig.json
```diff
@@ -29,7 +29,9 @@
...
+    "esModuleInterop": true
```
mbushkov, line 34: Why do we need to turn on the esModuleInterop setting?

mbushkov (changes requested):
Looks really good! A few minor things.

grr/server/grr_response_server/gui/ui/tsconfig.json
simstoykov: This was required for the json import (not sure what exactly), but there's no need for it now.

simstoykov:
@mbushkov can you PTAL

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_helper_model.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export interface TableCategory {
+  readonly categoryName: string;
+  readonly tableSpecs: ReadonlyArray<OsqueryTableSpec>;
+  readonly subjects: ReadonlyArray<string>;
+}
+
+/**
+ * Same as {@link TableCategory}, but also holds a mapping from string subjects
+ * too match results.
```
mbushkov, line 18: Typo: too -> to.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper.scss
```diff
@@ -0,0 +1,17 @@
+// Styles declared here are effectively global because of ViewEncapsulation.None
```
mbushkov, line 1: Why do we need to have these styles defined as global? They all seem to be used inside osquery_query_helper.ng.html and not inside any of its child components.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper_test.ts
```diff
@@ -0,0 +1,111 @@
...
+import {OsqueryQueryHelperModule} from './module';
+import {TestBed, fakeAsync, tick} from '@angular/core/testing';
+import {nameToTable} from './osquery_table_specs';
+import {isNonNull} from '@app/lib/preconditions';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
+import {take} from 'rxjs/operators';
+
+initTestEnvironment();
+
+fdescribe('OsqueryQueryHelper component', () => {
```
mbushkov, line 12: Nit: fdescribe.

mbushkov (commented):
Looks almost done! Just a few comments/questions.

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper.scss
simstoykov: Because of `mat-autocomplete`. I will update the comment so that it describes the reason. This issue is relevant: https://github.com/angular/components/issues/11764#issuecomment-400370834

grr/server/grr_response_server/gui/ui/components/flow_args_form/osquery_query_helper/osquery_query_helper.scss
```diff
@@ -0,0 +1,17 @@
...
+// Make sure to narrow their names enough
```
mbushkov, line 2: Nit: let's end a sentence with a dot.

mbushkov (approved):
LGTM, 1 super-minor nit.

PR #901, opened 2020-12-09
Initial PR for collecting filepaths from columns in the Osquery table result (Backend)

simstoykov:
An early stage PR. 
There are some things which need to be considered/documented/changed:
* Files which exceed the single file limit are truncated, however trying to collect files which exceed the total size limit results in a flow error
* The directory structure in the archive mappings, and any escaping that we might need to do
* There isn't any record of which files were found and which were not; we just give a best-effort archive

### DIFF
--- a/grr/core/grr_response_core/lib/rdfvalues/osquery.py
+++ b/grr/core/grr_response_core/lib/rdfvalues/osquery.py
@@ -10,21 +10,7 @@ from typing import Sequence
 from typing import Text
 
 from grr_response_core.lib.rdfvalues import structs as rdf_structs
-from grr_response_core.lib.rdfvalues import client_fs as rdf_client_fs
 from grr_response_proto import osquery_pb2
-
-
-class OsqueryFlowArgs(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryFlowArgs` proto."""
-
-  protobuf = osquery_pb2.OsqueryFlowArgs
-  rdf_deps = []
-
-  def __init__(self, *args, **kwargs):
-    super().__init__(*args, **kwargs)
-
-    if not self.HasField("timeout_millis"):
-      self.timeout_millis = 5 * 60 * 1000  # 5 minutes.
 
 
 class OsqueryArgs(rdf_structs.RDFProtoStruct):
@@ -124,9 +110,3 @@ class OsqueryProgress(rdf_structs.RDFProtoStruct):
   protobuf = osquery_pb2.OsqueryProgress
   rdf_deps = [OsqueryTable]
 
-
-class OsqueryCollectedFile(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryCollectedFile` proto."""
-
-  protobuf = osquery_pb2.OsqueryCollectedFile
-  rdf_deps = [rdf_client_fs.StatEntry]
--- a/grr/core/grr_response_core/lib/rdfvalues/osquery.py
+++ b/grr/core/grr_response_core/lib/rdfvalues/osquery.py
@@ -10,6 +10,7 @@ from typing import Sequence
 from typing import Text
 
 from grr_response_core.lib.rdfvalues import structs as rdf_structs
+from grr_response_core.lib.rdfvalues import client_fs as rdf_client_fs
 from grr_response_proto import osquery_pb2
 
 
@@ -24,6 +25,13 @@ class OsqueryArgs(rdf_structs.RDFProtoStruct):
 
     if not self.HasField("timeout_millis"):
       self.timeout_millis = 5 * 60 * 1000  # 5 minutes.
+
+
+class OsqueryActionArgs(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryActionArgs` proto."""
+
+  protobuf = osquery_pb2.OsqueryActionArgs
+  rdf_deps = []
 
 
 class OsqueryColumn(rdf_structs.RDFProtoStruct):
@@ -57,7 +65,7 @@ class OsqueryTable(rdf_structs.RDFProtoStruct):
     """Iterates over values of a given column.
 
     Args:
-      column_name: A nome of the column to retrieve the values for.
+      column_name: A name of the column to retrieve the values for.
 
     Yields:
       Values of the specified column.
@@ -115,3 +123,10 @@ class OsqueryProgress(rdf_structs.RDFProtoStruct):
 
   protobuf = osquery_pb2.OsqueryProgress
   rdf_deps = [OsqueryTable]
+
+
+class OsqueryFileCollectInfo(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryFileCollectInfo` proto."""
+
+  protobuf = osquery_pb2.OsqueryFileCollectInfo
+  rdf_deps = [rdf_client_fs.StatEntry]
--- a/grr/core/grr_response_core/lib/rdfvalues/osquery.py
+++ b/grr/core/grr_response_core/lib/rdfvalues/osquery.py
@@ -14,10 +14,10 @@ from grr_response_core.lib.rdfvalues import client_fs as rdf_client_fs
 from grr_response_proto import osquery_pb2
 
 
-class OsqueryArgs(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryArgs` proto."""
+class OsqueryFlowArgs(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryFlowArgs` proto."""
 
-  protobuf = osquery_pb2.OsqueryArgs
+  protobuf = osquery_pb2.OsqueryFlowArgs
   rdf_deps = []
 
   def __init__(self, *args, **kwargs):
@@ -27,10 +27,10 @@ class OsqueryArgs(rdf_structs.RDFProtoStruct):
       self.timeout_millis = 5 * 60 * 1000  # 5 minutes.
 
 
-class OsqueryActionArgs(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryActionArgs` proto."""
+class OsqueryArgs(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryArgs` proto."""
 
-  protobuf = osquery_pb2.OsqueryActionArgs
+  protobuf = osquery_pb2.OsqueryArgs
   rdf_deps = []
 
 
@@ -125,8 +125,8 @@ class OsqueryProgress(rdf_structs.RDFProtoStruct):
   rdf_deps = [OsqueryTable]
 
 
-class OsqueryFileCollectInfo(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryFileCollectInfo` proto."""
+class OsqueryCollectedFile(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryCollectedFile` proto."""
 
-  protobuf = osquery_pb2.OsqueryFileCollectInfo
+  protobuf = osquery_pb2.OsqueryCollectedFile
   rdf_deps = [rdf_client_fs.StatEntry]
--- a/grr/proto/grr_response_proto/osquery.proto
+++ b/grr/proto/grr_response_proto/osquery.proto
@@ -1,5 +1,8 @@
 syntax = "proto2";
 
+import "grr_response_proto/jobs.proto";
+
+
 package grr;
 
 // https://github.com/facebook/osquery/blob/5188ce5288abe0e323b8e8bd364f452134a62d00/include/osquery/tables.h#L134
@@ -14,6 +17,13 @@ enum OsqueryType {
 }
 
 message OsqueryArgs {
+  optional string query = 1;
+  optional uint64 timeout_millis = 2;
+  optional bool ignore_stderr_errors = 3;
+  repeated string file_collect_columns = 4;
+}
+
+message OsqueryActionArgs {
   optional string query = 1;
   optional uint64 timeout_millis = 2;
   optional bool ignore_stderr_errors = 3;
@@ -46,4 +56,8 @@ message OsqueryResult {
 message OsqueryProgress {
   optional OsqueryTable partial_table = 1;
   optional uint64 total_row_count = 2;
+}
+
+message OsqueryFileCollectInfo {
+  optional StatEntry stat_entry = 1;
 }
--- a/grr/proto/grr_response_proto/osquery.proto
+++ b/grr/proto/grr_response_proto/osquery.proto
@@ -16,14 +16,14 @@ enum OsqueryType {
   BLOB = 6;
 }
 
-message OsqueryArgs {
+message OsqueryFlowArgs {
   optional string query = 1;
   optional uint64 timeout_millis = 2;
   optional bool ignore_stderr_errors = 3;
   repeated string file_collect_columns = 4;
 }
 
-message OsqueryActionArgs {
+message OsqueryArgs {
   optional string query = 1;
   optional uint64 timeout_millis = 2;
   optional bool ignore_stderr_errors = 3;
@@ -58,6 +58,6 @@ message OsqueryProgress {
   optional uint64 total_row_count = 2;
 }
 
-message OsqueryFileCollectInfo {
+message OsqueryCollectedFile {
   optional StatEntry stat_entry = 1;
 }
--- a/grr/proto/grr_response_proto/osquery.proto
+++ b/grr/proto/grr_response_proto/osquery.proto
@@ -20,7 +20,7 @@ message OsqueryFlowArgs {
   optional string query = 1;
   optional uint64 timeout_millis = 2;
   optional bool ignore_stderr_errors = 3;
-  repeated string file_collect_columns = 4;
+  repeated string file_collection_columns = 4;
 }
 
 message OsqueryArgs {
--- a/grr/server/grr_response_server/flows/general/osquery.py
+++ b/grr/server/grr_response_server/flows/general/osquery.py
@@ -12,6 +12,7 @@ from grr_response_core.lib.util import compatibility
 from grr_response_core.lib.rdfvalues import paths as rdf_paths
 from grr_response_core.lib.rdfvalues import client_fs as rdf_client_fs
 from grr_response_core.lib.rdfvalues import osquery as rdf_osquery
+from grr_response_core.lib.rdfvalues import client_action as rdf_client_action
 from grr_response_server.rdfvalues import flow_objects as rdf_flow_objects
 from grr_response_server.rdfvalues import objects as rdf_objects
 from grr_response_server.flows.general import transfer
@@ -82,7 +83,7 @@ class FileCollectionLimitsExceeded(flow_base.FlowError):
   pass
 
 
-class OsqueryFlow(flow_base.FlowBase):
+class OsqueryFlow(transfer.MultiGetFileLogic, flow_base.FlowBase):
   """A flow mixin wrapping the osquery client action."""
 
   friendly_name = "Osquery"
@@ -124,15 +125,51 @@ class OsqueryFlow(flow_base.FlowBase):
       responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
   ) -> None:
     pathspecs = self._GetPathSpecsToCollect(responses)
-    self.CallFlow(
-        transfer.MultiGetFile.__name__,
-        pathspecs=pathspecs,
-        file_size=FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES,
-        network_bytes_limit=FILE_COLLECT_LIMIT_MAX_TOTAL_BYTES,
-        next_state=compatibility.GetName(self.ProcessCollectedFiles))
+
+    stub = server_stubs.GetFileStat
+    for pathspec in pathspecs:
+      request = rdf_client_action.GetFileStatRequest(pathspec=pathspec)
+      self.CallClient(
+          stub,
+          request,
+          next_state=compatibility.GetName(self._StatForFileArrived))
+
+  def _StatForFileArrived(
+      self,
+      responses: flow_responses.Responses[rdf_client_fs.StatEntry],
+  ) -> None:
+    if not responses.success:
+      # TODO(simstoykov): Take note of this and report to the user with other
+      # flow results, instead of failing the flow.
+      raise flow_base.FlowError(
+        f"Error when attempted to stat file: {responses.status}")
+
+    if len(responses) != 1:
+      raise NotImplementedError(
+          f"Response from stat has length {len(responses)}, instead of 1.")
+
+    stat_entry = list(responses)[0]
+
+    if stat_entry.st_size > FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES:
+      # TODO(simstoykov): Report to the user with other flow results instead of
+      # failing the flow.
+      file_path = stat_entry.pathspec.CollapsePath()
+      raise flow_base.FlowError(
+          f"File with path {file_path} is too big: {stat_entry.st_size} bytes "
+          f"when the limit is {FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES} bytes")
+
+    next_total_size = self.state.total_collected_bytes + stat_entry.st_size
+    if next_total_size > FILE_COLLECT_LIMIT_MAX_TOTAL_BYTES:
+      # TODO(simstoykov): Consider reporting to the user and giving the
+      # collected files so far and the other results.
+      raise flow_base.FlowError("Files for collection exceed the total size "
+          f"limit of ${FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES} bytes")
+
+    self.StartFileFetch(stat_entry.pathspec)
 
   def Start(self):
-    super(OsqueryFlow, self).Start()
+    super(OsqueryFlow, self).Start(
+        file_size=FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES)
 
     if len(self.args.file_collect_columns) > FILE_COLLECT_LIMIT_MAX_COLUMNS:
       message = (f"Requested file collection for "
@@ -142,6 +179,7 @@ class OsqueryFlow(flow_base.FlowBase):
 
     self.state.progress = rdf_osquery.OsqueryProgress()
     self.state.path_to_count = { }
+    self.state.total_collected_bytes = 0
 
     action_args = rdf_osquery.OsqueryArgs(
         query=self.args.query,
@@ -161,6 +199,7 @@ class OsqueryFlow(flow_base.FlowBase):
 
     self._UpdateProgress(responses)
 
+    # Make Osquery results available before files are collected
     for response in responses:
       self.SendReply(response)
 
@@ -169,15 +208,12 @@ class OsqueryFlow(flow_base.FlowBase):
   def GetProgress(self) -> rdf_osquery.OsqueryProgress:
     return self.state.progress
 
-  def ProcessCollectedFiles(
-      self,
-      responses: flow_responses.Responses[rdf_client_fs.StatEntry],
-  ) -> None:
-    if not responses.success:
-      raise flow_base.FlowError(responses.status)
-
-    for result in responses:
-      self.SendReply(result)
+  def ReceiveFetchedFile(self,
+                         stat_entry,
+                         file_hash,
+                         request_data=None,
+                         is_duplicate=False) -> None:
+    self.SendReply(stat_entry)
 
   def GetFilesArchiveMappings(
       self,
--- a/grr/server/grr_response_server/flows/general/osquery.py
+++ b/grr/server/grr_response_server/flows/general/osquery.py
@@ -15,6 +15,7 @@ from grr_response_core.lib.rdfvalues import osquery as rdf_osquery
 from grr_response_core.lib.rdfvalues import client_action as rdf_client_action
 from grr_response_server.rdfvalues import flow_objects as rdf_flow_objects
 from grr_response_server.rdfvalues import objects as rdf_objects
+from grr_response_server.rdfvalues import osquery as rdf_server_osquery
 from grr_response_server.flows.general import transfer
 from grr_response_server import flow_base
 from grr_response_server import flow_responses
@@ -24,10 +25,10 @@ from grr_response_server.databases import db
 
 TRUNCATED_ROW_COUNT = 10
 
-FILE_COLLECT_LIMIT_MAX_COLUMNS = 5
-FILE_COLLECT_LIMIT_MAX_ROWS = 1000
-FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES = 2**30 // 2  # 1/2 GiB
-FILE_COLLECT_LIMIT_MAX_TOTAL_BYTES = 2**30  # 1 GiB
+FILE_COLLECTION_MAX_COLUMNS = 5
+FILE_COLLECTION_MAX_ROWS = 1000
+FILE_COLLECTION_MAX_SINGLE_FILE_BYTES = 2**30 // 2  # 1/2 GiB
+FILE_COLLECTION_MAX_TOTAL_BYTES = 2**30  # 1 GiB
 
 
 def _GetTotalRowCount(
@@ -67,12 +68,12 @@ def _GetTruncatedTable(
 
 def _ExtractFileInfo(
     result: rdf_flow_objects.FlowResult,
-) -> rdf_osquery.OsqueryCollectedFile:
+) -> rdf_server_osquery.OsqueryCollectedFile:
   if not isinstance(result.payload, rdf_client_fs.StatEntry):
     raise _ResultNotRelevantError
 
   stat_entry = result.payload
-  return rdf_osquery.OsqueryCollectedFile(stat_entry=stat_entry)
+  return rdf_server_osquery.OsqueryCollectedFile(stat_entry=stat_entry)
 
 
 class _ResultNotRelevantError(ValueError):
@@ -90,7 +91,7 @@ class OsqueryFlow(transfer.MultiGetFileLogic, flow_base.FlowBase):
   category = "/Collectors/"
   behaviours = flow_base.BEHAVIOUR_BASIC
 
-  args_type = rdf_osquery.OsqueryFlowArgs
+  args_type = rdf_server_osquery.OsqueryFlowArgs
   progress_type = rdf_osquery.OsqueryProgress
 
   def _UpdateProgress(
@@ -104,13 +105,13 @@ class OsqueryFlow(transfer.MultiGetFileLogic, flow_base.FlowBase):
       self,
       responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
   ) -> List[rdf_paths.PathSpec]:
-    if len(self.args.file_collect_columns) == 0:
+    if not self.args.file_collect_columns:
       return []
 
     total_row_count = _GetTotalRowCount(responses)
-    if total_row_count > FILE_COLLECT_LIMIT_MAX_ROWS:
+    if total_row_count > FILE_COLLECTION_MAX_ROWS:
       message = (f"Requested file collection on a table with {total_row_count} "
-          f"rows, but the limit is {FILE_COLLECT_LIMIT_MAX_ROWS} rows")
+          f"rows, but the limit is {FILE_COLLECTION_MAX_ROWS} rows")
       raise FileCollectionLimitsExceeded(message)
 
     file_names = []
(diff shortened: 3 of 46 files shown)


grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -88,5 +166,41 @@ def Process(
...
+      except _ResultNotRelevantError:
+        continue
+
+      client_path = db.ClientPath.FromPathSpec(
+          self.client_id, osquery_file.stat_entry.pathspec)
+
+      # This approach keeps the client directory structure
+      # TODO(simstoykov): Check whether we should really keep the tree structure
+      # TODO(simstoykov): What checks need to be performed on the full_client_path?
+      full_client_path = osquery_file.stat_entry.pathspec.path
```
mbushkov, line 203: We should use this to convert the pathspec to a filename:
https://github.com/google/grr/blob/ec1bd364f889e03f26a97f091e7098bc6db8f6f3/grr/server/grr_response_server/rdfvalues/objects.py#L385

PathSpec is a complex data structure, and the function above handles a lot of edge cases correctly.

See, for example (note that it just uses the filename without the path, though):
https://github.com/google/grr/blob/252af5d85f02121a75f8d83ebbee8e3d7e4f5fc9/grr/server/grr_response_server/flows/general/webhistory.py#L428

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -266,6 +271,130 @@ def testTotalRowCountIncludesAllChunks(self):
 
     self.assertEqual(progress.total_row_count, row_count)
 
+  def testFlowTruncatesFileAboveSingleLimit(self):
+    with temp.AutoTempFilePath() as temp_file_path:
+      target_bytes = 2**20  # 1 MiB
+
+      phrase = "This is a > 1 MiB file\n"
```
mbushkov, line 278: I wonder if this could be written shorter:
(I didn't execute the code below)
```
with io.open(temp_file_path, "wb") as fd:
  fd.write(b"1" * (2**20 + 1))
```

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -266,6 +271,130 @@ def testTotalRowCountIncludesAllChunks(self):
...
+    client_path = db.ClientPath.FromPathSpec(
+        self.client_id, results[1].pathspec)
+    byte_content = file_store.OpenFile(client_path).read()
+    self.assertEqual(len(byte_content), less_than_necessary_bytes)
+
+  def testFlowDoesntCollectSingleFileAboveTotalLimit(self):
+    with temp.AutoTempFilePath() as temp_file_path:
+      target_bytes = 2**20  # 1 MiB
+
+      phrase = "This is a > 1 MiB file\n"
```
mbushkov, line 314: Same comment here. The code is easy to read, but feels a tad overly verbose.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -266,6 +271,130 @@ def testTotalRowCountIncludesAllChunks(self):
...
+        fd.write("Just sample text to put in the file.".encode('utf-8'))
+
+      table = f"""
+      [
+        {{ "collect1": "{temp_file_path}", "collect2": "{temp_file_path}"}}
+      ]
+      """
+
+      with mock.patch.object(osquery_flow, "FILE_COLLECT_LIMIT_MAX_COLUMNS", 1):
+        with osquery_test_lib.FakeOsqueryiOutput(stdout=table, stderr=""):
```
mbushkov, line 377: The flow should fail on startup. I'd change the flow logic and would just use "StartFlow" here and make sure that it raises.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -266,6 +271,130 @@ def testTotalRowCountIncludesAllChunks(self):
...
+          with self.assertRaises(RuntimeError):
```
mbushkov, line 378: Can we narrow down the exception?

mbushkov (changes requested):
Looks good. Please see my comments below.

grr/proto/grr_response_proto/osquery.proto
```diff
@@ -17,6 +20,13 @@ message OsqueryArgs {
   optional string query = 1;
   optional uint64 timeout_millis = 2;
   optional bool ignore_stderr_errors = 3;
+  repeated string file_collect_columns = 4;
+}
+
+message OsqueryActionArgs {
```
panhania, line 26: Unfortunately, this is not backward-compatible. It is stupid and I argued multiple times to change it because it breaks one of the fundamental rules of Protocol Buffers, but clients validate names of massages. So, renaming action arguments is not an option. For flow there is no such constraint, though.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -50,6 +62,30 @@ def _GetTruncatedTable(
...
+    raise _ResultNotRelevantError
+
+  stat_entry = result.payload
+  return rdf_osquery.OsqueryFileCollectInfo(stat_entry=stat_entry)
+
+
+def _PathSpecFromFileName(file_name: str) -> rdf_paths.PathSpec:
+  return rdf_paths.PathSpec(
+      path=file_name,
+      pathtype=rdf_paths.PathSpec.PathType.OS)
```
panhania, line 78: I don't think it is a big win over `rdf_paths.PathSpec.OS(path=filename)`.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +103,55 @@ def _UpdateProgress(
...
+        file_names.extend(osquery_result.table.Column(column))
+
+    return list(map(_PathSpecFromFileName, file_names))
+
+  def _FileCollectionFromColumns(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
+  ) -> None:
+    pathspecs = self._GetPathSpecsToCollect(responses)
+    self.CallFlow(
```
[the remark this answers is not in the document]
simstoykov: How would we limit the total number of bytes allowed? As you suggested, I used `network_bytes_limit` as a parameter to `CallClient`, but I'm not sure where this would fit here since we are not calling a separate flow?

Also how can we speed it up by using `MultiGetLogic`? I don't know too much about their implementation, and to me it sounds like we are going to implement new `MultiGetFile` ourselves if we use `MultiGetLogic`?

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -88,5 +166,41 @@ def Process(
     for response in responses:
       self.SendReply(response)
 
+    try:
+      self._FileCollectionFromColumns(responses)
+    except _FileCollectionLimitsExceeded as e:
+      raise flow_base.FlowError(str(e))
```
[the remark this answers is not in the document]
simstoykov: I wanted to get the error message, but I guess we can just pass the old exception again.

Also I can just make `FileCollectionLimitsExceeded` extend `FlowError` and don't re-rase (I am under the impression that we must throw `FlowError`s).

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Do you propose we use `ClientPath.FromPathSpec` (as in the `webhistory.py` example) or `PathInfo.FromPathSpec` (the one referenced)? Also what method would give the full path, and what edge cases would that cover? Please note that I use `full_client_path` to construct the `target_path`, not the `client_path`.

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: We will need to make changes to the test utils: https://github.com/google/grr/blob/v3.4.2.4-release/grr/test_lib/flow_test_lib.py#L426 . Do you reckon it is worth to do in this PR? We will also need to check if other tests use this `RuntimeError` I think.

grr/server/grr_response_server/flows/general/osquery.py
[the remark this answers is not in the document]
mbushkov: IIUC we can raise anything we want. See https://github.com/google/grr/blob/b3daa4d21e85395dc7a04143a796f0ddf0304dbf/grr/server/grr_response_server/flow_base.py#L713

Basically, any exception that's raised inside a flow's state method will be caught and logged.

grr/server/grr_response_server/flows/general/osquery.py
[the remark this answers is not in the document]
mbushkov: > I don't know too much about their implementation, and to me it sounds like we are going to implement new MultiGetFile ourselves if we use MultiGetLogic?

MultiGetFile is indeed a thin layer on top of the MultiGetFileLogic and, unfortunately, not the best one. The biggest issue - MultiGetFile doesn't report errors correctly. Another issue: there's no way you can influence MultiGetFile's truncation logic. With MultiGetFileLogic you could, for every file, first do a StatFile client action call to check its side and then make a decision whether you want to download it.

By "suboptimal" I meant that MultiGetFile has known issues and, coincidentally, using MultiGetFileLogic offers a workaround around these issues. Fundamentally, I think we need to fix MultiGetFile to make it usable in such scenarios as we have here. I don't think we should do any changes within this PR.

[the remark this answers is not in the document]
mbushkov: Using `network_bytes_limit` will, IIUC, have no effect on MultiGetFile. It also has a "file_size" parameter: https://github.com/google/grr/blob/252af5d85f02121a75f8d83ebbee8e3d7e4f5fc9/grr/proto/grr_response_proto/flows.proto#L1348

But it will silently truncate the files instead of skipping.

grr/server/grr_response_server/flows/general/osquery.py
mbushkov: Good catch - I talk about `PathInfo.FromPathSpec`, but `webhistory.py` has `db.ClientPath.FromPathSpec`. `db.ClientPath.FromPathSpec` is a thin wrapper on top of `PathInfo.FromPathSpec`, so I'd suggest using `PathInfo.FromPathSpec`.

The edge cases it handles are these:
* PathSpec is a recursive data structure. If you just use `pathspec.path`, you ignore the nested component of the pathspec. (see https://github.com/google/grr/blob/ec1bd364f889e03f26a97f091e7098bc6db8f6f3/grr/proto/grr_response_proto/jobs.proto#L612)
* Path elements might have stream names (used in NTFS), see https://github.com/google/grr/blob/ec1bd364f889e03f26a97f091e7098bc6db8f6f3/grr/server/grr_response_server/rdfvalues/objects.py#L394

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -4,16 +4,31 @@
...
 from grr_response_server import server_stubs
+from grr_response_server.databases import db
 
 
 TRUNCATED_ROW_COUNT = 10
 
+FILE_COLLECT_LIMIT_MAX_COLUMNS = 5
+FILE_COLLECT_LIMIT_MAX_ROWS = 1000
+FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES = 2**30 // 2  # 1/2 GiB
+FILE_COLLECT_LIMIT_MAX_TOTAL_BYTES = 2**30  # 1 GiB
```
panhania, line 30: Having both `MAX` and `LIMIT` feels kind of redundant. Why not just `FILLE_COLLECTION_COLUMN_LIMIT` etc.?

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -50,14 +65,32 @@ def _GetTruncatedTable(
   return result
 
 
-class OsqueryFlow(flow_base.FlowBase):
+def _ExtractFileInfo(
+    result: rdf_flow_objects.FlowResult,
+) -> rdf_osquery.OsqueryCollectedFile:
+  if not isinstance(result.payload, rdf_client_fs.StatEntry):
+    raise _ResultNotRelevantError
```
panhania, line 72: I would just throw `TypeError` and not bother with additional exception type.

grr/core/grr_response_core/lib/rdfvalues/osquery.py
```diff
@@ -10,13 +10,14 @@
...
 
 
-class OsqueryArgs(rdf_structs.RDFProtoStruct):
-  """An RDF wrapper class for the `OsqueryArgs` proto."""
+class OsqueryFlowArgs(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryFlowArgs` proto."""
 
-  protobuf = osquery_pb2.OsqueryArgs
+  protobuf = osquery_pb2.OsqueryFlowArgs
   rdf_deps = []
```
panhania, line 21: This is a server-only RDF value, putting them under `grr/server/grr_response_server/rdfvalues` would be more appropriate.

grr/core/grr_response_core/lib/rdfvalues/osquery.py
```diff
@@ -115,3 +123,10 @@ class OsqueryProgress(rdf_structs.RDFProtoStruct):
 
   protobuf = osquery_pb2.OsqueryProgress
   rdf_deps = [OsqueryTable]
+
+
+class OsqueryCollectedFile(rdf_structs.RDFProtoStruct):
+  """An RDF wrapper class for the `OsqueryCollectedFile` proto."""
+
+  protobuf = osquery_pb2.OsqueryCollectedFile
+  rdf_deps = [rdf_client_fs.StatEntry]
```
panhania, line 132: Same with this one, it is a server-side only RDF value.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +100,94 @@ def _UpdateProgress(
     self.state.progress.partial_table = _GetTruncatedTable(responses)
     self.state.progress.total_row_count = _GetTotalRowCount(responses)
 
+  def _GetPathSpecsToCollect(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
+  ) -> List[rdf_paths.PathSpec]:
+    if len(self.args.file_collect_columns) == 0:
+      return []
```
panhania, line 108: The internal linter will say to use `if self.args.file_collect_columns:` instead. I don't agree with this and I would prefer to do it like it is here, but it is no up to me ¯\\\_(ツ)\_/¯

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +100,94 @@ def _UpdateProgress(
...
+      responses: flow_responses.Responses[rdf_client_fs.StatEntry],
+  ) -> None:
+    if not responses.success:
+      # TODO(simstoykov): Take note of this and report to the user with other
+      # flow results, instead of failing the flow.
+      raise flow_base.FlowError(
+        f"Error when attempted to stat file: {responses.status}")
+
+    if len(responses) != 1:
+      raise NotImplementedError(
```
panhania, line 148: `ValueError` feels more appropriate, `NotImplementedError` is more for TODO-like things.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +100,94 @@ def _UpdateProgress(
...
+        file_size=FILE_COLLECT_LIMIT_MAX_SINGLE_FILE_BYTES)
+
+    if len(self.args.file_collect_columns) > FILE_COLLECT_LIMIT_MAX_COLUMNS:
+      message = (f"Requested file collection for "
+          f"{len(self.args.file_collect_columns)} columns, but the limit is"
+          f"{FILE_COLLECT_LIMIT_MAX_COLUMNS} columns")
+      raise FileCollectionLimitsExceeded(message)
+
     self.state.progress = rdf_osquery.OsqueryProgress()
+    self.state.path_to_count = { }
```
panhania, line 181: Unnecessary whitespace inside dictionary literal.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -155,17 +160,18 @@ def setUp(self):
     super(FakeOsqueryFlowTest, self).setUp()
     self.client_id = self.SetupClient(0)
 
-  def _InitializeFlow(self, query):
+  def _InitializeFlow(self, query, file_collect_columns=[]):
```
panhania, line 163: Mutable instances should not be used when specifying default parameters as they are global to the entire program. The correct way is something like this:

```python
def _InitializeFlow(self, query: str, file_collect_columns: Optional[List[str]] = None):
    if file_collect_columns is None:
        file_collect_columns = []
    (...)
```

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -155,17 +160,18 @@ def setUp(self):
...
         client_id=self.client_id,
         token=self.token,
-        query=query)
+        query=query,
+        file_collect_columns=file_collect_columns)
     return session_id
 
-  def _RunFlow(self, query):
-    session_id = self._InitializeFlow(query)
+  def _RunFlow(self, query, file_collect_columns=[]):
```
panhania, line 173: Ditto.

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Yes I agree it would be better to leave one. It's a small thing, but isn't it better to preserve `MAX` since technically we could `LIMIT` both the minimum or the maximum?

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Isn't the code more readable if the exception has a descriptive name at the place it is handled? https://github.com/google/grr/pull/901/files/31944c1ea3aadc3c54dc6e2f459464439685ef8a#diff-ad469360f11bc0231fc5350d52ad08e36d21fb93828c0f47aa45fa24393b54bbR242

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: I see. I think it should be `if not self.args.file_collect_columns:` since it is the case when the list has zero length (and it is a Falsy value)?

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: Good catch, didn't know that!

simstoykov: I decided to remove the default value altogether and pass an empty list explicitly in the other uses. I found this clearer, and adding `[]` to a few places doesn't clutter the code too much, but it will help in debugging in case there are errors. Please comment if you disagree.

grr/server/grr_response_server/flows/general/osquery.py
panhania: I didn't realize you catch it, you do then it makes sense. But then I am not sure if it is good behaviour: if there is a type error it means something is seriously wrong. If we silently ignore such errors, the system will pretend everything is fine even though it is not.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -261,11 +268,182 @@ def testTotalRowCountIncludesAllChunks(self):
...
+          with self.assertRaises(RuntimeError):
+            self._RunFlow("Doesn't matter", ["collect_collumn"])
+
+  def testArchiveMappingsForMultipleFiles(self):
+    with temp.AutoTempFilePath() as temp_file_path1, \
+         temp.AutoTempFilePath() as temp_file_path2:
+      with io.open(temp_file_path1, "wb") as fd:
+        fd.write("Just sample text to put in the file 1.".encode('utf-8'))
+      with io.open(temp_file_path2, "wb") as fd:
+        fd.write("Just sample text to put in the file 2.".encode('utf-8'))
```
panhania, line 377: You can use:

```python
with temp.AutoTempDirPath(remove_non_empty=True) as tempdir_path:
    temp_filepath_1 = os.path.join(tempdir_path, "foo")
    temp_filepath_2 = os.path.join(tempdir_path, "bar")
    (...)
```

This uses only one context manager with temporary state. And, as I have learned the hard way myself, is also easier to debug in case the test fails—you hit a breakpoint, go to the temporary folder and can observe the state of the only two files there rather than navigating through thousand garbage files in your `/tmp`.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -85,8 +200,51 @@ def Process(
...
+      try:
+        osquery_file = _ExtractFileInfo(result)
+      except _ResultNotRelevantError:
+        continue
+
+      client_path = db.ClientPath.FromPathSpec(
+          self.client_id, osquery_file.stat_entry.pathspec)
+      target_path = GenerateTargetPath(osquery_file.stat_entry.pathspec)
+
+      yield flow_base.ClientPathArchiveMapping(client_path, target_path)
```
panhania, line 250: `ClientPathArchiveMapping` is a named tuple, so constructing it should use named fields.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -85,8 +200,51 @@ def Process(
...
+      else:
+        return ""
+
+    def GenerateTargetPath(pathspec: rdf_paths.PathSpec) -> str:
+      path_info = rdf_objects.PathInfo.FromPathSpec(pathspec)
+      client_string_path = "/".join(path_info.components)
+
+      target_path = f"osquery_collected_files/{client_string_path}"
+      target_path += _GetUniqueSuffixFor(target_path)
+      return target_path
```
panhania, line 238: These utilities are used only once and declared here, making the code quite deeply nested. Sometimes refactoring logic into utility improves readability, here I feel like it only introduces unnecessary complexity.

If they make sense outside of this class, I would move them to stand alone methods. If they don't, I would just inline  the definition.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -155,17 +161,18 @@ def setUp(self):
     super(FakeOsqueryFlowTest, self).setUp()
     self.client_id = self.SetupClient(0)
 
-  def _InitializeFlow(self, query):
+  def _InitializeFlow(self, query: str, file_collect_columns: List[str]):
```
panhania, line 164: If majority of calls doesn't care about this parameter, I would just make it optional.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -261,11 +268,182 @@ def testTotalRowCountIncludesAllChunks(self):
...
+          "FILE_COLLECTION_MAX_TOTAL_BYTES",
+          less_than_necessary_bytes):
+        with osquery_test_lib.FakeOsqueryiOutput(stdout=table, stderr=""):
+          with self.assertRaises(RuntimeError):
+            self._RunFlow("Doesn't matter", ["collect_collumn"])
+
+  def testFlowCollectFile(self):
+    with temp.AutoTempFilePath() as temp_file_path:
+      with io.open(temp_file_path, "wb") as fd:
+        fd.write("Just sample text to put in the file.".encode('utf-8'))
```
panhania, line 323: You can also open the file using the UTF-8 encoding and write string directly. I find it a bit more natural:

```python
with io.open(temp_file_path, mode="w", encoding="utf-8") as fd:
    fd.write("foobar")
```

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: It's not really type **error**, it really is that the result is not relevant. Maybe I should remove the "Error" in _ResultNotRelevantError"? Our flow sends two types of results: `OsqueryResult` for the actual osquery table results, and `StatEntry` for each collected file. This is because of `GetFilesArchiveMappings`: it is called with the results generated by the flow (in order to generate the file archive).

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: That's a good point, thanks.

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: I agree; this should be changed at a few other places too, for example:
https://github.com/google/grr/blob/master/grr/server/grr_response_server/flows/general/webhistory.py#L439

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Even though the logic is not massively complicated, I would like to keep at least one of those methods as a separate named block in the code, since it is not exactly trivial (e.g. we are maintaining state for the number of paths that we have seen).

I will try to inline `_GetUniqueSuffixFor` into `_GenerateTargetArchivePath`.

simstoykov: I agree nesting is a bit controversial; I will move it out as a class method.

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: Please see my comment above: https://github.com/google/grr/pull/901#discussion_r545128095

panhania: Ah, terrible GitHub review flow strikes again.

No strong opinion here, can stay as it is.

grr/server/grr_response_server/flows/general/osquery.py
panhania: Class methods are disencouraged by our internal style guidelines. They are okay in some cases, but I don't think this is one of them. Why no make it a top-level private method?

grr/server/grr_response_server/flows/general/osquery_test.py
panhania: Ah, terrible GitHub review flow strikes again.

No strong opinion here, can stay as it is.

grr/server/grr_response_server/flows/general/osquery_test.py
panhania: Ah, terrible GitHub review flow strikes again.

No strong opinion here, can stay as it is.

grr/server/grr_response_server/flows/general/osquery.py
panhania: Yep!

grr/server/grr_response_server/flows/general/osquery.py
panhania: > Our flow sends two types of results: OsqueryResult for the actual osquery table results, and StatEntry for each collected file.

This is very unfortunate, actually. Flows should be well typed and return a specific type. Wouldn't it be possible to split `OsqueryResult` into `OsqueryResult` (for client action) and `OsqueryFlowResult` (for the flow), similarly how you do with the args, and add a `collected_files` field into the flow one?

That would make it correctly typed, avoids having this awkward `isinstance` logic and does not abuse exceptions for control flow.

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Sorry I meant instance method.

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Which one of the two are you suggesting (or maybe something else?):
1. Bundling osquery results with flow collection results.
2. Making `OsqueryFlowResult` effectively a union type of the current `OsqueryResult` and `StatEntry`, and just populating it with one or the other in the result callback methods (`Process` and `ReceiveFetchedFile`)?

Also by "and does not abuse exceptions for control flow" do you mean that the code is difficult to read in its current version, or maybe that it is inefficient? From my understanding it is not uncommon to use exception for control flow in Python, since they do not add much overhead. Also the construction looks at least as clear as having an if clause.

panhania: The first option. I think it makes sense to have a field with stat entries for the collected file, it is part of the result afterall.

> From my understanding it is not uncommon to use exception for control flow in Python

It is not uncommon, but I think it is unfortunate. Afterall, as the name suggests, exceptions are for exceptional situations, which we clearly not have here. It's not that they incur some performance hit (I mean it is Python, we don't care about that) but rather make the code feel awkward: you raise but with the assumption that the caller is going to catch this—however this cannot be specified in any way at the type level.

> Also the construction looks at least as clear as having an if clause.

For now, because you just wrote it and you know all the assumptions. Once the code evolves and is read by others, it might no longer be the case. I am not saying this should never be done (I did it myself couple times!), but if there is an easy solution that does not involve using this mechanism, why not?

simstoykov: > The first option. I think it makes sense to have a field with stat entries for the collected file, it is part of the result afterall.'

I like the idea, but I'm not sure it is suitable for this case, as it would introduce the following complications:

1. We receive the two kinds of results at different places: osquery results in the `Process` callback, and file collection results in the `ReceiveFetchedFile` callback. Thus we would need to add some state and logic in order to bundle them together.
2. Osquery results are received and inspected in full before file collection is initiated. In order to bundle them with the file collection results, osquery results would need to sit in memory while the file collection stage completes. I guess this wouldn't be a big issue now, since we are keeping all osquery results in memory anyways (as opposed to iterating through them), but it would stop us from optimizing this in the future too.
3. We send requests for file collection one by one, and receive the results one by one. We would need some mechanism to determine when we have received all files we have requested (or errors for them). And if for some reason we don't receive a response for all files we attempted to collect, then our flow will hang forever, waiting for them, so that they can be bundled and sent together with the osquery results (that wait in memory).

What are your thoughts on those issues?

panhania: > Thus we would need to add some state and logic in order to bundle them together.

It is extremely trivial logic (put a value into the flow state, get a value from the flow state), I don't think there is anything to be worried here.

> I guess this wouldn't be a big issue now, since we are keeping all osquery results in memory anyways (as opposed to iterating through them), but it would stop us from optimizing this in the future too.

Sounds a bit like premature optimization. If the list of files to download is so large that there are problems with keeping it in memory, it is likely not a list you would ever want to fetch from the client.

> We send requests for file collection one by one, and receive the results one by one.

Why? The whole point of `MultiGetFile` is that it can do multiple files. And why do we use `MultiGetFileMixin` rather than spawning a subflow? The whole mixin system in GRR is extremely fragile and depends on method names to glue everything together, so I would prefer to stay away from this model.

simstoykov: > Sounds a bit like premature optimization. If the list of files to download is so large that there are problems with keeping it in memory, it is likely not a list you would ever want to fetch from the client.

My point was that we would have to keep the osquery results in memory, not the list of files. However, I just realized that we enforce limits on the table size for file collection (at most 1000 rows), so it would not be as bad as having millions of rows in memory.

> And why do we use `MultiGetFileMixin` rather than spawning a subflow?

We would like to have full information about which files were successfully collected in full, which were truncated, and which encountered an error, so that we report it to the user. This is not currently supported by `MultiGetFile`. Unfortunately it is out of the scope of this PR to improve `MultiGetFile`, and this is why we went with `MultiGetFileLogic`.

Maybe @mbushkov has something to add to this discussion?

mbushkov: Interesting discussion!

There are a few practical issues with extending OsqueryResult to include the files. The Osquery client action takes measures to not grow OsqueryResult protobufs too much. I think, to be honest, that we should store the actual osquery output in the blobstore and just keep references to blobs in the results, since having 2Mb-sized rows in the database is suboptimal, but that's another discussion.

If we introduce the OsqueryFlowResult by adding file stats to the OsqueryResult, we have to make sure that the proto doesn't grow too much. If the paths are long and we download 1000 files, how do we guarantee that it stays within a reasonable limit? I mean, it should, but it feels like we should at least do some check and checking protos sizes is awkward.

Another practical concern: you would need to fetch all the files first and only then return results and this breaks the practice of flows returning results incrementally, as soon as they arrive.

> Flows should be well typed and return a specific type. Wouldn't it be possible to split OsqueryResult into OsqueryResult (for client action) and OsqueryFlowResult (for the flow), similarly how you do with the args, and add a collected_files field into the flow one?

We do have flows that return multiple things and I think this is a valid concept, primarily because it keeps combining flows easy/trivial. If your flow runs a bunch of other flows, you can always just forward the results. This is heavily used, for example, by the ArtifactCollector. This is what was done in the OsqueryFlow initially with MultiGetFile flow. The reason why Simeon switched to MultiGetFileLogic is because MultiGetFile doesn't return collection errors, but again - this is another problem.

I would suggest to have a wider discussion on the question of returning multiple types of results, but do this outside of the scope of this PR. 

This PR largely supports an already finished UX PR that introduces the Osquery files collection functionality to the UI - https://github.com/google/grr/pull/906. Refactoring the approach used here would also require a significant refactoring of the UX PR. Given the time constraints and the concerns I outlined above and the fact that we don't have a policy of "flows return one thing" yet, I'd suggest proceeding with the current implementation and, if needed, agreeing on future refactorings. @panhania, what do you think?

panhania: > I think, to be honest, that we should store the actual osquery output in the blobstore and just keep references to blobs in the results, since having 2Mb-sized rows in the database is suboptimal

But it is so only in some extreme cases. Usually, results of osquery queries are tiny. Cases where they are not (e.g. listing users on machines where there are hundred thousands of them) would give large results also using ordinary flows.

I don't think it makes sense to complicate logic and add overhead to accommodate for corner cases that other flows don't even consider (the fact that the osquery flow is able to handle them semi-gracefully is just a nice plus).

> If we introduce the OsqueryFlowResult by adding file stats to the OsqueryResult, we have to make sure that the proto doesn't grow too much. If the paths are long and we download 1000 files, how do we guarantee that it stays within a reasonable limit?

I would say: if it doesn't fit, it should not be collected. Maximum path length is 4096 bytes as far as I remember on Linux (on other platforms even less), any reasonable results will easily fit into proto. Other flows also don't consider extreme corner cases.

> We do have flows that return multiple things and I think this is a valid concept, primarily because it keeps combining flows easy/trivial. If your flow runs a bunch of other flows, you can always just forward the results. This is heavily used, for example, by the ArtifactCollector. This is what was done in the OsqueryFlow initially with MultiGetFile flow.

I mean, the fact that a broken mechanism is used in abusive way to implement other broken mechanisms is not a good argument. Because of how "easy/trivial" it is to combine flows with this weakly typed system is exactly the reason why nobody on the team dares to even look at the Frankenstein-style giant pile of intangible mess sewed together from several other mixins that we innocently called "the file finder flow".

The same way you could argue that having functions that return multiple types is a valid concept. And I am sure that there are examples of code that uses such travesties (even in this very codebase) but this is not the way one should program and we already learned this the hard way.

Code is not supposed to be easy or trivial to write. The only thing that matters is how easy it is to read it, reason about it and maintain it.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +101,107 @@ def _UpdateProgress(
     self.state.progress.partial_table = _GetTruncatedTable(responses)
     self.state.progress.total_row_count = _GetTotalRowCount(responses)
 
+  def _GetPathSpecsToCollect(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
```
panhania, line 106: Since you just iterate on responses, you can require `Iterable[rdf_osquery.OsqueryResult]` (and this is something `flow_responses.Responses` implements). It makes the code a bit simple and more reusable.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +101,107 @@ def _UpdateProgress(
...
+    file_names = []
+    for osquery_result in responses:
+      for column in self.args.file_collect_columns:
+        file_names.extend(osquery_result.table.Column(column))
+
+    return [rdf_paths.PathSpec.OS(path=file_name) for file_name in file_names]
+
+  def _FileCollectionFromColumns(
+      self,
+      responses: flow_responses.Responses[rdf_osquery.OsqueryResult],
```
panhania, line 126: Ditto.

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -85,8 +213,37 @@ def Process(
...
+    self._FileCollectionFromColumns(responses)
+
   def GetProgress(self) -> rdf_osquery.OsqueryProgress:
     return self.state.progress
+
+  def ReceiveFetchedFile(self,
+                         stat_entry,
+                         file_hash,
+                         request_data=None,
+                         is_duplicate=False) -> None:
```
panhania, line 229: Missing type annotations and `del` lines that dispose of unused arguments.

grr/server/grr_response_server/rdfvalues/osquery.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
```
panhania, line 1: No need for this line, this is not an executable file.

panhania (approved):
From the pseudo-offline discussion: the current design can be merged for now but should be refactored in the future.

Left some small nitpicks, other than that we are good to go.

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: We had an offline discussion covering the points above.

### `MultiGetFile` flow vs. `MultiGetFileLogic` mixin
We agreed to use the `MultiGetFileLogic` mixin instead of the `MultiGetFile` flow for the time being. Our main point was that `MultiGetFile` doesn't give precise per-file error information. However, after the discussion I realized that it actually does: it populates a list with file collection outcomes [in its state](https://github.com/simstoykov/grr/blob/filepaths_collection_backend/grr/server/grr_response_server/flows/general/transfer.py#L836), and makes it available [in its progress](https://github.com/simstoykov/grr/blob/filepaths_collection_backend/grr/server/grr_response_server/flows/general/transfer.py#L796).

There was another point, which is still valid: `MultiGetFileLogic` allows us to collect files one by one, incrementally. This is consistent with the current approach, because before we send requests for collection we first stat files (to check their size). The stat results are received one by one, so it is simpler to send requests for collection one by one again, instead of having to merge them together.

At this point the logic is not complicated or coupled with anything, so it should be easy to change `MultiGetFileLogic` mixin use to `MultiGetFile` flow use in the future.

### Bundling file collection and osquery results into one protobuf, so that the flow produces one type of results
This discussion produced what seems to be a better design than the current one: we split the joint osquery-and-file-collection flow into two flows - one for osquery, and another one for collection files from osquery results. In this way we can have a unique flow result type for each of the two, and we also don't need to buffer and bundle results together. Sadly, this is beyond the scope of this PR and my internship. That's why we decided to proceed with the current design, and implement the improvement in the future.

grr/server/grr_response_server/flows/general/osquery_test.py
mbushkov: Ah, I see. Then in this case, RuntimeError seems sufficient.

grr/server/grr_response_server/flows/general/osquery_test.py
mbushkov: Given that the check for FILE_COLLECTION_MAX_COLUMNS happens in the Start method, do you really need to call FakeOsqueryiOutput above? I suppose the flow would never get to actually running anything osquery-related, no?

grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -67,13 +104,107 @@ def _UpdateProgress(
...
+        file_size=FILE_COLLECTION_MAX_SINGLE_FILE_BYTES)
+
+    if len(self.args.file_collect_columns) > FILE_COLLECTION_MAX_COLUMNS:
+      message = (f"Requested file collection for "
+          f"{len(self.args.file_collect_columns)} columns, but the limit is"
+          f"{FILE_COLLECTION_MAX_COLUMNS} columns")
+      raise FileCollectionLimitsExceeded(message)
+
     self.state.progress = rdf_osquery.OsqueryProgress()
+    self.state.path_to_count = {}
```
mbushkov, line 198: I don't think you need to put path_to_count into the state. It's only used in _GenerateTargetArchivePath and _GenerateTargetArchivePath is only used in GetFilesArchiveMappings. A cleaner solution would be to simply pass the path_to_count as an argument to _GenerateTargetArchivePath. What do you think?

mbushkov (changes requested):
Looks good, just a few minor nits.

grr/server/grr_response_server/flows/general/osquery_test.py
simstoykov: Hm this comment is on the previous version of the code; I refactored it exactly as you advise:

https://github.com/google/grr/pull/901/files#diff-84b11633c72d1d91291437e8756c6edec25eb0172500d4a5265c47b5368c3c68R350

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Yes that makes sense to me too. I got confused that we need to serialize this object and pass it across "state" functions, but there is no need as we only use it in `GetFilesArchiveMappings`.

grr/server/grr_response_server/flows/general/osquery_test.py
mbushkov: Ah, GitHub can be so confusing at times... Thanks for clarifying. Resolved!

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: Okay I actually decided to split this functionality into a separate class for generating unique paths. Even though the logic is not super complicated, it bundles the state and the method that act on it; also I got the chance to make the code cleaner (subjective opinion).

mbushkov: Looks good! I like the approach. Even more, I'd consider making _UniquePathGenerator a general-purpose class, so that we can use it for other flows as well (not in this PR).

mbushkov (approved):

PR #910, opened 2020-12-23
MVP error reporting

simstoykov:
image

### DIFF
--- a/grr/proto/grr_response_proto/osquery.proto
+++ b/grr/proto/grr_response_proto/osquery.proto
@@ -14,9 +14,9 @@ enum OsqueryType {
 }
 
 message OsqueryArgs {
+  reserved 3;
   optional string query = 1;
   optional uint64 timeout_millis = 2;
-  optional bool ignore_stderr_errors = 3;
 }
 
 message OsqueryColumn {
@@ -39,8 +39,8 @@ message OsqueryTable {
 }
 
 message OsqueryResult {
+  reserved 2;
   optional OsqueryTable table = 1;
-  optional string stderr = 2;
 }
 
 message OsqueryProgress {
--- a/grr/proto/grr_response_proto/osquery.proto
+++ b/grr/proto/grr_response_proto/osquery.proto
@@ -46,4 +46,5 @@ message OsqueryResult {
 message OsqueryProgress {
   optional OsqueryTable partial_table = 1;
   optional uint64 total_row_count = 2;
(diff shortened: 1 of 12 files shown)


mbushkov (approved):

panhania (approved):

simstoykov:
@panhania can you please take another look? I modified the client action and the server plugin a bit. I also added tests for error reporting.

grr/proto/grr_response_proto/osquery.proto
```diff
@@ -14,9 +14,9 @@ enum OsqueryType {
 }
 
 message OsqueryArgs {
+  reserved 3;
```
panhania, line 27: I would not reorder fields and put the `reserved` on the same line for better diffs.

grr/proto/grr_response_proto/osquery.proto
```diff
@@ -39,8 +39,8 @@ message OsqueryTable {
 }
 
 message OsqueryResult {
+  reserved 2;
```
panhania, line 52: Ditto.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -161,17 +161,23 @@ def setUp(self):
     super(FakeOsqueryFlowTest, self).setUp()
     self.client_id = self.SetupClient(0)
 
-  def _InitializeFlow(self, query: str, file_collection_columns: List[str]):
+  def _InitializeFlow(
+      self,
+      query: str,
+      file_collection_columns: List[str],
+      **kwargs,
```
panhania, line 168: Since you add just one parameter, I would name and type it explicitly rather than use `**kwargs`. At this point it is questionable whether we don't just prefer to use `**kwargs` for everything, but this is something for another PR (if any).

simstoykov: My intention was to use explicit argument names and types for flow arguments (query and columns to collect files from), and `**kwargs` for other "internal" parameters (`check_flow_errors` in this case, which is used by the test library). With this in mind, would you still prefer to have `check_flow_errors` as an explicit argument (or all passed as `kwargs`)?

panhania: I forgot that `check_flow_errors` is part of the flow runner arguments, not a flow argument itself. In that case, I think it is fine to leave it as it is.

PR #911, opened 2020-12-30
Final polishing of the osquery flow

simstoykov:


grr/server/grr_response_server/flows/general/osquery.py
```diff
@@ -152,7 +154,13 @@ def _GetPathSpecsToCollect(
...
     for osquery_result in responses:
       for column in self.args.file_collection_columns:
-        file_names.extend(osquery_result.table.Column(column))
+        try:
+          values_in_column = osquery_result.table.Column(column)
+          file_names.extend(values_in_column)
+        except KeyError:
+          self._UpdateProgressWithError(
+              f"No such column '{column}' to collect files from.")
+          raise
```
panhania, line 163: Scope the `try` block only to the part where the exception can happen (i.e. move the `_.extend(_)` part out of it).

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -279,6 +279,43 @@ def testTotalRowCountIncludesAllChunks(self):
...
+    [
+
+    ]
+    """
+
+    with osquery_test_lib.FakeOsqueryiOutput(stdout=empty_table, stderr=""):
+      results = self._RunFlow("Doesn't matter", ["collect_column"])
+
+    self.assertLen(results, 1)
+    self.assertEqual(type(results[0]), rdf_osquery.OsqueryResult)
```
panhania, line 317: Ditto.

grr/server/grr_response_server/flows/general/osquery_test.py
```diff
@@ -279,6 +279,43 @@ def testTotalRowCountIncludesAllChunks(self):
...
+        {{ "collect_column": "{temp_file_path}" }}
+      ]
+      """
+
+      with osquery_test_lib.FakeOsqueryiOutput(stdout=table, stderr=""):
+        results = self._RunFlow("Doesn't matter", ["collect_column"])
+      
+    self.assertLen(results, 2)
+    self.assertEqual(type(results[0]), rdf_osquery.OsqueryResult)
+    self.assertEqual(type(results[1]), rdf_client_fs.StatEntry)
```
panhania, line 298: `assertIsInstance`

grr/server/grr_response_server/flows/general/osquery.py
simstoykov: `Column(1)` returns an iterator, so the exception actually occurs when I call `extend`: https://github.com/simstoykov/grr/blob/final_polish/grr/core/grr_response_core/lib/rdfvalues/osquery.py#L56

Should I move `values_in_column = osquery_result.table.Column(column)` out of it above, or maybe inline `values_in_column`?

panhania: Ah, these iterators... Personally, I would inline `values_in_column` but you can leave it as it is if you prefer it.

panhania (approved):

PR #969, opened 2022-04-13
Pin pytest and markupsafe versions.

Tati1701:


travis/install.sh
```diff
@@ -5,7 +5,7 @@
 set -ex
 
 source "${HOME}/INSTALL/bin/activate"
-pip install --upgrade pip wheel six setuptools nodeenv
```
Digitalisx, line 8: Since MarkupSafe is dependent on Jinja2, why don't you specify the version in the following areas rather than mentioning the version in Travis YML?
Travis thinks it looks clearer to focus only on testing and deployment, and collecting requirements related to the version.
I experimented with this in my fork Repository. (https://github.com/Digitalisx/grr/pull/1)
Feel free to leave your comments!

Digitalisx (commented):
Hello, @Tati1701 
This seems like a very good patch! 🙂
I'd like to suggest a better way.
Please tell me your opinion!

travis/install.sh
Digitalisx: Oh, I failed the final build, I think your method is right.
I'm sorry that I couldn't make a suggestion after checking a little more!
I hope your work goes well.

Tati1701: Hi @Digitalisx, good point! Do you mind creating a PR with your proposal for me to review? :)

PR #993, opened 2022-07-14
Add Monitoring.http_address to expose Prometheus stats server

coperni:
## Summary
Added configuration option to specify the stats server IP address to bind to. Default option will remain 127.0.0.1 loopback.

I use separate containers (Pods) for the GRR components and need the ability to expose the Prometheus  metric endpoints to  Kubernetes Services which forwards the results to a Prometheus operator. With the current setup the metrics are only accessible from within the container's local loopback device (localhost). 

Edit:

Taking @max-vogler


coperni:
Hi @max-vogler. Any updates on the review?

max-vogler (approved):

PR #1026, opened 2023-03-15
Upload GRR image to GitHub container registry instead of DockerHub.

mbushkov:


.github/workflows/build.yml
```diff
@@ -259,13 +259,14 @@ jobs:
           export BRANCH=$(echo $GITHUB_REF | cut -d'/' -f 3)
           ./appveyor/docker_build/build_docker_image.sh
       - if: ${{ github.event_name == 'push' }}
-        name: Log in to Docker Hub
+        name: Login with Github Container registry
```
panhania, line 262: Nit: `GitHub`.

Also, from what I see, it is called "GitHub Packages", not "GitHub Container".

.github/workflows/build.yml
```diff
@@ -259,13 +259,14 @@ jobs:
...
         uses: docker/login-action@v2
         with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_PASSWORD }}
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - if: ${{ github.event_name == 'push' }}
-        name: Push to ${{ env.DOCKER_REPOSITORY }}
+        name: Push to GitHub Container registry
```
panhania, line 269: Ditto.

panhania (changes requested):

.github/workflows/build.yml
mbushkov: I humbly disagree: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

mbushkov: Fixed Github -> GitHub.

.github/workflows/build.yml
mbushkov: As said above, I think `GitHub container registry` is a valid term (this is what `ghcr.io` stands for) and what's used in the documentation.

panhania (approved):

PR #1128, opened 2025-04-10
GCS Output Plugin for FileFinderResults 

daschwanden:
This PR proposes a GCS Output Plugin implementation for FileFinderResults


mbushkov (approved):
Thanks for implementing this!

PR #1162, opened 2026-04-15
fix: harden robot router GetFileBlob access validation

TristanInSec:
## Summary
- Add client-level access check to `GetFileBlob` in `ApiCallRobotRouter`
- New `_CheckClientRobotAccess` verifies the requesting user has at least one flow on the target client
- Consistent with the per-flow checks (`_CheckFlowRobotId`) on other data-retrieval methods in the same router
- Adds two tests: rejection when no flow exists, success when flow exists

## Test plan
- [x] `testGetFileBlobRaisesIfNoFlowCreatedBySameUser` -- another user's flow does not grant access
- [x] `testGetFileBlobWorksIfFlowWasCreatedBySameUser` -- own flow grants access
- [x] Existing `testGetFileBlobIsDisabledByDefault` unaffected

### DIFF
--- a/grr/server/grr_response_server/gui/api_call_robot_router.py
+++ b/grr/server/grr_response_server/gui/api_call_robot_router.py
@@ -286,15 +286,14 @@ class ApiCallRobotRouter(api_call_router.ApiCallRouterStub):
     """
     flows = data_store.REL_DB.ReadAllFlowObjects(
         client_id=str(client_id),
+        created_by=frozenset([context.username]),
         include_child_flows=False,
     )
-    for flow_obj in flows:
-      if flow_obj.creator == context.username:
-        return
-    raise access_control.UnauthorizedAccess(
-        "Robot user %s has no flows on client %s."
-        % (context.username, client_id)
-    )
+    if not flows:
+      raise access_control.UnauthorizedAccess(
+          "Robot user %s has no flows on client %s."
+          % (context.username, client_id)
+      )
 
   def _FixFileFinderArgs(
       self, source_args: any_pb2.Any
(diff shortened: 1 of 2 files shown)


grr/server/grr_response_server/gui/api_call_robot_router.py
```diff
@@ -277,6 +277,25 @@ def _CheckFlowRobotId(self, client_id, flow_id, context=None):
...
     return flow_obj.flow_class_name
 
+  def _CheckClientRobotAccess(self, client_id, context=None):
+    """Verify the requesting user has at least one flow on this client.
+
+    Used for endpoints that take a client_id but no flow_id, such as
+    GetFileBlob.  This ensures the robot user has some prior relationship
+    with the client, consistent with the per-flow checks on other methods.
+    """
+    flows = data_store.REL_DB.ReadAllFlowObjects(
```
Tati1701, line 287: This can impact the performance quite a bit, depending on the deployment.

Would you consider extending `ReadAllFlowObjects` with an optional `created_by` filter in the DB on a separate PR, then use it here?

Tati1701 (commented):
Thanks for the PR! :D

grr/server/grr_response_server/gui/api_call_robot_router.py
TristanInSec: Good point, that makes sense for deployments with high flow volume. I'll open a separate PR adding a created_by filter to ReadAllFlowObjects, then rebase this one to use it.

TristanInSec:
Hi @mbushkov, friendly ping on this security fix (and the related #1163). CLA is passing, happy to make any changes needed. Thanks!

PR #1164, opened 2026-04-16
feat: add created_by filter to ReadAllFlowObjects

TristanInSec:
## Summary
- Add optional `created_by` parameter to `ReadAllFlowObjects` for filtering flows by creator at the database level
- Implemented in both MySQL and in-memory backends, with input validation in `ValidatingDB`
- Allows callers to efficiently query flows for a specific user without fetching all flows first

Companion to #1162, which currently iterates all flows in Python to find one owned by the requesting user.

## Test plan
- [x] `testReadAllFlowObjectsWithCreatedBy` covers match, single-match, and no-match cases
- [x] Existing `testReadAllFlowObjectsWithNotCreatedBy` and `testReadAllFlowObjectsWithAllConditions` unaffected

### DIFF
--- a/grr/server/grr_response_server/databases/db.py
+++ b/grr/server/grr_response_server/databases/db.py
@@ -2061,7 +2061,7 @@ class Database(metaclass=abc.ABCMeta):
       max_create_time: Optional[rdfvalue.RDFDatetime] = None,
       include_child_flows: bool = True,
       not_created_by: Optional[Iterable[str]] = None,
-      created_by: Optional[str] = None,
+      created_by: Optional[Iterable[str]] = None,
   ) -> list[flows_pb2.Flow]:
     """Returns all flow objects.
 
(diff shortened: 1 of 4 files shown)


grr/server/grr_response_server/databases/db.py
```diff
@@ -2061,6 +2061,7 @@ def ReadAllFlowObjects(
       max_create_time: Optional[rdfvalue.RDFDatetime] = None,
       include_child_flows: bool = True,
       not_created_by: Optional[Iterable[str]] = None,
+      created_by: Optional[str] = None,
```
Tati1701: Let's make this one `Iterable` too, like in other methods.

Tati1701 (changes requested):
Thanks for the quick turnaround!

grr/server/grr_response_server/databases/db.py
TristanInSec: Done. Changed to `Optional[Iterable[str]]` across all four files, with `IN %s` in MySQL and `list()` membership in the in-memory backend. Test updated to cover multi-user filtering.

Tati1701 (approved):

